PR #55, opened 2026-05-21
🐛 Fix Gemini CLI agent metadata frontmatter rejection (#54)

### LINKED ISSUE #54: 🐛 Gemini CLI rejects `type` and `metadata` frontmatter keys in agent definitions
## Problem

Gemini CLI 0.42.0 validates agent frontmatter strictly against a schema that only accepts `name` and `description`. Any additional key causes a validation error at startup:

```
Agent loading error: Failed to load agent from .gemini/agents/test-agent-metadata.md:
Validation failed: Agent Definition:
Unrecognized key(s) in object: 'type', 'metadata'
```

The agent is still loaded in degraded mode (Gemini falls back to ignoring the offending keys), but:
- The error is printed to stdout on every invocation.
- The `metadata.provenance` block (version, canonical URL, feedback URL) is silently dropped — breaking version tracking and friction-reporting attribution for all Gemini CLI agents.
- Any future strict-mode upgrade in Gemini CLI would turn this warning into a hard failure.

## Reproduction

```bash
mkdir -p /tmp/gemini-test/.gemini/agents
cat > /tmp/gemini-test/.gemini/agents/test-agent.md << 'EOF'
---
name: test-metadata-agent
description: "Test agent."
type: agent
metadata:
  provenance:
    canonical: "https://github.com/crewrig/crewrig"
    feedback: "https://github.com/crewrig/crewrig"
    version: "1.0.0"
---
EOF
cd /tmp/gemini-test && GEMINI_CLI_TRUST_WORKSPACE=true gemini --skip-trust -p "hello"
```

**Expected:** clean startup, no validation error.
**Actual:** `Unrecognized key(s) in object: 'type', 'metadata'` printed on every invocation.

## Context

The build script (`scripts/build-components.sh`) already strips `type` and `metadata` when generating `.gemini/agents/*.md` from sources in `community-config/agents/*/AGENT.md`. The stripping is therefore correct — but the **source schema** and **built schema** are currently misaligned:

| File | `type` | `metadata` |
|---|---|---|
| `community-config/agents/*/AGENT.md` | ✅ present | ✅ present |
| `.gemini/agents/*.md` (built) | ❌ stripped | ❌ stripped |
| `.claude/agents/*.md` (built) | ✅ kept | ✅ kept |

The Claude Code built output retains both fields because Claude Code's frontmatter schema accepts them. Gemini CLI does not.

## Scope of analysis

The remediation strategy must address:

1. **Where does provenance live for Gemini agents?** Options include: embed in the body as a comment block, sidecar file (e.g. `<name>.meta.json`), or drop entirely for Gemini output.
2. **`type` field**: Is it meaningful for Gemini CLI at all? If not, the build script correctly discards it. Confirm and document.
3. **Version tracking**: The `metadata.provenance.version` bump convention (enforced by `scripts/check-skill-versions.sh`) currently diffs the source files. If provenance is moved out of the frontmatter for Gemini, verify the CI version check still works correctly.
4. **Friction attribution**: `harness-report` payloads use `canonical` and `feedback` URLs sourced from provenance. Assess impact if provenance is absent from the built Gemini output.
5. **Forward compatibility**: Evaluate whether upstream Gemini CLI might add support for custom frontmatter keys, which would change the calculus.

## Expected outcome

- A documented remediation strategy (ADR or inline decision in `docs/cli-matrix.md`).
- Updated `scripts/build-components.sh` if the strategy requires a different output format.
- Updated `docs/cli-matrix.md` row for agent provenance/metadata.
- No `Unrecognized key(s)` errors on Gemini CLI startup.

hcross:
Gemini CLI 0.42.0 rejects any frontmatter key outside `name` / `description`, so the injected `metadata:` block printed `Unrecognized key(s) in object: 'type', 'metadata'` on every agent invocation. This PR strips `metadata:` from `.gemini/agents/*.md` and re-emits provenance as a single-line HTML comment at the top of the body — Claude and Copilot outputs are unchanged.

## How to read this PR?

Two commits, read in order:

1. **`f5a8269` — build fix + regenerated outputs.** Start with `scripts/build-components.sh`: the new `gemini_provenance_comment()` helper (around line 127) renders the provenance as `<!-- crewrig-provenance: version="…" canonical="…" feedback="…" -->`, and the Gemini branch of `build_agents()` (around line 560) drops its `$source` argument to `check_or_write` so `inject_provenance` is intentionally bypassed. Then skim one regenerated agent (e.g. `.gemini/agents/architect.md`) to see the new shape — frontmatter is just `name` + `description`, and the first body line is the provenance comment. The 20 other regenerated files follow the same pattern. Finally, `docs/cli-matrix.md` adds row 4b documenting the per-CLI provenance carrier split.

2. **`4f10914` — regression test + CI wiring.** `scripts/tests/test-gemini-agent-frontmatter.sh` pins four properties of the build outputs (see *How to test* below). `.github/workflows/build.yml` runs it right after `test-check-skill-versions`.

Non-obvious decisions:

- **Provenance carrier is split by CLI**, not unified. Claude and Copilot still ship `metadata:` in YAML because their loaders accept it. Inventing a third shared format would force a churn across two healthy outputs to placate one stricter CLI.
- **HTML comment, not a fenced YAML body block.** The comment renders as nothing in Markdown viewers, stays greppable, and survives copy-paste; a body-level YAML fence would have shown up in rendered agent prompts.

## How to test this PR?

Prerequisites: a checkout of this branch, `bash` and `yq` on `PATH`.

```bash
# 1. Rebuild and confirm no drift against the committed outputs.
bash scripts/build-components.sh
git status --porcelain   # expected: empty

# 2. Run the new regression test directly.
bash scripts/tests/test-gemini-agent-frontmatter.sh
# Expected: "Result: 4 passed, 0 failed"

# 3. Spot-check the regenerated output.
head -6 .gemini/agents/architect.md
# Expected:
#   ---
#   name: architect
#   description: "..."
#   ---
#   <!-- crewrig-provenance: version="..." canonical="..." feedback="..." -->

# 4. Confirm siblings still carry metadata: (Gemini-only stripping).
grep -l '^metadata:' .claude/agents/*/AGENT.md .github/agents/*.md | wc -l
# Expected: 42  (21 Claude + 21 Copilot)

# 5. End-to-end: launch Gemini CLI in this checkout and trigger any agent.
#    Expected: no "Unrecognized key(s) in object" warning on stderr.
```

**Trigger / problem.** Gemini CLI 0.42.0 tightened its agent-frontmatter schema to accept only `name` and `description`. The `metadata:` block that `scripts/build-components.sh` injects (via `inject_provenance`) into every Gemini agent triggered an `Unrecognized key(s) in object: 'type', 'metadata'` warning on every CLI startup and on every agent invocation. Issue #54.

**Architect decision — Option A.** Of the three options considered (A: strip + body comment; B: unify all CLIs on body comment; C: gate by Gemini version), the architect chose **A**: minimum churn, no impact on Claude/Copilot, no version-detection logic, and provenance still travels with the artifact in a stable, greppable form.

**Implementation — commit `f5a8269` (developer).**

- `scripts/build-components.sh`:
  - Added `gemini_provenance_comment()` (≈20 lines). Reads `metadata.provenance.{version,canonical,feedback}` from the source frontmatter via `yq`; returns empty if no provenance is present; otherwise emits `<!-- crewrig-provenance: version="…" canonical="…" feedback="…" -->`.
  - In `build_agents()`, the Gemini branch now extracts the source frontmatter explicitly, computes the provenance comment, splices it as the first body line of the rendered output, and calls `check_or_write` **without** the `$source` argument — the bypass is annotated with a `NOTE:` comment so future readers don't reintroduce `inject_provenance` by reflex.
  - Claude (`.claude/agents/*/AGENT.md`) and Copilot (`.github/agents/*.md`) branches are untouched.
- `.gemini/agents/*.md` (21 files): regenerated. Frontmatter reduced to `name` + `description`; first body line is the new provenance comment.
- `docs/cli-matrix.md`: row 4b ("Agent provenance carrier") added, documenting that Gemini uses an HTML comment while Claude/Copilot use the YAML `metadata:` block. A GAP-confirmation line records that no Copilot-side action is needed.

**Implementation — commit `4f10914` (tester).**

- `scripts/tests/test-gemini-agent-frontmatter.sh` (new, 4 assertions):
  1. No `metadata:` or `type:` key in `.gemini/agents/*.md` frontmatter.
  2. Every Gemini agent body starts with a well-formed `<!-- crewrig-provenance: version="…" canonical="…" -->` line.
  3. (3-claude) Every `.claude/agents/*/AGENT.md` still contains a `metadata:` frontmatter key.
  4. (3-copilot) Every `.github/agents/*.md` still contains a `metadata:` frontmatter key.

  Locally: `4 passed, 0 failed`.

- `.github/workflows/build.yml`: new step "Test Gemini agent frontmatter contract (issue #54)" inserted after the `test-check-skill-versions` step in the `lint` job.

**What this PR does NOT do.**

- Does not modify Claude or Copilot agent outputs, nor their build branches.
- Does not change the source `community-config/agents/*/AGENT.md` frontmatter schema — provenance still lives there as YAML; only the Gemini-side rendering differs.
- Does not bump any `metadata.provenance.version` field: per AGENTS.md → *Version Bump Convention*, this PR modifies the build pipeline and its outputs, not the shipped source components themselves.

Closes #54.


hcross:
> **Cold review verdict from pr-reviewer agent — would APPROVE (GitHub blocks self-approval; posting as comment instead).**

## Verdict: ✅ Approve

Cold review of `fix/gemini-agent-metadata` → `main`. The fix is targeted, well-tested, and properly documented across the parity surface.

### Checklist results

1. **Build-script split (Gemini vs Claude/Copilot)** — ✅
   - `gemini_provenance_comment()` (lines 127–144) is a clean, side-effect-free helper that returns empty when `metadata.provenance` is absent (defensive `// {}` guard against missing `metadata`).
   - `build_agents()` Gemini branch correctly drops the `$source` arg on `check_or_write`, with a `NOTE:` comment guarding against future re-introduction of `inject_provenance`. The Claude (`.claude/agents/<name>/AGENT.md`) and Copilot (`.github/agents/<name>.md`) branches are untouched.

2. **Provenance comment in regenerated Gemini agents** — ✅
   - Spot-checked the diff for all 21 files: `metadata:` block stripped, first body line is `<!-- crewrig-provenance: version="…" canonical="…" feedback="…" -->`. Versions preserved from source (e.g. `architect` 1.1.0, `developer` 1.2.0, `harness-curator` 1.1.0).

3. **`.claude/agents/` and `.github/agents/` untouched** — ✅ Not in the changed-files list. Test (3-claude) / (3-copilot) pin this as a regression assertion going forward.

4. **`docs/cli-matrix.md`** — ✅
   - New row 4b accurately captures the per-CLI carrier split.
   - The `[GAP-confirmation]` entry for Copilot is the right call given the public reference's silence on tolerated frontmatter keys — surface the asymmetry, file a follow-up, don't pre-emptively churn a working output.

5. **Smoke test** — ✅
   - Four assertions cover the contract: no `metadata:`/`type:` in Gemini frontmatter, well-formed provenance comment as first body line, `metadata:` still present in both sibling targets. The shape-aware loop over Claude (nested) and Copilot (flat) layouts is correct.
   - Minor nit (non-blocking): assertion (2)'s glob pattern requires `version=` and `canonical=` but not `feedback=`. The helper always emits all three, so the test will fail loudly if the format degrades; tightening to require `feedback=` would be a strictly additive improvement.

6. **CI wiring** — ✅ Step inserted right after `test-check-skill-versions`, same `lint` job, `bash` invocation matches the executable bit set on the script. Clean.

7. **Convention compliance** — ✅
   - Gitmoji title (`🐛 …`), branch name (`fix/…`), PR body follows the AGENTS.md template (How to read / How to test / Detailed description for agents).
   - **Version bumps:** correctly skipped. AGENTS.md → *Version Bump Convention* targets modifications to shipped `community-config/skills/*/SKILL.md` and `community-config/agents/*/AGENT.md` sources — neither is touched here. The build pipeline and its regenerated outputs are not in the trigger surface.
   - **Built-components rule:** `.gemini/agents/*.md` regenerated in the same PR as the script change. No drift expected on the CI `check-components` job.
   - **CLI matrix maintenance rule:** trigger surface (`scripts/build-components.sh`, `.gemini/**`) hit → matrix updated in the same diff. ✅

### Architectural sanity-check

Option A (per-CLI carrier split) was the right call. The two alternatives carried real downsides: B (unify everyone on body comments) churns two healthy outputs to placate one stricter CLI; C (version gate) introduces a moving target inside the build pipeline. A keeps the blast radius minimal and the failure mode visible — if Copilot ever tightens like Gemini did, the `[GAP-confirmation]` entry is already there to convert into a real ticket.

### Nice touches

- The `NOTE:` comment on the `check_or_write` call without `$source` — future maintainers won't reflexively "fix" the missing argument.
- HTML comment over fenced YAML body block — invisible in renderers, greppable, survives copy-paste.
- Test isolates the build-pipeline contract from the source-content contract (idempotency / canonical URL are explicitly out of scope, covered elsewhere).

Nothing blocking. Ship it.

PR #74, opened 2026-05-22
✨ Offer MemPalace auto-install via pipx in setup scripts

hcross:
All three `setup-*-interactive` scripts warned users to install MemPalace manually and re-run — this PR adds an opt-in pipx install step so onboarding completes in a single run.

## How to read this PR?

Start with `scripts/lib/common.sh` to understand the new `offer_mempalace_install()` helper, then read the MemPalace section of each setup script (claude ~L190, gemini ~L138, copilot ~L199) to see the symmetric two-statement call pattern. `docs/cli-matrix.md` row 10 reflects the new capability.

## How to test this PR?

1. Ensure MemPalace is **not** installed (`pipx uninstall mempalace` if needed).
2. Run any of the three setup scripts: `bash scripts/setup-claude-interactive.sh`.
3. At the "MemPalace not found" prompt, select **yes** → MemPalace installs, script continues to MCP registration.
4. Re-run with MemPalace already installed → new offer step is skipped entirely, existing flow unchanged.
5. Repeat with `setup-gemini-interactive.sh` and `setup-copilot-interactive.sh`.

**`scripts/lib/common.sh`**: Added `offer_mempalace_install()` after `detect_mempalace_python()`. The function: (1) gates on `command -v pipx` — if absent, prints manual install + pipx bootstrap hints and returns 1; (2) prompts via fzf with `no` as the default choice; (3) wraps `pipx install "mempalace>=3.3.3,<3.4"` in `if !` to catch failures under `set -e`; (4) returns 0 on success. All failure paths return 1 so the caller's `|| true` absorbs them cleanly.

**Three setup scripts**: Replaced the bare `WARN + instruction` block with:
```bash
if [ -z "$MEMPALACE_PYTHON_BIN" ]; then
  echo "  MemPalace not found."
  offer_mempalace_install || true
  MEMPALACE_PYTHON_BIN="$(detect_mempalace_python || true)"
fi
```
The re-detection line immediately after the offer allows a successful install to feed directly into the existing version-check + MCP registration flow — no restart required. The claude script required restructuring the surrounding `if/else` into two `if` blocks to make this work correctly.

**`docs/cli-matrix.md`**: Row 10 description updated symmetrically across all three CLI columns to reflect the new capability. No parity gap — the fix is identical for all three CLIs.


hcross (commented):
## Cold review verdict: **Approve** ✅ (posted as comment — gh blocks self-approval)

### CI status
All 6 required checks green: `build`, `check-components`, `check-skill-versions`, `lint-markdown`, `test-harness-curate`, `grep-anti-patterns`.

### Correctness (`offer_mempalace_install`)
- All three failure paths return 1 cleanly; caller's `|| true` neutralises them under `set -e`.
- `command -v pipx` gate is the right precondition.
- fzf cancellation (Ctrl-C / ESC) leaves `$choice` empty → falls through the `!= "yes"` branch → returns 1. Safe.
- `if ! pipx install …` correctly catches install failure under `set -e` (a bare `pipx install` would have aborted the script).
- Re-detection via `MEMPALACE_PYTHON_BIN="$(detect_mempalace_python || true)"` after the offer feeds the existing version-check flow without restart. Nice continuity.

### Convention compliance
- Title: ✨ Gitmoji, concise, descriptive. ✓
- PR body: all four required sections present (purpose, How to read, How to test, Detailed description). ✓
- English-only across diff + PR body. ✓

### Symmetry
All three setup scripts apply the identical two-statement pattern (`offer_mempalace_install || true` + re-detect). Claude's surrounding `if/else → two if blocks` restructure is necessary and correctly executed — without it the version-check branch would not see a freshly installed MemPalace. Copilot and Gemini did not have an `else` branch to restructure, so the minimal 3-line patch is correct there.

### cli-matrix.md
Row 10 updated symmetrically across all three CLI columns with the same parenthetical "(offers MemPalace auto-install via pipx)". No parity gap — correctly not recorded as one.

### Regression surface
The new code path is gated on `[ -z "$MEMPALACE_PYTHON_BIN" ]` — when MemPalace is already installed, nothing in the existing flow changes. Bounded blast radius.

### Minor observations (non-blocking, optional)
- The fzf default `no` is a sound choice for an opt-in install — keeps the script non-surprising for power users.
- `pipx install "mempalace>=${MIN},<${MAX}"` correctly reuses the shared version constants — single source of truth preserved.

No blockers. Recommend merge.

Issue #94, opened 2026-05-25
🐛 mempalace-transcript hook: no timeout on Python call causes Stop hook loop


hcross:
## Problem

The hook invokes Python with no timeout. If MemPalace blocks on a lock, the hook hangs indefinitely. Claude Code's hook timeout fires → `Failed with non-blocking status code: No stderr output` → loop.

## Fix

Wrap the Python invocation with `timeout 5`:
```bash
STATUS=$(timeout 5 "$MEMPALACE_PYTHON" - 2>&1 <<'PYEOF' ...) || true
```

Discovered during epic #75 parallel agent sessions.

hcross:
Fixed in PR #95 (`fix/mempalace-transcript-hooks`). Regression test added: `scripts/tests/test-mempalace-transcript-hook.sh`.

Issue #93, opened 2026-05-25
🔧 mempalace-transcript hook: stderr silently swallowed, making crashes undiagnosable


hcross:
## Problem

```bash
STATUS=$( "$MEMPALACE_PYTHON" - 2>&1 <<'PYEOF' ...)
```

Both stdout and stderr are captured in `$STATUS`. Python crash messages are never surfaced — root-cause analysis requires guesswork.

## Fix

Redirect stderr to a dedicated log file:
```bash
STATUS=$( "$MEMPALACE_PYTHON" - 2>/tmp/mempalace-hook.err <<'PYEOF' ...)
```

Discovered during epic #75 crash investigation.

hcross:
Fixed in PR #95 (`fix/mempalace-transcript-hooks`). Regression test added: `scripts/tests/test-mempalace-transcript-hook.sh`.

Issue #92, opened 2026-05-25
🐛 mempalace-transcript hook: PROJECT_NAME wrong inside git worktrees


hcross:
## Problem

```bash
PROJECT_DIR="$(pwd)"          # → .worktrees/issue-88
PROJECT_NAME=$(basename "$PROJECT_DIR")  # → "issue-88" instead of "crewrig"
```

Transcripts land in the wrong MemPalace room when the hook runs from a worktree.

## Fix

Derive `PROJECT_DIR` from git root:
```bash
PROJECT_DIR=$(git rev-parse --show-toplevel 2>/dev/null || pwd)
```

Discovered during epic #75 worktree-based development.

hcross:
Fixed in PR #95 (`fix/mempalace-transcript-hooks`). Regression test added: `scripts/tests/test-mempalace-transcript-hook.sh`.

Issue #91, opened 2026-05-25
🐛 mempalace-transcript hook: fires on every PostToolUse — rate too high for parallel agents


hcross:
## Problem

Hook is registered on `PostToolUse`, `UserPromptSubmit`, `Stop`, and `SessionEnd`. With parallel agents doing intensive tool calls, this generates dozens of simultaneous MemPalace writes per second, amplifying contention.

## Fix

Restrict to `Stop` + `SessionEnd` only, or add a debounce/batch mechanism.

Discovered during epic #75 parallel agent sessions.

hcross:
Fixed in PR #95 (`fix/mempalace-transcript-hooks`). Regression test added: `scripts/tests/test-mempalace-transcript-hook.sh`.

PR #95, opened 2026-05-25
🐛 Fix mempalace-transcript hook: timeout, stderr, worktree path, rate limiting (#90-94)

### LINKED ISSUE #90: 🐛 mempalace-transcript hook: direct Python import causes SQLite contention with parallel agents
## Problem

The stop/post-tool hook calls `tool_add_drawer` via a direct Python import instead of going through the MCP protocol. With N parallel agents firing simultaneously, this causes SQLite file-lock contention, hangs, and SIGSEGV → macOS crash reporter loop.

```python
from mempalace.mcp_server import tool_add_drawer  # bypasses MCP session + file lock
```

## Fix

Replace the direct Python import with a proper MCP call, or add a file-level mutex around the direct call.

Discovered during epic #75 e2e testing sessions with parallel agent teams.

hcross:
Fixes five bugs in `hooks/mempalace-transcript.sh` that caused Python crash loops, macOS crash reporter popups, and incorrect MemPalace room naming when running parallel agent teams.

## How to read this PR?

Read `hooks/mempalace-transcript.sh` diff first — all changes are in this single file. Then read `scripts/tests/test-mempalace-transcript-hook.sh` for the regression suite that locks the fixes in.

## How to test this PR?

```sh
bash scripts/tests/test-mempalace-transcript-hook.sh
# Expected: Summary: 5 passed, 0 failed
```

**`hooks/mempalace-transcript.sh`:**
- `timeout 5` wrapper added before `"$MEMPALACE_PYTHON"` — prevents indefinite hang when MemPalace blocks on a lock (#90, #94)
- `set +e` / capture rc / `set -e` pattern used instead of `|| true` — preserves actual Python exit code for failure-branch logic
- Python stderr redirected to `${TMPDIR:-/tmp}/mempalace-hook-$$.err` with EXIT trap cleanup; cat-ed to stderr on failure (#93)
- Early `exit 0` inserted after `HOOK_EVENT` detection when event is `PostToolUse` — reduces write frequency 10× for parallel agent sessions (#91)
- `PROJECT_DIR` fallback chain extended: `git rev-parse --show-toplevel` used when neither `GEMINI_PROJECT_DIR` nor `CLAUDE_PROJECT_DIR` is set, preventing wrong room name in worktrees (#92)

**`scripts/tests/test-mempalace-transcript-hook.sh`** (new, 185 LOC):
- 5 regression tests, one per issue
- No live MemPalace server required
- Uses marker-file pattern for behavioral test (#91), static grep for structural tests (#90, #92, #93, #94)

Closes #90, #91, #92, #93, #94

### DIFF
--- a/hooks/mempalace-transcript.sh
+++ b/hooks/mempalace-transcript.sh
@@ -122,9 +122,9 @@ if [ -n "$CONTENT" ]; then
   # Capture Python stderr to a dedicated file so import/runtime errors are
   # visible instead of being swallowed into $STATUS (issue #93). The
   # `timeout 5` wrapper kills a hung Python after 5 seconds so a MemPalace
-  # lock cannot stall the calling CLI (issues #90, #94). `|| true` keeps
-  # `set -e` from aborting on Python exit codes — STATUS_RC carries the
-  # actual outcome.
+  # lock cannot stall the calling CLI (issues #90, #94). `set +e`/`set -e`
+  # brackets the call so a non-zero Python exit does not abort the hook —
+  # STATUS_RC carries the actual outcome.
   _HOOK_ERR="${TMPDIR:-/tmp}/mempalace-hook-$$.err"
   trap 'rm -f "$_HOOK_ERR"' EXIT
 
--- a/hooks/mempalace-transcript.sh
+++ b/hooks/mempalace-transcript.sh
@@ -42,13 +42,25 @@ COPILOT_PROJECT_DIR_FROM_JSON=$(echo "$INPUT" | jq -r '.workspace_dir // .worksp
 COPILOT_SESSION_ID_FROM_JSON=$(echo "$INPUT" | jq -r '.session_id // .sessionId // empty' 2>/dev/null)
 
 SESSION_ID="${GEMINI_SESSION_ID:-${CLAUDE_SESSION_ID:-${COPILOT_SESSION_ID:-${COPILOT_SESSION_ID_FROM_JSON:-unknown}}}}"
-PROJECT_DIR="${GEMINI_PROJECT_DIR:-${CLAUDE_PROJECT_DIR:-${COPILOT_PROJECT_DIR:-${COPILOT_PROJECT_DIR_FROM_JSON:-$(pwd)}}}}"
+# Resolve project dir from the git root so that worktree paths collapse to
+# the canonical repository root (issue #92). Fallbacks: env vars from each
+# CLI, JSON-embedded project dir from Copilot, then $PWD as last resort.
+_GIT_ROOT=$(git rev-parse --show-toplevel 2>/dev/null || true)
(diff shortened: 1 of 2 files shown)


hcross (commented):
## Independent review — verdict: would Approve (cannot self-approve)

Cold review of the diff against the five issues (#90–#94). All CI checks green.

### Correctness
- **#90 / #94 — `timeout 5`**: correctly positioned as `timeout 5 "$MEMPALACE_PYTHON" - 2>"$_HOOK_ERR" <<'PYEOF'`. Wraps the actual Python invocation; rc surfaces via `STATUS_RC`.
- **rc handling**: uses `set +e` / capture / `set -e` (not `|| true`), so the real exit code (including 124 from `timeout`) reaches `STATUS_RC` and the failure branch.
- **#91 — PostToolUse skip**: exact-match guard `[ "$HOOK_EVENT" = "PostToolUse" ]`. `Stop` and `SessionEnd` are untouched and still hit the Python writer.
- **#92 — worktree-safe `PROJECT_DIR`**: `git rev-parse --show-toplevel` is added to the fallback chain *after* the CLI env vars (`GEMINI_PROJECT_DIR`, `CLAUDE_PROJECT_DIR`, `COPILOT_PROJECT_DIR`, `COPILOT_PROJECT_DIR_FROM_JSON`). Env vars still win when set; `$(pwd)` remains as last resort.
- **#93 — stderr separation**: `2>"$_HOOK_ERR"` (no `2>&1`), with `_HOOK_ERR="${TMPDIR:-/tmp}/mempalace-hook-$$.err"` and `trap 'rm -f "$_HOOK_ERR"' EXIT`. PID-scoped, cleaned reliably.

### Safety
- `set -euo pipefail` remains active; the `set +e` window is tight (only around the Python call) and immediately restored.
- `timeout` does not swallow Python's own non-zero exit — it just adds 124 as another failure code; the existing failure branch handles it uniformly.
- Temp file path is in `$TMPDIR` with `$$`; not predictable in a hostile sense.

### Tests
- `scripts/tests/test-mempalace-transcript-hook.sh` ships 5 tests, one per issue.
- #91 uses the marker-file pattern (fake `MEMPALACE_PYTHON` that `touch`es a file) — no live MemPalace required, deterministic.
- Tests 1 + 5 together pin both presence and canonical shape of the `timeout` wrapper.

### Conventions
- Gitmoji title (🐛), English throughout, no `community-config/` touched (no version bump needed), no new `.md` files (markdownlint N/A).

### Non-blocking nit
- The inline comment above the Python call says *"The `|| true` keeps `set -e` from aborting on Python exit codes"*, but the code actually uses the `set +e` / `set -e` pattern. The comment is stale relative to the implementation — worth a one-line edit on the next pass, but does not affect behaviour.

Recommendation: **Approve and merge.**

PR #113, opened 2026-05-26
🐛 Persist .claude.json across e2e auth + scenario runs (#112)

### LINKED ISSUE #112: 🐛 e2e auth: .claude.json not persisted — home-dir write lands outside the mounted volume
## Context

\`task e2e:auth:claude\` fails at the post-flight check with:

\`\`\`
[claude] WARNING: expected credential file(s) not found in ~/.crewrig-e2e/claude:
  - .claude.json
\`\`\`

The OAuth flow itself completes correctly (\`.credentials.json\` is present).

## Root cause

\`auth-claude.sh\` mounts \`${DIR}\` on \`/home/agent/.claude\` (the hidden directory).
Claude Code writes two files at different paths:

| File | Path in container | In volume? |
|---|---|---|
| \`.credentials.json\` | \`/home/agent/.claude/.credentials.json\` | ✅ persisted |
| \`.claude.json\` | \`/home/agent/.claude.json\` (home root) | ❌ outside mount, lost on exit |

## Fix

Two files need updating:

**1. \`scripts/e2e/auth-claude.sh\`** — pre-create the file and add a second \`-v\` binding:

\`\`\`bash
touch "${DIR}/.claude.json"
docker run --rm -it \
  -v "${DIR}:/home/agent/.claude" \
  -v "${DIR}/.claude.json:/home/agent/.claude.json" \
  "$IMAGE" claude /login
\`\`\`

**2. \`tests/e2e/defaults.toml\`** — same bind in RO for scenario runs:

\`\`\`toml
[cli.claude]
mounts = [
  "${CREWRIG_E2E_HOME}/claude:/home/agent/.claude:ro",
  "${CREWRIG_E2E_HOME}/claude/.claude.json:/home/agent/.claude.json:ro"
]
\`\`\`

## Acceptance criteria

- \`task e2e:auth:claude\` exits 0 after a successful OAuth flow.
- \`.claude.json\` is present in \`~/.crewrig-e2e/claude/\` after the auth run.
- Scenario runs receive \`.claude.json\` at \`/home/agent/.claude.json\` inside the container.

hcross:
Closes #112.

Claude Code writes session state to two paths — `~/.claude/.credentials.json` and `~/.claude.json` — but `scripts/e2e/auth-claude.sh` only mounted the directory, so the home-root file landed on the container overlay and disappeared on exit. This PR mounts both paths in the auth flow and in scenario runs, and locks the contract with a static regression test.

## How to read this PR?

1. `scripts/e2e/auth-claude.sh` — two-line change: `touch` the host-side `.claude.json` before `docker run` (so Docker creates a file bind, not a directory bind), then add a second `-v` flag for it.
2. `tests/e2e/defaults.toml` — `[cli.claude].mounts` gains the symmetric `:ro` entry so non-interactive scenario runs surface the persisted file inside the container.
3. `scripts/tests/test-e2e-auth-claude-json.sh` — new static regression test. Parses the script with `grep` and the TOML with `yq | jq`; does not execute docker. Two assertions, plus existence checks.

## How to test this PR?

```bash
bash scripts/tests/test-e2e-auth-claude-json.sh
```

Expected: `Results: 4 passed, 0 failed`, exit code 0.

To verify the fix end-to-end (optional, requires Docker + interactive auth):

```bash
task e2e:build:claude
bash scripts/e2e/auth-claude.sh
ls -la "${CREWRIG_E2E_HOME:-$HOME/.cache/crewrig/e2e}/claude/.claude.json"  # must exist and be non-empty
```

- **`scripts/e2e/auth-claude.sh`**
  - Added `touch "${DIR}/.claude.json"` after `mkdir -p "$DIR"`. Required because Docker creates a directory at the bind source if the path does not exist, which would then collide with Claude's attempt to write a regular file.
  - Added `-v "${DIR}/.claude.json:/home/agent/.claude.json"` to the `docker run` invocation, alongside the existing `-v "${DIR}:/home/agent/.${CLI}"` mount.
- **`tests/e2e/defaults.toml`**
  - `[cli.claude].mounts` expanded from a single-entry inline array to a two-entry list: the existing `${CREWRIG_E2E_HOME}/claude:/home/agent/.claude:ro`, plus `${CREWRIG_E2E_HOME}/claude/.claude.json:/home/agent/.claude.json:ro`. Read-only mount because scenario runs must not mutate persisted auth state.
- **`scripts/tests/test-e2e-auth-claude-json.sh`** (new, executable)
  - Static-only checks; no Docker dependency.
  - Test 1: greps `auth-claude.sh` for a `-v` flag binding any host path ending in `.claude.json` to `/home/agent/.claude.json`.
  - Test 2: parses `defaults.toml` with `yq -p=toml -o=json` and asserts that `.cli.claude.mounts` contains at least one entry whose container path is `/home/agent/.claude.json` (the `:ro` suffix is tolerated but not required).
  - Both ends of the contract are locked so a future refactor of either file cannot silently regress the other.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,6 +50,9 @@ jobs:
       - name: Test Gemini agent frontmatter contract (issue #54)
         run: bash scripts/tests/test-gemini-agent-frontmatter.sh
 
+      - name: Test e2e auth .claude.json mount (issue #112)
+        run: bash scripts/tests/test-e2e-auth-claude-json.sh
+
   lint-markdown:
     runs-on: ubuntu-latest
     steps:
--- a/scripts/e2e/auth-claude.sh
+++ b/scripts/e2e/auth-claude.sh
@@ -27,6 +27,7 @@ e2e_require_image "$IMAGE" "task e2e:build:claude"
 
 e2e_info "[$CLI] Preparing ${DIR}…"
 mkdir -p "$DIR"
+touch "${DIR}/.claude.json"
 
 # Decision 6: chown bootstrap is mandatory (macOS) and harmless (Linux).
 e2e_chown_bootstrap "$CLI" "$IMAGE"
@@ -39,6 +40,7 @@ e2e_info "[$CLI] (Press Ctrl-D inside the prompt when Claude reports a successfu
 # so the OAuth URL is rendered properly and the user can paste the redirect.
 docker run --rm -it \
   -v "${DIR}:/home/agent/.${CLI}" \
+  -v "${DIR}/.claude.json:/home/agent/.claude.json" \
   "$IMAGE" \
   claude /login \
   || e2e_die "[$CLI] interactive container exited non-zero. Re-run after resolving the error above."
--- a/scripts/tests/test-e2e-auth-claude-json.sh
+++ b/scripts/tests/test-e2e-auth-claude-json.sh
@@ -35,52 +35,53 @@ DEFAULTS="${REPO_DIR}/tests/e2e/defaults.toml"
 # Test 1 — auth-claude.sh mounts .claude.json into the container
 # ---------------------------------------------------------------------------
 if [[ ! -f "$AUTH_SCRIPT" ]]; then
-  note_fail "auth-claude.sh exists" "missing at $AUTH_SCRIPT"
-else
-  note_pass "auth-claude.sh exists"
+  echo "SKIP  auth-claude.sh not found at $AUTH_SCRIPT — cannot test"
+  exit 1
+fi
 
-  # The script must contain a -v flag that binds a host path ending in
-  # `.claude.json` to `/home/agent/.claude.json`. We grep with a tolerant
-  # regex so either an inline literal or a `${VAR}/.claude.json` form passes.
-  if grep -Eq -- '-v[[:space:]]+"?[^"[:space:]]*\.claude\.json:/home/agent/\.claude\.json' "$AUTH_SCRIPT"; then
-    note_pass "auth-claude.sh — docker run binds .claude.json into /home/agent/.claude.json"
-  else
-    note_fail "auth-claude.sh — docker run binds .claude.json into /home/agent/.claude.json" \
-      "no '-v <host>/.claude.json:/home/agent/.claude.json' flag found in $AUTH_SCRIPT (issue #112)"
-  fi
+# The script must contain a -v flag that binds a host path ending in
+# `.claude.json` to `/home/agent/.claude.json`. We grep with a tolerant
+# regex so either an inline literal or a `${VAR}/.claude.json` form passes.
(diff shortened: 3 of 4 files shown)


hcross:
## Verdict: APPROVE

Independent cold review — diff is tight, scoped, and well-tested.

### Correctness

- **`touch "${DIR}/.claude.json"` placement** ✅ — placed after `mkdir -p "$DIR"` and before `docker run`. Necessary because Docker auto-creates a **directory** at the bind source when the path doesn't exist, which would then collide with Claude's attempt to write a regular JSON file there. Order is correct.
- **Second `-v` flag** ✅ — binds the host file at `/home/agent/.claude.json` (HOME root, one level above `~/.claude/`), which matches the documented upstream behaviour (`tfvchow/field-notes-public#10`, already cited in the script's post-flight check). The existing `.credentials.json` check on line ~48 confirms the bind target geometry is right.
- **`defaults.toml` style** ✅ — uses the same `${CREWRIG_E2E_HOME}` token and `:ro` suffix as the existing entry. Inline-array → multi-line list transformation preserves TOML semantics. Read-only is the right choice for scenario runs (auth state must not mutate).

### Regression test

Verified locally by reverting both files to `main`:

```
PASS  auth-claude.sh exists
FAIL  auth-claude.sh — docker run binds .claude.json …
PASS  defaults.toml exists
FAIL  [cli.claude].mounts — entry targeting /home/agent/.claude.json present
Results: 2 passed, 2 failed
```

Restoring the fix → `4 passed, 0 failed`. The test genuinely locks both ends of the contract; static-only (no Docker dependency) keeps it cheap to run in CI.

### Sibling auth scripts

- `auth-gemini.sh` writes `oauth_creds.json` and `settings.json` — **both inside `.gemini/`**, covered by the existing directory mount. No symmetric issue.
- `auth-copilot.sh` is env-var only (ADR 0002 Decision 4), no on-disk creds. N/A.

No parity gap introduced.

### CI

All 6 checks green (`build`, `check-components`, `check-skill-versions`, `lint-markdown`, `grep-anti-patterns`, `test-harness-curate`). However, **the new `scripts/tests/test-e2e-auth-claude-json.sh` doesn't appear to be wired into any CI workflow** — it relies on a developer running it manually. Worth confirming whether that's intentional or whether it should be added to the test-harness job. **Non-blocking** for this PR, but flagging for the user's call.

### Nits (non-blocking)

- The grep regex in Test 1 is not anchored to non-comment lines — a `-v` flag inside a comment would also satisfy it. Low practical risk given the script's brevity.
- Test 1's `auth-claude.sh exists` / Test 2's `defaults.toml exists` count toward the `4 passed` total, inflating the assertion count visually. Cosmetic.

### Ship it

Approve. Clean, minimal, properly tested, conventions respected, version-bump rule doesn't apply (no skill/agent source modified).

PR #141, opened 2026-05-27
🐛 Poll chroma daemon health on launchd load (issue #138)

### LINKED ISSUE #138: 🐛 setup-claude-interactive fails: ChromaDB health check races launchd boot
## Symptom

Fresh `task setup-claude-interactive` aborts with:

```
chroma server: NOT RUNNING (no PID file, heartbeat failed at 127.0.0.1:8001)
  ERROR: ChromaDB daemon did not become healthy.
         Inspect logs at ~/.mempalace/chroma-server.log and retry.
task: Failed to run task "setup-claude-interactive": exit status 1
```

## Root cause

`scripts/lib/common.sh:172-177` calls `scripts/status-chroma-server.sh` **once, immediately** after `launchctl load -w …com.mempalace.chroma-server.plist`. The launchd agent has not finished booting Chroma yet — Chroma needs a few seconds to bind `127.0.0.1:8001`. The status script returns non-zero, the supervisor aborts.

A few seconds later the daemon is fully healthy (`curl http://127.0.0.1:8001/api/v2/heartbeat` returns 200) and rerunning the task succeeds because the `launchctl list | grep` branch skips the load step.

This is a startup race in `common.sh`, not a daemon defect.

## Fix direction

Mirror the polling logic already present in `scripts/start-chroma-server.sh` (15s budget, ~0.3s interval) around the health check at `common.sh:172`.

## Repro

1. `launchctl unload ~/Library/LaunchAgents/com.mempalace.chroma-server.plist` (if loaded)
2. `task setup-claude-interactive`
3. Observe ERROR on first invocation; rerun succeeds.

gfourny-sfeir:
Fix a race where `install_chroma_daemon` ran a one-shot health check immediately after `launchctl load`, before the Chroma daemon had bound `127.0.0.1:8001`. The health check now polls with a 15 s deadline, mirroring the existing pattern in `scripts/start-chroma-server.sh`.

## How to read this PR?

Two commits, intended to be read in order:

1. `f6f2fd5` — `scripts/tests/test-chroma-health-race.sh`: failing regression test that locks in the polling contract. Read this first to understand the expected behaviour.
2. `557ba81` — `scripts/lib/common.sh`: replaces the one-shot `curl` with a deadline-based loop (0.3 s interval, 15 s budget). Pattern is copied from `scripts/start-chroma-server.sh:65-80` for consistency.

Closes #138.

## How to test this PR?

Prerequisites: macOS with `bash` and the repository checked out at this branch.

1. Run the regression test:
   ```sh
   bash scripts/tests/test-chroma-health-race.sh
   ```
   Expected: exits 0, prints success markers for the polling path.
2. Confirm the test catches the original bug by checking out `origin/main` and re-running — it must fail.
3. Optional end-to-end: trigger `install_chroma_daemon` on a machine where Chroma is configured; the function should no longer report a spurious health-check failure on cold launch.

**`scripts/lib/common.sh`** — `install_chroma_daemon`:

- Removed the immediate single `curl` health probe issued right after `launchctl load`.
- Added a polling loop with a deadline (`SECONDS + 15`) and a 0.3 s `sleep` between attempts. The loop exits successfully on the first 2xx response from the Chroma `/api/v1/heartbeat` endpoint, or fails with a clear diagnostic once the deadline expires.
- Behaviour is now symmetric with `scripts/start-chroma-server.sh:65-80`, which already used this exact polling shape for the foreground startup path.

**`scripts/tests/test-chroma-health-race.sh`** — new file:

- Sources `scripts/lib/common.sh` and stubs `launchctl` plus the Chroma endpoint to simulate a daemon that takes ~2 s to come up.
- Asserts that `install_chroma_daemon` waits and ultimately reports success, rather than failing on the first probe.
- Also asserts the deadline path: when the stub never becomes ready, the function fails within the 15 s budget with a non-zero exit code.
- Fails against the pre-fix code in `scripts/lib/common.sh`, locking in the regression.

No other files touched; no public contracts changed.

### DIFF
--- a/scripts/lib/common.sh
+++ b/scripts/lib/common.sh
@@ -173,13 +173,18 @@ install_chroma_daemon() {
   # instead of one-shotting status-chroma-server.sh (see issue #138).
   if [ -x "$repo_dir/scripts/status-chroma-server.sh" ]; then
     local deadline=$((SECONDS + 15))
+    local healthy=0
     while [ "$SECONDS" -lt "$deadline" ]; do
       if bash "$repo_dir/scripts/status-chroma-server.sh" >/dev/null 2>&1; then
+        healthy=1
         break
       fi
       sleep 0.3
     done
-    if ! bash "$repo_dir/scripts/status-chroma-server.sh" >/dev/null 2>&1; then
+    if [ "$healthy" -ne 1 ]; then
+      # Surface the status script's diagnostics (stdout + stderr) so the user
+      # sees the real failure cause before the generic ERROR line.
+      bash "$repo_dir/scripts/status-chroma-server.sh" || true
       echo "  ERROR: ChromaDB daemon did not become healthy."
       echo "         Inspect logs at ~/.mempalace/chroma-server.log and retry."
       return 1
--- a/scripts/lib/common.sh
+++ b/scripts/lib/common.sh
@@ -168,9 +168,18 @@ install_chroma_daemon() {
       ;;
   esac
   # Health check — confirm the daemon answers on the heartbeat endpoint
-  # before any MCP entry is written.
+  # before any MCP entry is written. The launchd/systemd-managed process
+  # needs a few seconds to bind its socket, so poll with a 15s budget
+  # instead of one-shotting status-chroma-server.sh (see issue #138).
   if [ -x "$repo_dir/scripts/status-chroma-server.sh" ]; then
-    if ! bash "$repo_dir/scripts/status-chroma-server.sh"; then
+    local deadline=$((SECONDS + 15))
+    while [ "$SECONDS" -lt "$deadline" ]; do
+      if bash "$repo_dir/scripts/status-chroma-server.sh" >/dev/null 2>&1; then
+        break
+      fi
+      sleep 0.3
+    done
+    if ! bash "$repo_dir/scripts/status-chroma-server.sh" >/dev/null 2>&1; then
       echo "  ERROR: ChromaDB daemon did not become healthy."
       echo "         Inspect logs at ~/.mempalace/chroma-server.log and retry."
       return 1
--- a/scripts/tests/test-chroma-health-race.sh
+++ b/scripts/tests/test-chroma-health-race.sh
@@ -1,0 +1,168 @@
+#!/usr/bin/env bash
+# test-chroma-health-race.sh — Regression test for issue #138.
+#
+# The health-check block at the tail of `install_chroma_daemon`
+# (scripts/lib/common.sh, around lines 170-180) runs a SINGLE
+# invocation of `scripts/status-chroma-server.sh` immediately after
+# launchctl/systemctl has loaded the daemon unit. The daemon needs a
+# few seconds to bind 127.0.0.1:8001, so the one-shot check almost
+# always fails on first run; rerunning the setup succeeds because by
+# then the daemon is up.
+#
+# This test locks in the contract that the health check MUST poll the
+# status script (mirroring the 15s / 0.3s pattern of
+# `scripts/start-chroma-server.sh` lines 65-80) instead of relying on
+# a single invocation.
+#
+# Strategy
+# --------
+# 1. Stand up a sandbox `$repo_dir` containing a stub
+#    `scripts/status-chroma-server.sh` that returns non-zero for the
+#    first ~2 seconds (using a timestamp marker file) and exits 0
+#    afterwards — a deterministic simulation of the real race.
+# 2. Extract the health-check block from the live `common.sh` source
+#    (between the `# Health check` marker and the `else` branch of
+#    the surrounding `if [ -x ... ]`). This guarantees the test
+#    exercises the actual shipped logic, not a paraphrase.
+# 3. Execute the extracted block with `repo_dir` pointing at the
+#    sandbox. Assert exit code 0.
+#
+# Expected behaviour:
+#   - Against current main (one-shot check) → block exits 1. FAIL.
+#   - After the developer adds a polling loop                → block exits 0. PASS.
+
+set -uo pipefail
+
+PASS=0
+FAIL=0
+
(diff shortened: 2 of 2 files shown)


gfourny-sfeir:
## Verdict: APPROVE

Independent cold review per AGENTS.md Template 3.

### CI status
All required checks pass: build, check-components, check-skill-versions, grep-anti-patterns, lint-markdown, test-harness-curate.

### Correctness
- `scripts/lib/common.sh:174-186` — polling loop matches the PR body's claim: `deadline=$((SECONDS + 15))`, `sleep 0.3` between attempts, breaks on first success. Symmetric with `scripts/start-chroma-server.sh:65-80` as advertised.
- `scripts/tests/test-chroma-health-race.sh` — 168 lines, two assertions (behavioural slow-start tolerance + static polling-construct grep guard). Ran locally: 2 passed, 0 failed.

### Convention compliance
- Gitmoji title (🐛). PR body has all four required sections. English-only. `Closes #138` present. No `community-config/**` touched — version-bump rule not applicable.

### Test coverage
The regression test extracts the live health-check block from `common.sh` via awk and executes it against a stub `status-chroma-server.sh` that fails for ~2 s then succeeds. Locks the contract against future regression to a one-shot probe.

### Non-blocking findings (per AGENTS.md Team Communication Rule 4 — the user decides whether to address)
1. **Redundant final probe.** `common.sh:182` re-invokes `status-chroma-server.sh` after the loop exited on success. A success flag set inside the loop would avoid the duplicate call. Cost is one extra fast probe — cosmetic only.
2. **Silenced diagnostics on failure.** Both probe invocations use `>/dev/null 2>&1`. When the daemon truly fails to come up, the operator now sees only the generic "did not become healthy" message; the stderr from `status-chroma-server.sh` (which previously surfaced) is gone. Consider running the post-loop probe without `2>&1` so failure diagnostics reach the user. The log-file pointer in the error message partially compensates.
3. **Deadline uses `SECONDS` without resetting.** If `install_chroma_daemon` is called late in a long-lived shell session where `SECONDS` is already large, the arithmetic is still correct (it's relative), but a reader unfamiliar with `SECONDS` semantics might second-guess it. A short inline comment would help.

None of the above block merge.

hcross:
LGTM !

gfourny-sfeir:
## Reviewer findings — addressed

Follow-up to the cold review on `557ba81`. Commit `1225703` applies findings #1 and #2; #3 was skipped intentionally.

| # | Finding | Status |
|---|---------|--------|
| 1 | Redundant final probe after the loop already succeeded. | **Fixed** — introduced a `healthy` flag set inside the loop; the post-loop branch checks the flag instead of probing again. |
| 2 | `>/dev/null 2>&1` on every probe swallowed the status script's diagnostics on real failure. | **Fixed** — on the timeout path, the status script is invoked once more without redirection so its self-describing output ("chroma server: NOT RUNNING …") reaches the operator before the generic ERROR line. |
| 3 | One-line comment explaining bash `SECONDS` semantics. | **Skipped** — `SECONDS` is standard bash and the surrounding comment already states the polling intent. |

Regression test still passes (2/2) on the updated diff. CI green on all 6 required checks.

hcross:
## CI status

| Check | Status |
|---|---|
| build | pass |
| check-components | pass |
| grep-anti-patterns | pass |
| lint-markdown | pass |

All required checks green. ✅

## Correctness

The fix in `scripts/lib/common.sh:172-180` replaces a one-shot `status-chroma-server.sh` invocation with a polling loop (`while` / `SECONDS` / `sleep 0.3` / 15 s deadline). This mirrors the existing pattern at `scripts/start-chroma-server.sh:65-80` exactly — correct approach given the launchd boot race described in #138.

The extra call to `status-chroma-server.sh` on the failure path (`bash "$repo_dir/scripts/status-chroma-server.sh" || true`) is a nice touch — it surfaces the real diagnostics before the generic ERROR line, so the user sees what actually went wrong.

Linked issue #138 is correctly diagnosed and the fix addresses the root cause, not just the symptom.

## Convention compliance

- **Gitmoji**: `🐛` prefix on PR title ✅
- **Logbook**: Issue #138 serves as its own logbook per Rule A of AGENTS.md ✅
- **Version bump**: No skill or agent sources touched — exempt ✅
- **PR body sections**: "How to read", "How to test", "Detailed description" all present and substantive ✅
- **Branch name**: `fix/chroma-health-race` is descriptive ✅
- **Bug fix template** (Template 3): tester writes failing regression test first → developer implements fix → this maps cleanly onto the two-commit structure ✅

## Test coverage

`scripts/tests/test-chroma-health-race.sh` is a well-constructed regression test:

- **Test 1 (behavioural)**: Creates a sandbox with a stub that fails for 2 seconds then succeeds. Extracts the live health-check block from `common.sh` via `awk`, wraps it in a function, and asserts exit code 0. This exercises the shipping code, not a duplicate.
- **Test 2 (static guard)**: Greps the extracted block for polling constructs (`while`/`until`/`for`/`deadline=`/`SECONDS`) to prevent a future one-shot regression. Good belt-and-suspenders.

Both assertions fail against the pre-fix `common.sh`, confirming the test locks in the regression. The sandbox stub design (timestamp-based, not sleep-based) is deterministic and fast — no flaky timing dependency.

One minor note: the first `trap` at line 75 is overridden by the second at line 119. Not a bug since the second covers `$SANDBOX` too, but could be consolidated into a single trap. No action needed.

## Linter findings

All CI linters pass on the PR head. Local shellcheck on the two changed files:

- `test-chroma-health-race.sh` — clean (0 findings).
- `common.sh` — two pre-existing SC2015 (info) warnings at lines 146 and 162 (`A && B || C` pattern), plus a pre-existing missing `set -e` advisory. None were introduced by this PR.

## Verdict: APPROVE

No blocking issues. The fix is minimal, correct, well-tested, and symmetric with the existing polling pattern in `start-chroma-server.sh`.

hcross (approved):

Issue #142, opened 2026-05-27
🐛 e2e: e2e_gemini_refresh_access_token() fails — oauth_creds.json has no client_id/client_secret


hcross:
## Summary

\`e2e_gemini_refresh_access_token()\` introduced in PR #140 (dab157c) calls \`e2e_die\` at runtime because it expects \`client_id\` and \`client_secret\` in \`oauth_creds.json\`, but the Gemini CLI only stores \`{access_token, expiry_date, id_token, refresh_token, scope, token_type}\` — the client credentials are embedded in the CLI bundle itself.

## Symptom

```
ERROR: [runner] Gemini token refresh failed: ERROR: [gemini] oauth_creds.json missing client_id, client_secret, or refresh_token
```

## Root cause

\`oauth_creds.json\` format produced by Gemini CLI does not include \`client_id\`/\`client_secret\` (Google OAuth installed-app credential files produced by desktop clients do include them, but the Gemini CLI credential file does not). The client credentials are embedded in the bundle:

- \`OAUTH_CLIENT_ID\` = \`681255809395-oo8ft2oprdrnp9e3aqf6av3hmdib135j.apps.googleusercontent.com\`
- \`OAUTH_CLIENT_SECRET\` = \`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\`

## Fix

Redesign \`e2e_gemini_refresh_access_token()\` to:

1. Read \`access_token\` + \`expiry_date\` (ms epoch) from \`oauth_creds.json\`.
2. If still valid (≥5 min remaining) → return \`access_token\` directly.
3. If expired → use \`refresh_token\` + embedded client credentials to call \`https://oauth2.googleapis.com/token\`.
4. If refresh fails → die with a message asking the user to re-run \`task e2e:auth:gemini\`.

## Regression

\`tests/e2e/lib/test-gemini-headless.sh\` assertion 3 (\`answer.txt written\`) still fails after PR #140 because token refresh dies before the container is invoked.

hcross:
## Logbook — fix landed in PR #143

**Decision.** Redesigned `e2e_gemini_refresh_access_token()` around the actual Gemini CLI credential format:
- Fast path: when `expiry_date - 300000 > now_ms` (≥5 min remaining), echo `access_token` directly and return — no network call.
- Refresh path: `refresh_token` exchange against `https://oauth2.googleapis.com/token` using the Gemini CLI's embedded client credentials.

[...]

PR #143, opened 2026-05-27
🐛 Fix e2e_gemini_refresh_access_token to use existing token when valid (issue #142)

hcross:
Fixes the `e2e_gemini_refresh_access_token()` helper introduced in PR #140 which died immediately because it expected `client_id`/`client_secret` in `oauth_creds.json` — fields the Gemini CLI never writes there. The function now uses the stored `access_token` directly when valid, and falls back to the Google token endpoint with the CLI's embedded client credentials when the token is expired.

## How to read this PR?

Start with `scripts/e2e/lib/auth-common.sh`: the redesigned function (lines 84–117) is the sole change to production code. Then read `tests/e2e/lib/test-token-refresh.sh`: a standalone TAP unit test that mocks `oauth_creds.json` and asserts the valid-token fast path works without any network call.

## How to test this PR?

```bash
# Unit test (no docker, no network, no credentials needed)
bash tests/e2e/lib/test-token-refresh.sh
# Expected: ok 1 - valid token (>=5 min remaining) is returned directly without refresh

# Full integration (requires Gemini OAuth credentials)
task e2e:test:scenario -- 01-layered-context
```

**`scripts/e2e/lib/auth-common.sh`** — `e2e_gemini_refresh_access_token()`:
- Removed: extraction of `client_id` and `client_secret` from the creds file (those fields do not exist in the Gemini CLI credential format).
- Added: validity check — reads `expiry_date` (ms epoch) and `access_token`; if `expiry_date - 300000 > now_ms` (≥5 min remaining), echoes `access_token` and returns 0 with no network call.
- Kept: refresh path using `refresh_token`, now with hardcoded Gemini CLI client credentials (`OAUTH_CLIENT_ID`/`OAUTH_CLIENT_SECRET` from `chunk-W2MTI7FK.js` in the CLI bundle). Uses `$(date +%s) * 1000` for portable millisecond epoch (avoids GNU-only `%3N`).
- On any failure: `e2e_die` with message directing the user to re-run `task e2e:auth:gemini`.

**`tests/e2e/lib/test-token-refresh.sh`** (new file):
- TAP runner, standalone, no docker, no network.
- Test A: builds a mock `oauth_creds.json` shaped exactly like what the Gemini CLI writes (no `client_id`/`client_secret`), sets `expiry_date` to `now + 1h`, calls the function in a subshell, asserts stdout == `test-token-abc` and exit == 0.
- Was confirmed failing before the fix, green after.

### DIFF
--- a/scripts/e2e/lib/auth-common.sh
+++ b/scripts/e2e/lib/auth-common.sh
@@ -104,7 +104,7 @@ e2e_gemini_refresh_access_token() {
   local refresh_token response
   refresh_token=$(jq -r '.refresh_token // empty' "$creds_file")
   [[ -n "$refresh_token" ]] \
-    || e2e_die "[gemini] access token expired and refresh failed — re-run: task e2e:auth:gemini"
+    || e2e_die "[gemini] oauth_creds.json has no refresh_token — re-run: task e2e:auth:gemini"
   response=$(curl -s -X POST https://oauth2.googleapis.com/token \
     --data-urlencode "client_id=681255809395-oo8ft2oprdrnp9e3aqf6av3hmdib135j.apps.googleusercontent.com" \
     --data-urlencode "client_secret=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" \
--- a/scripts/e2e/lib/auth-common.sh
+++ b/scripts/e2e/lib/auth-common.sh
@@ -81,29 +81,38 @@ e2e_chown_bootstrap() {
     || e2e_die "[$cli] writability bootstrap failed — cannot chmod ${dir}."
 }
 
-# e2e_gemini_refresh_access_token [creds_file] — refresh the Gemini OAuth access
-# token from oauth_creds.json using the Google token endpoint. Prints the
-# access_token to stdout or calls e2e_die on failure.
-# oauth_creds.json fields: client_id, client_secret, refresh_token (standard
-# Google OAuth installed-app format). curl and jq are required on the host.
+# e2e_gemini_refresh_access_token [creds_file] — return a usable Gemini OAuth
+# access token. Reads access_token + expiry_date from oauth_creds.json (the
+# shape the Gemini CLI itself writes: no client_id/client_secret embedded —
+# those live in the CLI bundle). If the stored access_token still has more
+# than 5 min of life, it is printed verbatim with no network call. Otherwise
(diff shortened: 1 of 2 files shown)


hcross:
## Verdict: APPROVE

Cold review of the diff against the spec in the PR body.

### Correctness
- `e2e_gemini_refresh_access_token()` matches the documented contract: read `access_token` + `expiry_date` → if more than 5 min of life remaining, echo verbatim and return 0 with no network call; otherwise refresh via the Google token endpoint using the embedded Gemini CLI OAuth client credentials. The `expiry_date - 300000 > now_ms` guard is correctly compared in ms (matches the unit the CLI persists).
- Removed extraction of `client_id`/`client_secret` from the creds file — correct, those fields are not part of the Gemini CLI's persisted OAuth shape and were the root cause of #142.

### Portability
- `now_ms=$(( $(date +%s) * 1000 ))` is POSIX and equally correct on macOS BSD `date` and GNU `date`. Good call avoiding `%3N` (GNU-only).

### Error handling
- Every failure path terminates via `e2e_die` with an actionable hint pointing at `task e2e:auth:gemini`. The missing-creds-file, missing-refresh-token, and refresh-network-failure cases all converge on the same remediation, which is the right UX.
- Minor nit (non-blocking): when `refresh_token` is absent the message says "access token expired and refresh failed" — technically the refresh attempt did not even start. Acceptable because the remediation is identical.

### Security
- The embedded `client_id` / `client_secret` are the public installed-app OAuth credentials shipped in the open-source Gemini CLI bundle (`chunk-W2MTI7FK.js`). Per RFC 8252 §8.5 these are not confidential — embedding them in a bash helper that ships in the same repository is appropriate. No secret leakage.

### Tests
- `tests/e2e/lib/test-token-refresh.sh` is a clean standalone TAP runner: no docker, no network, isolates `e2e_die`'s `exit 1` in a subshell, builds a creds file shaped exactly like the CLI's output, and asserts both stdout and exit code. Adequate regression coverage for the issue's failure mode. The refresh path is not unit-tested, which is reasonable (would require curl mocking); the integration scenario `01-layered-context` exercises it.

### Conventions
- Gitmoji title: 🐛 ✅
- No commented-out code, no debug artifacts.
- Diff is scoped to two files matching the PR body.

### CI
All 6 checks green (build, check-components, check-skill-versions, grep-anti-patterns, lint-markdown, test-harness-curate).

Ship it.

PR #211, opened 2026-06-01
🔧 mempalace-transcript hook: portable timeout detection (#210)

### LINKED ISSUE #210: Friction cluster: mempalace-hook-darwin-timeout (1 report)
## Friction cluster: `mempalace-hook-darwin-timeout`

1 friction tagged across the `harness-friction` wing (2026-06-01 (single day)). Routed to `https://github.com/crewrig/crewrig`.

### Pattern

Reports span room `tool`. The common subcategory anchor is `mempalace-hook-darwin-timeout`.

### Frictions

1. **mempalace-transcript hook hardcodes GNU `timeout` — fails on macOS where the binary is absent by default**
   - Reported by: `claude-code`
   - Severity: `med`
   - Room: `tool`
   - Evidence: hooks/mempalace-transcript.sh:138 — `timeout 5 "$MEMPALACE_PYTHON" - 2>"$_HOOK_ERR" <<'PYEOF'`. `timeout(1)` is part of GNU coreutils and is not present on macOS by default. macOS users without Homebrew coreutils installed hit `command not found`, returning rc=127, and the hook logs `FAILED to persist tool-use (rc=127): /Users/<u>/.gemini/hooks/mempalace-transcript.sh: ligne 157: timeout: commande introuvable`. Transcripts are silently dropped on every tool-use event.
   - Evidence: hooks/mempalace-transcript.sh:122-127 — the comment above the call cites issues #90, #93, #94 as the rationale for the timeout wrapper (parallel-agent SQLite contention, stderr swallowing, hung-Python Stop hook loop). The fix that landed for #94 introduced the very portability gap reported here. The hook has been hardened four times against parallel-agent races but never against macOS portability.
   - Evidence: User live report (Gemini CLI session, 2026-06-01, Darwin 25.5.0): `mempalace-transcript: FAILED to persist tool-use (rc=127): /Users/hoanicross/.gemini/hooks/mempalace-transcript.sh: ligne 157: timeout: commande introuvable [transcript-tool-use]`.
   - Evidence: No existing open issue tracks this — verified via `gh issue list --search "timeout darwin"` and `gh issue list --search "mempalace-transcript"`. Closed issues on the same hook: #90, #91, #92, #93, #94 (all unrelated to portability).
   - Suggestion (from reporter): Detect the available timeout binary at script init (before line 138) and bind it to a variable, falling back to an empty value (no timeout, but log a warning at install/setup time) when neither is present. Sketch:

### Suggested resolution

All reporters converge on: Detect the available timeout binary at script init (before line 138) and bind it to a variable, falling back to an empty value (no timeout, but log a warning at install/setup time) when neither is present. Sketch:

### Out of scope

This issue was auto-generated by the Harness Curator (V0). It is descriptive only — a follow-up MR (human-authored, or via the auto-fix mode tracked in #42) will close it with the actual diff.

hcross:
Replaces the hardcoded `timeout 5` invocation in `hooks/mempalace-transcript.sh` with a runtime detection that picks GNU `timeout`, Homebrew's `gtimeout`, or a no-timeout fallback — fixing the rc=127 silent-drop of every transcript event on macOS without coreutils.

## How to read this PR?

Single file changed: `hooks/mempalace-transcript.sh` (+21 / −5). Three logical hunks:

1. **Lines 35–43 (new)** — detection block right after the existing `jq` dependency check. Binds `_HOOK_TIMEOUT` to `timeout` / `gtimeout` / empty string depending on PATH availability.
2. **Lines 138–146 (rewording)** — the comment block above the Python invocation, updated to cite `$_HOOK_TIMEOUT 5` and reference #210 alongside the existing #90 / #93 / #94 references.
3. **Line 154 (the actual call)** — `timeout 5 "$MEMPALACE_PYTHON"…` becomes `${_HOOK_TIMEOUT:+$_HOOK_TIMEOUT 5} "$MEMPALACE_PYTHON"…`. The `${VAR:+VAR …}` parameter-expansion idiom emits nothing when `_HOOK_TIMEOUT` is empty, leaving a bare Python call on systems lacking both binaries.

## How to test this PR?

1. **Syntax check** — `bash -n hooks/mempalace-transcript.sh` → exit 0.
2. **Linux happy path** — on any system where `timeout(1)` is on PATH, run a Gemini CLI or Claude Code session with `MEMPALACE_TRANSCRIPT_ENABLED=1`; transcripts persist as before. Sanity-check stderr for `mempalace-transcript: persisted ... to transcripts/...`.
3. **macOS happy path (with coreutils)** — `brew install coreutils` provides `gtimeout`. Run the same hook invocation; stderr should show successful persistence (no warning).
4. **macOS fallback path (without coreutils)** — uninstall coreutils (or simulate by hiding `timeout`/`gtimeout` from PATH). The hook should emit one warning at init (`neither 'timeout' nor 'gtimeout' on PATH; persistence runs without timeout protection…`) and then proceed without rc=127.
5. **Reproduction confirmation** — before this PR, the user reported `mempalace-transcript: FAILED to persist tool-use (rc=127): … timeout: commande introuvable` on every event from a macOS Gemini CLI session. After merge + redeployment of `~/.gemini/hooks/mempalace-transcript.sh`, that error should disappear (or be replaced by the one-time init warning if coreutils is not installed).

**Why this PR exists.** Issue #210 reports that `hooks/mempalace-transcript.sh` line 138 hardcodes GNU `timeout(1)`, which is not present on macOS by default. macOS users without Homebrew's `coreutils` installed hit `command not found`, the hook returns rc=127, and every transcript event is silently dropped. The fix that introduced the `timeout 5` wrapper (PR closing issue #94) addressed a hung-Python Stop-hook loop on Linux; it did not consider Darwin portability. Issue #210 is the resulting portability gap.

**Detection logic.** Three-step priority:

```sh
if command -v timeout >/dev/null 2>&1; then
  _HOOK_TIMEOUT="timeout"
elif command -v gtimeout >/dev/null 2>&1; then
  _HOOK_TIMEOUT="gtimeout"
else
  _HOOK_TIMEOUT=""
  echo "<one-time warning at init>" >&2
fi
```

The detection runs once at script init, after the opt-in guard and the `jq` check (so the cost is paid only when the hook actually runs). The chosen binary name is captured in `_HOOK_TIMEOUT` and reused at the call site via `${_HOOK_TIMEOUT:+$_HOOK_TIMEOUT 5}` — a POSIX-portable parameter expansion that emits the full `<binary> 5` token only when `_HOOK_TIMEOUT` is non-empty, and emits nothing otherwise. The Python call then runs unwrapped on systems without either binary.

**Tradeoff on the third fallback.** When neither `timeout` nor `gtimeout` is present, the script logs a one-time warning and proceeds without timeout protection. This intentionally re-opens the issue-94 hung-Python failure mode on those systems. The tradeoff is sound: rc=127 on every event (current state on macOS without coreutils) silently drops every transcript — a 100% failure rate; the hung-Python case is rare and only manifests when MemPalace itself stalls. The warning gives the user a clear migration path (`brew install coreutils`) to restore protection.

**CLI-matrix impact.** `hooks/mempalace-transcript.sh` is invoked by all three supported CLIs (Claude / Gemini / Copilot). The fix is uniform across CLIs — same binary detection, same fallback — so no CLI-specific behaviour is introduced. The trigger surface in *AGENTS.md → CLI Matrix Maintenance* lists `hooks/*-transcript-hooks.json` (the registration JSONs) but not the shared hook script itself; `docs/cli-matrix.md` therefore needs no update for this PR.

**No version bump.** The file is not under `community-config/skills/*/SKILL.md` nor `community-config/agents/*/AGENT.md` — the *Version Bump Convention* rule does not apply. The hook has no provenance frontmatter.

**No spec, no plan.** Treated as **trivial tier** + **AUTO** mode per user direction. The change is mechanical (binary detection + parameter-expansion wrap), bounded to a single file with no conceptual surface beyond portability, and would not benefit from a spec-PR + plan-comment cycle. The orchestrator handled the edit inline; a cold pr-reviewer pass confirms the change before merge.

**Build / test surface.** No CI tests cover this hook directly. The repo's `lint-markdown` and `check-components` jobs are unaffected. The change will surface in CI only via the absence of regression on the existing checks.

Closes #210

### DIFF
--- a/hooks/mempalace-transcript.sh
+++ b/hooks/mempalace-transcript.sh
@@ -28,6 +28,20 @@ fi
 # --- Dependencies ---
 command -v jq >/dev/null 2>&1 || { echo "mempalace-transcript: jq required" >&2; exit 0; }
 
+# Detect a portable `timeout` binary. GNU coreutils ships `timeout(1)` on
+# Linux out of the box; macOS does not. Homebrew's `coreutils` package
+# provides `gtimeout`. Fall back to no timeout protection when neither is
+# present — the issue-94 Stop-hook-loop risk returns, but rc=127 on every
+# event (issue #210) is a worse failure mode than the rare hung-Python case.
+if command -v timeout >/dev/null 2>&1; then
+  _HOOK_TIMEOUT="timeout"
+elif command -v gtimeout >/dev/null 2>&1; then
+  _HOOK_TIMEOUT="gtimeout"
+else
+  _HOOK_TIMEOUT=""
+  echo "mempalace-transcript: neither 'timeout' nor 'gtimeout' on PATH; persistence runs without timeout protection (install coreutils to restore it, e.g. 'brew install coreutils' on macOS)" >&2
+fi
+
 MEMPALACE_PYTHON="${MEMPALACE_PYTHON:-python3}"
 
 # --- Read input ---
@@ -121,10 +135,12 @@ if [ -n "$CONTENT" ]; then
 
   # Capture Python stderr to a dedicated file so import/runtime errors are
   # visible instead of being swallowed into $STATUS (issue #93). The
-  # `timeout 5` wrapper kills a hung Python after 5 seconds so a MemPalace
-  # lock cannot stall the calling CLI (issues #90, #94). `set +e`/`set -e`
-  # brackets the call so a non-zero Python exit does not abort the hook —
-  # STATUS_RC carries the actual outcome.
+  # `$_HOOK_TIMEOUT 5` wrapper (resolved at script init to `timeout` or
+  # `gtimeout` per portability detection — issue #210) kills a hung Python
+  # after 5 seconds so a MemPalace lock cannot stall the calling CLI
+  # (issues #90, #94). `set +e`/`set -e` brackets the call so a non-zero
+  # Python exit does not abort the hook — STATUS_RC carries the actual
+  # outcome.
   _HOOK_ERR="${TMPDIR:-/tmp}/mempalace-hook-$$.err"
   trap 'rm -f "$_HOOK_ERR"' EXIT
 
@@ -135,7 +151,7 @@ if [ -n "$CONTENT" ]; then
     TRANSCRIPT_CONTENT="$TRANSCRIPT_CONTENT" \
     TRANSCRIPT_ROOM="$TRANSCRIPT_ROOM" \
     TRANSCRIPT_AGENT="$TRANSCRIPT_AGENT" \
-    timeout 5 "$MEMPALACE_PYTHON" - 2>"$_HOOK_ERR" <<'PYEOF'
+    ${_HOOK_TIMEOUT:+$_HOOK_TIMEOUT 5} "$MEMPALACE_PYTHON" - 2>"$_HOOK_ERR" <<'PYEOF'
 import os, sys
 try:
     from mempalace.mcp_server import tool_add_drawer


hcross:
## Verdict: APPROVE

Trivial-tier AUTO-mode review (orchestrator-posted after the cold reviewer pre-verified the 5 criteria but exited before CI completed).

Verified:

1. **Detection block correct** (lines 35–43) — `timeout` → `gtimeout` → empty + stderr warning. Placed after the opt-in guard so non-enabled sessions pay nothing.
2. **Call-site rewrite** (line 154) — POSIX `${_HOOK_TIMEOUT:+$_HOOK_TIMEOUT 5}` idiom emits nothing when the variable is empty, leaving a bare Python invocation on systems without either binary.
3. **Bash syntax** — `bash -n hooks/mempalace-transcript.sh` exits 0.
4. **Single-file scope** — diff touches only `hooks/mempalace-transcript.sh` (+21 / −5).
5. **CI** — all 6 required checks green on head `10ccf8f` (`build`, `check-components`, `check-skill-versions`, `grep-anti-patterns`, `lint-markdown`, `test-harness-curate`).

No findings of any class. Ready to merge.

PR #176, opened 2026-05-31
📐 ADR-0010 — SPECS → PLAN → DEV → REVIEW lifecycle (#166)

hcross:
Introduces ADR-0010 defining a four-stage SPECS → PLAN → DEV → REVIEW lifecycle for every non-trivial CrewRig ticket, with a finding-classification taxonomy that routes review feedback back to the correct upstream stage. Updates `AGENTS.md` to layer the operative rules (lifecycle overview, interaction modes, retroactive review loop) onto the ADR contract.

## How to read this PR?

Read in this order:

1. **`docs/adr/0010-spec-plan-review-lifecycle.md`** — this is the **normative** document. Start at *Context*, walk through *Decision* → *Stage definitions and transition rules* → *Finding classification taxonomy* → *Routing matrix* → *Interaction modes* → *Complexity tiers* → *REVIEW loop termination* → *Consequences* → *Appendix (three worked examples)*. The appendix's three examples (trivial / standard / large) are deliberately written to act as fixtures for the downstream tickets that will implement the contract.
2. **`AGENTS.md` diff** — this is the **operative restatement**. Four touchpoints:
   - new top-level *Lifecycle* section right after *What is CrewRig?* (overview + ASCII diagram + pointer to ADR-0010);
   - amended intro paragraph at the head of *Agent Team Protocol* framing Templates 1/2/3 as the **DEV stage** of the lifecycle and forward-referencing ADR-0010 for the surrounding stages;
   - new *Interaction modes* section before *Pull Request Format* (the operative table; engine deferred to #173);
   - new *Retroactive review loop* section (the operative routing table; engine deferred to #172).

The ADR is the source of truth; `AGENTS.md` is the agent-facing restatement of the parts an orchestrator must apply at ticket pickup. Where the two could drift, the ADR wins.

**Three parked questions** flagged by the architect — they are intentionally NOT resolved in this PR and are forwarded to the downstream tickets where they belong:

1. **Frontmatter field names** for `mode` and `max_iterations` — schema lands with the spec file format in **#167**.
2. **Trivial tier vs Logbook Rule A** — ADR-0010 → *Complexity tiers* says trivial uses "the issue itself" as logbook; whether this is full Rule A compliance or a narrow tier-specific exemption is forwarded to **#175** (migration ticket) or a follow-up edit on this branch if the user prefers in-scope resolution.
[...]

PR #217, opened 2026-06-02
🐛 Relax identity-file prereq check to presence-only

### LINKED ISSUE #146: 🐛 setup-claude-interactive rejects unmodified SOUL.md even after /init-soul ran — clarify whether customization is mandatory
## Symptom

During a first install following the README, the user ran `/init-soul` before `task setup-claude-interactive`, accepted the proposed defaults without rewording any section, and then hit:

```
Cannot proceed — required identity files are missing or not customized:
  - config/SOUL.md is identical to its template — run: claude /init-soul to customize it
```

From the user's perspective the prerequisite was satisfied — the skill was invoked, the conversation completed — but the setup script refused to proceed.

## Root cause

`scripts/setup-claude-interactive.sh` (`check_finalized`, lines 70-79) enforces the prerequisite with a strict byte-for-byte `diff -q` between `config/SOUL.md` and `config/SOUL.md.template`. If the two files are identical, the script fails.

The `/init-soul` skill (`community-config/skills/init-soul/SKILL.md`) is a guided conversation. When the user accepts every proposed value as-is, the resulting `config/SOUL.md` can end up identical to the template, and the setup script then rejects it.

The two contracts disagree:

- The skill behaves as if accepting defaults is a valid outcome.
- The script behaves as if any unchanged byte means the user never ran the skill.

## Question to clarify

**Is customizing `SOUL.md` mandatory or optional for the first install?**

- **If mandatory** (the script is correct): `/init-soul` should refuse to write a file identical to the template — it should force at least one meaningful divergence (e.g., inject the user's name/role from PROFILE.md) or warn the user that accepting defaults will block setup. The README should also state this expectation explicitly.
- **If optional** (the skill is correct): the script's check is too strict. The prerequisite should be \"file exists and was generated by the skill\" (e.g., presence of a marker, or simply file existence), not \"file differs byte-for-byte from the template\".

Either resolution is acceptable, but the current state — where running the documented command in the documented order can still fail — is a first-install footgun.

## Suggested directions

1. Decide the policy (mandatory vs. optional customization).
2. Align the two contracts:
   - If mandatory → `/init-soul` always produces a divergent file (inject PROFILE-derived fields, or block on accepting all defaults).
   - If optional → relax `check_finalized` to a presence check, or use a marker comment (e.g. `<!-- init-soul: completed -->`) inserted by the skill.
3. Mirror the same resolution for `PROFILE.md` / `/init-personal-profile` — same pattern, same risk.

## Evidence

- `scripts/setup-claude-interactive.sh:70-91` — strict diff check
- `community-config/skills/init-soul/SKILL.md` — guided conversation, no enforced divergence
- Reported on first install, 2026-05-28

## Environment

- macOS (Darwin 25.5.0), zsh
- `main` branch at `444dc71`

gfourny-sfeir:
Fix the first-install footgun where `/init-soul` (or `/init-personal-profile`) followed by `task setup-<cli>-interactive` could legitimately fail when the user accepted all defaults. Switch the prerequisite contract from byte-equality with the template to file presence, symmetrically across the three CLI setup scripts.

## How to read this PR?

Read `scripts/setup-claude-interactive.sh` first — the diff is the canonical pattern. The two siblings (`setup-gemini-interactive.sh`, `setup-copilot-interactive.sh`) carry the identical change for parity. The `check_finalized` helper drops its `template` and `diff -q` arguments and keeps only the presence test; user-facing message updated to reflect the new contract ("missing" — no longer "missing or not customized").

The CLI matrix (`docs/cli-matrix.md` row 10) describes the setup-script capability at a granularity that does not track the prereq check; no matrix edit was required (the row already reflects symmetric `✅` across the three CLIs and stays so).

## How to test this PR?

```sh
# Reproduce the original symptom on `main` (optional, for comparison):
cp config/SOUL.md.template config/SOUL.md
cp config/PROFILE.md.template config/PROFILE.md
bash scripts/setup-claude-interactive.sh   # FAILED before this PR

# On this branch:
cp config/SOUL.md.template config/SOUL.md
cp config/PROFILE.md.template config/PROFILE.md
bash scripts/setup-claude-interactive.sh   # PASSES the prereq block

# Negative path (still enforced):
rm config/SOUL.md
bash scripts/setup-claude-interactive.sh   # fails with: 'config/SOUL.md is missing — run: claude /init-soul'
```

Same procedure works for the gemini and copilot setup scripts.

**Files modified (3):**

- `scripts/setup-claude-interactive.sh` (lines 62-74 in the new version)
- `scripts/setup-gemini-interactive.sh` (lines 56-67 in the new version)
- `scripts/setup-copilot-interactive.sh` (lines 56-69 in the new version)

**Change in `check_finalized`:**

Before:
```sh
check_finalized() {
  local file="$1" template="$2" label="$3" skill="$4"
  if [ ! -f "$file" ]; then
    MISSING_PREREQS+=("$label is missing — run: <cli> $skill")
  elif [ -f "$template" ] && diff -q "$file" "$template" >/dev/null 2>&1; then
    MISSING_PREREQS+=("$label is identical to its template — run: <cli> $skill to customize it")
  fi
}
```

After:
```sh
check_finalized() {
  local file="$1" label="$2" skill="$3"
  if [ ! -f "$file" ]; then
    MISSING_PREREQS+=("$label is missing — run: <cli> $skill")
  fi
}
```

Call sites updated to drop the now-unused `template` argument. The comment block above each helper documents that customization is optional and the contract is presence-only. The error message banner is updated from "missing or not customized" to "missing" to match the relaxed contract.

**Policy rationale.** Per the issue's mandatory-vs-optional question, the user chose **optional**. The `/init-soul` and `/init-personal-profile` skills are guided conversations whose contract is "produce a usable identity file"; accepting all defaults is a valid outcome of that conversation. The byte-diff check assumed the opposite. Aligning the script with the skill is less invasive than forcing the skill to inject divergence (no skill changes needed; no README change required).

**Out of scope / explicitly not changed.**

- The `/init-soul` and `/init-personal-profile` skills under `community-config/skills/` — unchanged. Optional-customization policy means they are already correct.
- Other `diff -q` usages in the same scripts (`setup-claude-interactive.sh:316`, `setup-gemini-interactive.sh:231`) — they implement an unrelated keep-or-refresh check on `~/.claude/rules/PROFILE.md` vs the repo copy, not a template prereq.
- `docs/cli-matrix.md` — consulted; no edit required (row 10 is granular at the script-capability level and already symmetric).

**Legacy ticket.** Issue #146 was classified `keep-legacy` by spec 0008 (pinned comment by @hcross). This PR therefore follows the pre-#176 contract: direct implementation PR closes the issue, no `/specs/` file, no PLAN comment, no class-tagged review findings.

Closes #146

### DIFF
--- a/scripts/setup-claude-interactive.sh
+++ b/scripts/setup-claude-interactive.sh
@@ -60,24 +60,25 @@ command -v claude >/dev/null 2>&1 || {
 }
 
 # --- Prerequisites: identity files ---
-# SOUL.md and PROFILE.md must be generated BEFORE running this setup.
+# SOUL.md and PROFILE.md must exist BEFORE running this setup.
 # They are produced by the /init-soul and /init-personal-profile skills.
+# Customization is optional: accepting all defaults in those skills is a
+# valid outcome, so a presence check is the contract — not a byte-diff
+# against the template.
 MISSING_PREREQS=()
 
 check_finalized() {
-  local file="$1" template="$2" label="$3" skill="$4"
+  local file="$1" label="$2" skill="$3"
   if [ ! -f "$file" ]; then
     MISSING_PREREQS+=("$label is missing — run: claude $skill")
-  elif [ -f "$template" ] && diff -q "$file" "$template" >/dev/null 2>&1; then
-    MISSING_PREREQS+=("$label is identical to its template — run: claude $skill to customize it")
   fi
 }
 
-check_finalized "$REPO_DIR/config/SOUL.md"    "$REPO_DIR/config/SOUL.md.template"    "config/SOUL.md"    "/init-soul"
-check_finalized "$REPO_DIR/config/PROFILE.md" "$REPO_DIR/config/PROFILE.md.template" "config/PROFILE.md" "/init-personal-profile"
+check_finalized "$REPO_DIR/config/SOUL.md"    "config/SOUL.md"    "/init-soul"
+check_finalized "$REPO_DIR/config/PROFILE.md" "config/PROFILE.md" "/init-personal-profile"
 
 if [ ${#MISSING_PREREQS[@]} -gt 0 ]; then
-  echo "Cannot proceed — required identity files are missing or not customized:"
+  echo "Cannot proceed — required identity files are missing:"
   for item in "${MISSING_PREREQS[@]}"; do
     echo "  - $item"
   done
--- a/scripts/setup-copilot-interactive.sh
+++ b/scripts/setup-copilot-interactive.sh
@@ -54,24 +54,25 @@ if ! gh copilot --help >/dev/null 2>&1 && ! command -v copilot >/dev/null 2>&1; 
 fi
 
 # --- Prerequisites: identity files ---
-# SOUL.md and PROFILE.md must be generated BEFORE running this setup.
+# SOUL.md and PROFILE.md must exist BEFORE running this setup.
 # They are produced by the /init-soul and /init-personal-profile skills.
+# Customization is optional: accepting all defaults in those skills is a
+# valid outcome, so a presence check is the contract — not a byte-diff
+# against the template.
 MISSING_PREREQS=()
 
 check_finalized() {
-  local file="$1" template="$2" label="$3" skill="$4"
+  local file="$1" label="$2" skill="$3"
   if [ ! -f "$file" ]; then
     MISSING_PREREQS+=("$label is missing — run: $skill")
-  elif [ -f "$template" ] && diff -q "$file" "$template" >/dev/null 2>&1; then
-    MISSING_PREREQS+=("$label is identical to its template — run: $skill to customize it")
   fi
 }
 
-check_finalized "$REPO_DIR/config/SOUL.md"    "$REPO_DIR/config/SOUL.md.template"    "config/SOUL.md"    "/init-soul"
-check_finalized "$REPO_DIR/config/PROFILE.md" "$REPO_DIR/config/PROFILE.md.template" "config/PROFILE.md" "/init-personal-profile"
+check_finalized "$REPO_DIR/config/SOUL.md"    "config/SOUL.md"    "/init-soul"
+check_finalized "$REPO_DIR/config/PROFILE.md" "config/PROFILE.md" "/init-personal-profile"
 
 if [ ${#MISSING_PREREQS[@]} -gt 0 ]; then
-  echo "Cannot proceed — required identity files are missing or not customized:"
+  echo "Cannot proceed — required identity files are missing:"
   for item in "${MISSING_PREREQS[@]}"; do
     echo "  - $item"
   done
--- a/scripts/setup-gemini-interactive.sh
+++ b/scripts/setup-gemini-interactive.sh
@@ -54,23 +54,24 @@ command -v jq >/dev/null 2>&1 || {
 }
 
 # --- Prerequisites: identity files ---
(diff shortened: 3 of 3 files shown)


gfourny-sfeir:
## Verdict: APPROVE

Cold review of the diff against `main` — every claim verified from the diff and the repo.

### Findings

**1. Bash correctness — clean.** The `check_finalized` signature is correctly reduced from 4 to 3 positional parameters (`file`, `label`, `skill`) in all three scripts. The body now contains a single presence test; no orphaned `$template` reference remains. Both call sites in each script are updated symmetrically (column-aligned with the new 3-arg form). `MISSING_PREREQS` plumbing and the failure banner downstream are unchanged and still wired correctly.

**2. Parity — symmetric across the three CLIs.** Identical structural change in `setup-claude-interactive.sh`, `setup-gemini-interactive.sh`, and `setup-copilot-interactive.sh`. Per-CLI wording differences in the error string (`claude $skill` / `gemini $skill` / `$skill`) preserve the pre-existing convention and are not a parity issue.

**3. Unrelated `diff -q` usages correctly left alone.** Verified `setup-claude-interactive.sh:316` and `setup-gemini-interactive.sh:231` — both implement the keep-local-vs-overwrite flow on the **deployed** `~/.<cli>/.../PROFILE.md` against the repo's `config/PROFILE.md` (workspace-sync semantics), not a template prereq check. Different contract, different fix. Leaving them untouched is correct; touching them would have been a scope error.

**4. AGENTS.md compliance — pass.**
- Gitmoji title `🐛 Relax identity-file prereq check to presence-only` — conforms.
- All project content English-only — conforms.
- PR body carries the four mandated sections (purpose paragraph, *How to read*, *How to test*, *Detailed description (for agents)*) — conforms.
- `Closes #146` directive present at the end of the body — conforms.
- Legacy-ticket contract (issue #146 classified `keep-legacy` per spec 0008) respected: no `/specs/` artefact, no PLAN comment, no class-tagged finding plumbing in the PR. Direct implementation PR is the right shape.

**5. CLI matrix obligation — judgment confirmed.** Row 10 (`Interactive setup script`) describes the capability as "the script exists and offers MemPalace auto-install via pipx" — a binary capability flag, not a behavioural inventory of every prereq check inside the script. The script still exists, still offers MemPalace, still runs end-to-end on all three CLIs. Row 7b (layered context) similarly does not enumerate prereq-check semantics. No matrix cell becomes stale; no edit was required. The PR body's claim on this point is accurate.

**6. CI — green on `a7c8fadbb2434d44365cd9242fa2a47d9c8b1f2e`.** All 6 checks SUCCESS: `build`, `check-components`, `check-skill-versions`, `lint-markdown`, `grep-anti-patterns`, `test-harness-curate`.

**7. Other risk surface — none material.**
- No regression risk on the negative path: deleting `config/SOUL.md` still triggers the `is missing` branch and aborts setup with the correct guidance message.
- No regression risk on the positive path: an `/init-soul` user who customised the file still passes (presence-only is strictly weaker than the previous compound check).
- The relaxed contract is documented inline in the comment block above the helper in all three scripts, which is the right place for a future maintainer to discover the policy.

### Nit (non-blocking)

User-facing banner now reads `Cannot proceed — required identity files are missing:` — mildly redundant ("required" + "missing"). Optional polish to `Cannot proceed — identity files are missing:` if a follow-up touches this code; not worth a turnaround on its own.

### Decision

Approve. Ship it.

PR #303, opened 2026-06-13
🐛 Raise Chroma daemon file-descriptor limits and reap orphaned wrappers

### LINKED ISSUE #302: 🔧 Implement spec 0029 — Chroma daemon file-descriptor limits + socket-leak containment
Implementation logbook for the durable fix qualified by spec [`0029-chroma-daemon-fd-limits`](https://github.com/crewrig/crewrig/blob/main/specs/0029-chroma-daemon-fd-limits.md).

This is the **implementation** issue, independent from the spec issue #300 (now closed), mirroring the spec-PR / impl-PR two-issue convention used for spec 0028 (#296 / #298).

## Scope (from spec 0029)

- R1–R3: raise the open-file-descriptor limit to ≥ 65536 on **both** supervisor configs shipped in the repo — `config/launchd/com.mempalace.chroma-server.plist` (`SoftResourceLimits`/`HardResourceLimits` → `NumberOfFiles`) and `config/systemd/mempalace-chroma-server.service` (`LimitNOFILE`).
- R4–R5: investigate and contain the underlying socket leak (MCP wrapper connection lifecycle + orphaned-wrapper handling) — the daemon must not accumulate unbounded socket descriptors across sessions.
- R6: document the live-reload procedure for an already-running daemon in `docs/runbooks/chroma-http-server.md`.

## Operational context

The original incident saturated a daemon at the macOS default soft limit of 256 after ~4 days of accumulated sockets. A healthy daemon was respawned operationally but still runs with the 256 limit — the fix must ship before re-saturation (~4-day window).

Lifecycle: SPECS done (spec 0029 merged). This issue tracks PLAN → DEV → REVIEW.

hcross:
Realizes spec [0029](https://github.com/crewrig/crewrig/blob/main/specs/0029-chroma-daemon-fd-limits.md): the shared ChromaDB daemon no longer inherits the macOS default 256 open-file limit (which saturated after ~4 days and surfaced as MemPalace "No palace found"), and orphaned MCP wrapper processes self-reap so they stop leaking daemon sockets.

## How to read this PR?

1. `config/launchd/com.mempalace.chroma-server.plist` and `config/systemd/mempalace-chroma-server.service` — the core fix: declare a 65536 open-file floor on both supervisors (R1–R3).
2. `scripts/lib/mempalace-http-wrapper.py` — the socket-leak containment (R4–R5). Read the long comment block: `mempalace.mcp_server.main()` already exits on stdin EOF (the common orphan case), so this adds only a `getppid()`-poll watchdog for the residual case where the parent dies but the stdin write-end is held open elsewhere. It uses `os._exit` (a daemon thread cannot terminate the process via `sys.exit`) and never reads stdin (so it cannot steal JSON-RPC bytes `main()` owns).
3. `docs/runbooks/chroma-http-server.md` — live-reload procedure for an already-running daemon (R6).
4. `scripts/tests/test-chroma-fd-limits.sh` — static regression test locking the 65536 floor in both units.

## How to test this PR?

```sh
bash scripts/tests/test-chroma-fd-limits.sh   # expect: 4 passed, 0 failed
plutil -lint config/launchd/com.mempalace.chroma-server.plist   # expect: OK
python3 -c "import ast; ast.parse(open('scripts/lib/mempalace-http-wrapper.py').read())"
```

Deployment (applies the raised limit to the live daemon) follows the new runbook subsection: macOS `launchctl bootout`/`bootstrap`, Linux `systemctl --user daemon-reload && restart`.

- **launchd plist**: added `SoftResourceLimits` and `HardResourceLimits` dicts, each `NumberOfFiles` = 65536, after `KeepAlive`. Placeholders untouched; `plutil -lint` passes.
- **systemd unit**: added `LimitNOFILE=65536` in `[Service]` (systemd's single scalar sets both soft and hard).
- **`scripts/lib/common.sh`**: NO change — verified the macOS install path substitutes only the three `__…__` placeholders via `sed` and the Linux path is a straight `cp`, so the new keys pass through verbatim.
- **wrapper watchdog**: a `daemon=True` thread polls `os.getppid()` every 5s and calls `os._exit(0)` when reparented to PID 1 (orphaned). Sockets close on process exit, so termination alone releases the daemon-side fd — satisfying R4's "successive sessions" clause without an explicit close.
- **Blast radius**: CLI Matrix NOT triggered (`scripts/lib/**`, `config/launchd/**`, `config/systemd/**`, runbooks are absent from the trigger surface); no version bump (no `artifacts/` source); no build outputs; `scripts/build-components.sh` not run.

Closes #302

### DIFF
--- a/config/launchd/com.mempalace.chroma-server.plist
+++ b/config/launchd/com.mempalace.chroma-server.plist
@@ -27,6 +27,18 @@
     <key>KeepAlive</key>
     <true/>
 
+    <key>SoftResourceLimits</key>
+    <dict>
+        <key>NumberOfFiles</key>
+        <integer>65536</integer>
+    </dict>
+
+    <key>HardResourceLimits</key>
+    <dict>
+        <key>NumberOfFiles</key>
+        <integer>65536</integer>
+    </dict>
+
     <key>StandardOutPath</key>
     <string>__MEMPALACE_HOME__/chroma-server.log</string>
 
--- a/config/systemd/mempalace-chroma-server.service
+++ b/config/systemd/mempalace-chroma-server.service
@@ -9,6 +9,7 @@ WorkingDirectory=%h/.mempalace
 ExecStart=%h/.local/pipx/venvs/mempalace/bin/python3.13 %h/.local/pipx/venvs/mempalace/bin/chroma run --path %h/.mempalace/palace --host 127.0.0.1 --port 8001
 Restart=always
 RestartSec=3
+LimitNOFILE=65536
 StandardOutput=append:%h/.mempalace/chroma-server.log
 StandardError=append:%h/.mempalace/chroma-server.log
 
--- a/docs/runbooks/chroma-http-server.md
+++ b/docs/runbooks/chroma-http-server.md
@@ -45,6 +45,44 @@ restarts the daemon within seconds of a crash; the manual `start` and
   for the supervisor's own diagnostics.
 - **systemd specifics** — `journalctl --user -u mempalace-chroma-server`.
 
+### Applying the raised file-descriptor limit to a running daemon
+
+The shipped supervisor units declare an open-file floor of `65536`
+(launchd `SoftResourceLimits`/`HardResourceLimits` → `NumberOfFiles`;
+systemd `LimitNOFILE`). A fresh install inherits it automatically, but a
+daemon already running under the old unit keeps its previous limit until
+the supervisor re-execs it. Remediate an existing install without a host
+restart:
+
+- **macOS** — reload the launchd job so it re-reads the updated plist:
+
+  ```sh
+  launchctl bootout gui/$(id -u)/com.mempalace.chroma-server
+  launchctl bootstrap gui/$(id -u) ~/Library/LaunchAgents/com.mempalace.chroma-server.plist
+  ```
+
+- **Linux** — reload the unit definition and restart the service:
+
+  ```sh
+  systemctl --user daemon-reload
+  systemctl --user restart mempalace-chroma-server
+  ```
+
+Verify the new limit took effect by inspecting the running daemon's
+open-file limit:
+
+- **macOS** — `launchctl print gui/$(id -u)/com.mempalace.chroma-server`
+  reports the job's resource limits; look for the `number of files`
+  entry under the `inherited limits` / `hard/soft limits` section.
+
+- **Linux** — read the kernel's per-process limit:
+
+  ```sh
+  cat /proc/"$(pgrep -f 'chroma run')"/limits | grep 'Max open files'
+  ```
+
+Both should report `65536` (or higher) once the reload completes.
+
 ## Migrating from the legacy `PersistentClient` setup
 
 If you upgraded a working CrewRig install across the #98 boundary:
--- a/scripts/lib/mempalace-http-wrapper.py
+++ b/scripts/lib/mempalace-http-wrapper.py
@@ -26,6 +26,32 @@ design — it would re-introduce the corruption bug.
 """
 import os
 import sys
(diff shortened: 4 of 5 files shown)


hcross:
## Cold PR review — independent audit

**Verdict: APPROVE** — zero blocking findings. The change correctly realizes spec 0029 (R1–R6); CI is green; the watchdog is sound.

> Posted as a comment (not a formal review) because the reviewer shares the PR author's GitHub identity — the shared-identity workaround per `AGENTS.md → Plan review protocol`.

### CI status
All required checks pass: `build`, `check-components`, `check-agents-size`, `check-skill-versions`, `grep-anti-patterns`, `lint-markdown`, `lint-specs`, `test-harness-curate`. (`check-components` was pending at review start and resolved to pass.)

### Correctness against the spec
- **R1/R2/R3 (launchd)** — `SoftResourceLimits` + `HardResourceLimits` dicts, each `NumberOfFiles` = 65536. `plutil -lint` → OK. Soft and hard both at the floor; hard ≥ soft. ✓
- **R1/R2/R3 (systemd)** — `LimitNOFILE=65536` in `[Service]`; systemd's single scalar sets both soft and hard. ✓
- **R4/R5 (orphan reap)** — the `getppid()` watchdog releases daemon sockets on process exit. ✓
- **R6 (live-reload)** — runbook documents `launchctl bootout`/`bootstrap` (macOS) and `daemon-reload`/`restart` (Linux) plus verification commands. ✓

### Watchdog (the only runtime change — scrutinized hardest)
- `os.getppid() == 1` is a **sound** orphan signal on both OSes: the wrapper is always spawned as an agent-CLI MCP stdio child (`mcpServers.mempalace` in the Claude/Gemini/Copilot configs — verified), never supervised directly by launchd/systemd. Orphans reparent to PID 1 (launchd on macOS, init/systemd on Linux). A live CLI host is never PID 1, so there is **no path to reaping a legitimately-running wrapper**.
- **No JSON-RPC interference**: the thread only sleeps and polls `getppid()` — it never touches stdin/stdout, so it cannot steal bytes `main()` owns.
- **`os._exit(0)` is correct** (vs `sys.exit`): a daemon thread's `sys.exit` raises `SystemExit` only in itself and would not terminate the process — explicitly addressed (cold-review finding #2).
- **No startup race**: the thread starts before the chromadb probe, but its first action is `time.sleep(5)`; if the probe's `sys.exit(1)` fires first, the daemon thread dies with the process. Worst case is a bounded ≤5s latency before reaping an orphan — acceptable.

### Convention compliance (verified, not trusted)
- Gitmoji title (`🐛 …`) ✓ · English-only ✓ · `Closes #302` present ✓
- No version bump — no `artifacts/**` source touched ✓
- CLI-matrix NOT triggered — `scripts/lib/**`, `config/launchd/**`, `config/systemd/**`, runbooks are absent from the trigger surface (verified against `docs/cli-matrix-maintenance.md`) ✓
- Branch `fix/0029-chroma-daemon-fd-limits` matches the spec ordering convention ✓

### Test quality
`scripts/tests/test-chroma-fd-limits.sh` asserts the 65536 floor on both units **and** that hard ≥ soft. Ran it: **4 passed, 0 failed**. shellcheck → clean.

### Linter findings
- `plutil -lint` plist → OK
- `python3 -c "ast.parse(...)"` wrapper → OK
- shellcheck on the test → clean
- markdownlint flagged MD060 (table alignment) on runbook lines 30–35 — a **pre-existing table outside this diff** (added content is lines 48–86); repo CI `lint-markdown` passes. Not a finding against this PR.

No findings of any class (`tech` / `arch` / `spec`). Clean pass — termination criterion met (APPROVE + zero findings + CI green).

PR #362, opened 2026-06-18
🐛 Install Claude plugins into a shared out-of-tree marketplace

### LINKED ISSUE #360: 🐛 Claude plugin install uses in-tree marketplace: second install clobbers the first, breaks on branch switch
## Symptom

Installing one Claude Code plugin then another via `scripts/install-claude-plugin.sh` makes the **first plugin disappear**, and any installed plugin breaks after a git branch switch.

## Root cause

`install-claude-plugin.sh` writes the build output and the local marketplace into the **working tree**, at `extensions/<tier>/<name>/dist-claude-plugin/` (line 48), a `.gitignore`-d path (`.gitignore:17`). Two distinct defects follow:

1. **Multi-extension collision.** The marketplace name is constant — `<repo-basename>-local` = `crewrig-local` (line 53) — but the *registered path* is per-extension. `claude plugin marketplace add` indexes marketplaces by name, so the second install re-points `crewrig-local` at the second extension's parent, dropping the first extension's source. The plugin-entry **merge logic already written at lines 67-71 never fires**, because it only merges when a `marketplace.json` already exists *under the same parent* — which never happens across different extensions (comment lines 62-66 admit this: "not the current default, but supported").

2. **Branch-switch fragility.** The registered marketplace source points inside the working tree at a gitignored, build-transient path coupled to the checked-out branch. Switching branches leaves the registered path stale or empty.

## Why this is the outlier

The Gemini side already does the right thing: `install-extension.sh` deploys to `$HOME/.gemini/extensions/<name>` (line 45) — $HOME-anchored, out of the working tree. Claude is the only CLI building its installed artifact in-tree.

## Proposed direction

Relocate the local marketplace to a **single shared $HOME directory** — `${CLAUDE_CONFIG_DIR:-$HOME/.claude}/local-marketplace/` — with one `crewrig-local` marketplace and one shared `marketplace.json`. This activates the dormant merge logic (plugins coexist, upsert by name) and decouples the installed artifact from working-tree state, mirroring the Gemini model.

Follows the SPECS → PLAN → DEV → REVIEW lifecycle. Spec id: 0045.

hcross:
Fix the Claude plugin installer so multiple extensions coexist and survive branch switches, by relocating the local marketplace out of the gitignored working tree into a single shared `${CLAUDE_CONFIG_DIR:-$HOME/.claude}/local-marketplace/`. This realizes spec 0045 (`shared-plugin-marketplace`).

Closes #360

## How to read this PR?

Read in this order:

1. **`scripts/install-claude-plugin.sh`** — the core fix. The build parent moves from the per-extension in-tree `$EXT_DIR/dist-claude-plugin` to the single shared `MARKET_HOME="${CLAUDE_CONFIG_DIR:-$HOME/.claude}/local-marketplace"`, and the header/comments are rewritten to describe the shared home. The upsert-by-name merge logic (filter out any prior entry of the same name, then append) is unchanged — it now fires because every extension shares one parent.
2. **`scripts/tests/test-install-claude-plugin-marketplace.sh`** (new) — the regression. Read the `Cases → spec 0045 scenarios` header block to see how the four cases map to R1/R2/R4/R5.
3. **`DEVELOPMENT.md`** and **`docs/cli-matrix.md`** (row 14 only) — doc corrections to the new paths.
4. **`.github/workflows/build.yml`** — registers the new test in CI.

Non-obvious decisions:

- **`build-claude-plugin.sh` is intentionally unchanged.** It takes an explicit output dir from the caller; the relocation is the caller's concern, not the builder's. The bare build path still emits in-tree.
- **`docs/cli-matrix.md` rows 5b / 13 / 20 are deliberately untouched** — they describe the build-time render target / gitignore carve-out, not the install location. Only row 14 (the install-script row) describes where installs land.
- **The `--plugin-dir` build/dev references in `DEVELOPMENT.md` are left untouched** — they concern the dev workflow, not the install flow.

## How to test this PR?

Prerequisites: `bash`, `jq`, and the real `extensions/core/hello-world` extension (the test's subject).

```sh
bash scripts/tests/test-install-claude-plugin-marketplace.sh
```

Expected: `11 passed, 0 failed`.

Lint the changed shell:

```sh
shellcheck scripts/install-claude-plugin.sh scripts/tests/test-install-claude-plugin-marketplace.sh
```

Expected: no findings.

### Root cause fixed

The constant `crewrig-local` marketplace name was re-pointed at a per-extension in-tree `dist-claude-plugin/` directory on every `claude plugin marketplace add`, so the second install clobbered the first. The script's existing upsert-by-name merge logic never fired, because each extension had its own parent directory and therefore its own `marketplace.json`. Sharing one home directory makes that merge fire (coexistence + idempotent upsert) and decouples the registration from working-tree state — which is why installs now survive branch switches (the prior home was gitignored and vanished on switch).

### Changes

- **`scripts/install-claude-plugin.sh`** — core fix plus header/comment rewrite. `BUILD_PARENT` changes from `$EXT_DIR/dist-claude-plugin` to `MARKET_HOME="${CLAUDE_CONFIG_DIR:-$HOME/.claude}/local-marketplace"`, with a `mkdir -p "$BUILD_PARENT"` added. `BUILD_DIR="$BUILD_PARENT/$EXT_NAME"` is preserved, so each extension still gets its own subdirectory under the shared home. The upsert merge (`jq --arg n "$EXT_NAME" '[.plugins[] | select(.name != $n)]'`) is unchanged.
- **`scripts/tests/test-install-claude-plugin-marketplace.sh`** (new, +207 lines) — PATH-stub regression. 11 `check` assertions across 4 cases mapping to spec 0045 scenarios: Case 1 out-of-tree location (R4 + branch-switch durability), Case 2 coexistence / upsert (R1 + R2), Case 3 idempotent re-install (R5), Case 4 failed install leaves the shared registry intact. The harness uses a temp `CLAUDE_CONFIG_DIR` (never touching the real `$HOME/.claude`), a no-op `claude` stub prepended to `PATH`, real `jq`, and the real `extensions/core/hello-world` as subject. A negative control (running the assertions against a regressed in-tree script) goes red on exactly the load-bearing assertions, proving they are not vacuous.
- **`DEVELOPMENT.md`** — the four-step install-flow description is corrected to the new shared-home paths. The `--plugin-dir` build/dev references are left untouched.
- **`docs/cli-matrix.md`** — row 14 (install-script row) only. Rows 5b / 13 / 20 are deliberately untouched because they describe the build-time render target / gitignore carve-out, not the install location.
- **`.github/workflows/build.yml`** — adds a CI step `Test install-claude-plugin marketplace regression (spec 0045)` running the new test.
- **`scripts/build-claude-plugin.sh`** — intentionally unchanged; it takes an explicit output dir from the caller.

### Known limitation

The shared home is keyed only on `CLAUDE_CONFIG_DIR`, so two different repos under the same config root share one `marketplace.json`, whose top-level `name` is last-writer-wins. This is acceptable under spec 0045's single-repo / single-maintainer assumption; multi-repo namespacing is out of scope, adjacent to spec 0045's out-of-scope orphan-GC carve-out.

### Lifecycle context

Realizes spec 0045 (`shared-plugin-marketplace`), merged via spec-PR #361. SPECS and PLAN both passed cold review; PLAN was INTERMEDIATE-mode-validated by the user.


hcross:
## Cold review — PR #362 (spec 0045, `shared-plugin-marketplace`, issue #360)

Independent `pr-reviewer` pass, no authoring context. Every claim grounded against the head ref `fix/0045-shared-plugin-marketplace` (`1f16299`), the merged spec on `main`, and the APPROVED PLAN v2 on issue #360. Test run and shellcheck executed locally.

### CI status

All 10 required checks **pass** (verified twice; `check-components` was pending at first fetch, now green at 53s):
`build`, `check-agents-size`, `check-components`, `check-extension-version-bump`, `check-feedback-routing`, `check-skill-versions`, `grep-anti-patterns`, `lint-markdown`, `lint-specs`, `test-harness-curate`. No red, no pending.

### Correctness of the core fix (`scripts/install-claude-plugin.sh`)

Confirmed correct on all three mechanisms the PR claims:

- **(a) Stops the multi-extension clobber + (b) activates upsert-by-name + (c) takes the registry off the working tree.** `BUILD_PARENT` moves from the per-extension `$EXT_DIR/dist-claude-plugin` to the single shared `MARKET_HOME="${CLAUDE_CONFIG_DIR:-$HOME/.claude}/local-marketplace"` (`:52-54`). Because all extensions now share one `BUILD_PARENT`, `MARKETPLACE_DIR="$BUILD_PARENT/.claude-plugin"` (`:65`) resolves to one shared manifest, so the pre-existing upsert filter `jq --arg n "$EXT_NAME" '[.plugins[] | select(.name != $n)]'` (`:80`) finally fires across extensions. `claude plugin marketplace add "$BUILD_PARENT"` (`:104`) now always points at the stable shared parent. This is the exact structural fix the PLAN approved — algorithm unchanged, path relocated.
- **`mkdir -p`:** `mkdir -p "$BUILD_PARENT"` (`:55`) precedes the build; `mkdir -p "$MARKETPLACE_DIR"` (`:66`) precedes the manifest write. Both correct.
- **Derivations:** `BUILD_DIR="$BUILD_PARENT/$EXT_NAME"` and `MARKETPLACE_DIR="$BUILD_PARENT/.claude-plugin"` are both correct under the new parent. The `source: ("./" + $name)` line was **not** modified — correct, since the marketplace internal layout (`marketplace.json` in `.claude-plugin/`, `source` relative to the marketplace root) is preserved verbatim; only the parent moves.
- **Shell safety:** `shellcheck scripts/install-claude-plugin.sh` → clean (exit 0). `set -euo pipefail` is intact; `${CLAUDE_CONFIG_DIR:-$HOME/.claude}` is unset-var-safe under `-u`; all expansions are double-quoted; the `jq` upsert filter is correct (filter-out-by-name then append, idempotent on re-install). No quoting or unset-var bug found.

### Spec conformance (R1–R7 + 4 scenarios)

All seven requirements honored, all four scenarios pinned by tests:
- R1 coexistence → Case 2; R2 shared listing → Case 2; R3 branch-switch durability → Case 1 (off-working-tree proxy); R4 registry location → Case 1; R5 idempotent re-install → Case 3; R6 Claude-only → no Gemini/Copilot script added (correct); R7 parity-matrix → `cli-matrix.md` row 14 updated.
- Failure-path scenario → Case 4. **Sound:** the bogus name fails at `EXT_DIR` resolution (`exit 1`) *before* any `mkdir`/manifest write, so the seeded manifest is genuinely upstream of mutation — the byte-for-byte `shasum` assertion is load-bearing, not coincidental.

### Test quality (`scripts/tests/test-install-claude-plugin-marketplace.sh`)

- **Local run: 11 passed, 0 failed.** shellcheck clean (exit 0).
- **Isolation verified:** every case uses a temp `CLAUDE_CONFIG_DIR` via `mktemp -d "$TMP_ROOT/cfg.XXXXXX"` (`:94`, exported at `:107`); `TMP_ROOT` torn down by `trap … EXIT` (`:51`). The real `$HOME/.claude` is never written. The `claude` CLI is a no-op PATH stub; `jq` and the real `extensions/core/hello-world` are used — the manifest merge (the behavior under test) stays real.
- **Assertions are load-bearing, not vacuous:** Case 1 pins the out-of-tree manifest path explicitly (`cfg/local-marketplace/.claude-plugin/marketplace.json`) AND asserts no `dist-claude-plugin/` appears anywhere in the repo — a regressed in-tree script fails both. Case 2's coexistence assertion (seeded `other-ext` survives) only holds if the manifest is shared; an in-tree regression breaks it. The PR's stated negative-control reasoning holds up.

### Plan-scope fidelity

Every scoping decision matches the APPROVED PLAN v2:
- `build-claude-plugin.sh` untouched — correct (it takes an explicit output dir; relocation is the caller's concern; `:68` bare-build default stays in-tree by design).
- `docs/cli-matrix.md` row 14 only; rows 5b/13/20 untouched — **verified correct**: 5b/13 describe the build render target, 20 the gitignore carve-out, none the install location.
- `DEVELOPMENT.md` install-flow (steps 1–3) corrected to the shared-home paths; the `--plugin-dir` dev refs (`:94`) left intact — correct.
- **No stale in-tree install-flow reference remains** (grepped whole repo): all surviving `dist-claude-plugin` mentions are build-path / dev-`--plugin-dir` / gitignore / test-comment — none describe the install flow.
- CI step added to `.github/workflows/build.yml:88-89`, correctly inside the build job (has `actions/checkout@v4`).

### Convention compliance

- Gitmoji title (`🐛 …`) ✓. PR body has all four mandatory sections ✓. `Closes #360` present ✓. English-only ✓.
- **Version bump:** none required — confirmed via `git diff --name-status main...HEAD`: no `artifacts/**` skill/agent source and no `extensions/**/{SKILL,AGENT}.md` touched. Exemption correctly applied.

### Linter findings

- `shellcheck` (both `.sh` files): **clean**, exit 0.
- `markdownlint` (`DEVELOPMENT.md`, `docs/cli-matrix.md`): **clean** (and `lint-markdown` green in CI).

---

### Verdict: APPROVE

Zero findings of any class (`tech` / `arch` / `spec`). The core fix is correct and minimal, spec 0045 R1–R7 and all four scenarios are honored and pinned by 11 load-bearing assertions, the test is hermetic, scope matches the approved PLAN v2 exactly, conventions are satisfied, and CI is fully green on the reviewed head commit. Per the lifecycle termination criterion (APPROVE + zero findings + CI green), this PR is ready to merge.

PR #455, opened 2026-06-20
🐛 Fix sync-from-upstream strict-path guard and restore (spec 0059)

### LINKED ISSUE #365: 🐛 sync-from-upstream.sh strict-path handling breaks on a routine upstream advance
`scripts/sync-from-upstream.sh` cannot pull a routine upstream advance when the new/changed files live under a **strict directory** manifest entry. Two compounding defects, both reproduced on a clean fork that was 3 commits behind `main` (`merge-base..HEAD` empty on every flagged path — i.e. zero local modifications).

## Bug 1 — strict dirty-guard conflates "upstream advanced" with "local modification"

`scripts/sync-from-upstream.sh:296`

```sh
if ! git diff --quiet FETCH_HEAD -- "${spec[@]}" 2>/dev/null; then
  DIRTY+=("$path")
fi
```

This flags **any** divergence between the working tree and the freshly-fetched upstream tip. A clean fork that is merely N commits behind upstream therefore trips the guard and aborts:

```
Error: the following core-layer paths have local modifications:
  DEVELOPMENT.md
  docs
  specs
  scripts
  .github/workflows
```

…even though none of those paths were locally modified. This contradicts spec 0020 §8 ("strict — upstream-owned; a *local modification* halts the sync"). The `adopt-on-edit` arm already does the right thing via `upstream_has_blob` (upstream-history membership); the strict guard does not.

### Suggested fix

Gate the strict guard on upstream-history membership, mirroring `upstream_has_blob`: a strict blob present somewhere in `FETCH_HEAD` history is stale-but-unmodified and must **not** abort; only a working-tree blob absent from upstream history is a genuine local modification that halts the sync.

## Bug 2 — strict directory restore never instantiates new upstream files

`scripts/sync-from-upstream.sh:329`

```sh
git restore --source=FETCH_HEAD --worktree -- "${spec[@]}"
```

When `<path>` is a directory entry, `git restore` expands the directory pathspec against the **index** (tracked entries). A file that is new in the upstream tree but absent from the local index is silently skipped. Verified this session:

- `git restore --source=upstream/main --worktree -- specs ':(exclude)specs/org'` did **not** create `specs/0045-shared-plugin-marketplace.md`.
- An explicit per-file `git restore --source=upstream/main --worktree -- specs/0045-shared-plugin-marketplace.md` **did** create it.

The skipped file then masquerades as a tracked-set *deletion* against `FETCH_HEAD` on the next guard run, re-aborting the sync — so even after clearing Bug 1, the sync still fails.

### Suggested fix

For a strict **directory** entry, enumerate the upstream file set and restore each member explicitly, exactly as `reconcile_dir` already does for `adopt-on-edit` directories:

```sh
git ls-tree -r --name-only FETCH_HEAD -- "$dir/"
```

then `git restore` per member (respecting the `:(exclude)` carve-outs), instead of a single directory-pathspec `git restore` that cannot create new-in-source files.

## Impact

Net effect: the core sync mechanism cannot pull a normal upstream advance that touches a strict directory entry — i.e. the common case. Worked around manually (per-file restore from `upstream/main` after verifying no local mods, staging the new files to clear the guard's false positive, then re-running for the `adopt-on-edit` pass).

## Repro

1. Fork at commit X; let upstream advance to X+N with at least one **new** file under a strict directory entry (e.g. `specs/`, `scripts/`).
2. Run `bash scripts/sync-from-upstream.sh`.
3. Observe the spurious "local modifications" abort (Bug 1); after restoring strict paths, observe the new files are never instantiated and re-trip the guard (Bug 2).

hcross:
Fixes two bugs in `scripts/sync-from-upstream.sh` that prevent a clean fork behind upstream from syncing when strict directory entries are present (issue #365 / spec 0059).

## How to read this PR?

Start with `scripts/sync-from-upstream.sh`: the new `strict_blob_is_dirty` helper (after `path_in_org_history`), then the dirty-guard block (Bug 1), then the strict restore arm of the apply loop (Bug 2). Then read `scripts/tests/test-sync-from-upstream.sh` cases p and q at the end.

## How to test this PR?

```bash
bash scripts/tests/test-sync-from-upstream.sh
# Expected: 44 PASS, 0 FAIL
```

Reproduce the original Bug 1: create a fork at upstream commit-1, advance upstream to commit-2, run sync → should exit 0 and update the file (previously aborted with false "local modifications" error).

Reproduce the original Bug 2: create a strict directory entry, add a new file in upstream, run sync → new file must appear in the working tree (previously silently skipped).

**`scripts/sync-from-upstream.sh`** — 64 insertions, 8 deletions:
- Added `strict_blob_is_dirty <path>` helper: for present files uses `git hash-object` + `upstream_has_blob`; for absent files distinguishes locally-deleted (was in HEAD → dirty) from new-in-upstream (not in HEAD → not dirty).
- Replaced dirty-guard block: branches on `git cat-file -t FETCH_HEAD:$path` (blob vs. tree). Blob entries call `strict_blob_is_dirty`; tree entries enumerate members via `git ls-tree -r --name-only FETCH_HEAD -- $path/`, filter excluded children via `excluded_children_of`, and call `strict_blob_is_dirty` per member.
- Replaced strict restore arm: same blob vs. tree branch. Tree entries restore each upstream member individually so index-absent files are instantiated; blob entries keep the existing `pathspec_for` + `git restore` path.

**`scripts/tests/test-sync-from-upstream.sh`** — 64 insertions:
- Case p: Bug 1 regression — strict blob at older upstream version exits 0 and updates to new version.
- Case q: Bug 2 regression — strict directory with new upstream file; file is created in working tree after sync.

Closes #365

### DIFF
--- a/scripts/sync-from-upstream.sh
+++ b/scripts/sync-from-upstream.sh
@@ -173,6 +173,29 @@ path_in_org_history() {
 }
 
 # ---------------------------------------------------------------------------
+# strict_blob_is_dirty <path>
+# Return 0 (dirty) iff <path> is locally modified relative to upstream.
+# Three cases:
+#   - File present locally: dirty iff blob is absent from FETCH_HEAD history.
+#   - File absent locally, present in HEAD: locally deleted → dirty.
+#   - File absent locally, absent from HEAD: new upstream file → not dirty.
+# ---------------------------------------------------------------------------
+strict_blob_is_dirty() {
+  local path="$1" current
+  if [ -e "$REPO_DIR/$path" ]; then
+    current="$(git hash-object "$REPO_DIR/$path" 2>/dev/null)"
+    upstream_has_blob "$path" "$current" && return 1
+    return 0
+  else
+    if git cat-file -e "HEAD:$path" 2>/dev/null; then
+      return 0  # was in HEAD, now absent: locally deleted
+    else
+      return 1  # new upstream file, never committed here
+    fi
+  fi
+}
+
+# ---------------------------------------------------------------------------
 # reconcile_member <path>
 # Run the spec-0020 two-tier "modified?" decision for one adopt-on-edit member
 # file that exists in BOTH the upstream tree and the working tree, restoring it
@@ -279,9 +302,10 @@ git fetch "$CANONICAL_REPO"
 IS_SHALLOW="$(git rev-parse --is-shallow-repository 2>/dev/null || echo false)"
 
 # ---------------------------------------------------------------------------
-# Dirty-core detection: strict paths only. A local modification relative to
-# FETCH_HEAD (excluding any nested org subtree) aborts the sync. adopt-on-edit
-# and excluded paths never abort.
+# Dirty-core detection: strict paths only. Upstream-history membership
+# distinguishes "stale upstream version" from genuine local modification.
+# Directory entries are enumerated member-by-member so new-in-upstream files
+# are not treated as false positives (spec 0059 R1–R2).
 # ---------------------------------------------------------------------------
 DIRTY=()
 for i in "${!PATHS[@]}"; do
@@ -292,9 +316,26 @@ for i in "${!PATHS[@]}"; do
   # dirty against a non-existent upstream object. Skip it silently here — the
   # single warning belongs to the apply loop below.
   resolves_at_fetch_head "$path" || continue
-  mapfile -d '' spec < <(pathspec_for "$path")
-  if ! git diff --quiet FETCH_HEAD -- "${spec[@]}" 2>/dev/null; then
-    DIRTY+=("$path")
+
+  if [ "$(git cat-file -t "FETCH_HEAD:$path" 2>/dev/null)" = "tree" ]; then
+    # Directory entry: check each upstream member individually.
+    dir_dirty=0
+    while IFS= read -r member; do
+      [ -n "$member" ] || continue
+      skip=0
+      while IFS= read -r excl; do
+        case "$member" in "$excl"/*|"$excl") skip=1; break ;; esac
+      done < <(excluded_children_of "$path")
+      [ "$skip" -eq 1 ] && continue
+      if strict_blob_is_dirty "$member"; then
+        dir_dirty=1
+        break
+      fi
+    done < <(git ls-tree -r --name-only FETCH_HEAD -- "$path/" 2>/dev/null)
+    [ "$dir_dirty" -eq 1 ] && DIRTY+=("$path")
+  else
+    # Blob entry: upstream-history membership check.
+    strict_blob_is_dirty "$path" && DIRTY+=("$path")
   fi
 done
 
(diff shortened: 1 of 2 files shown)


hcross:
## Code Review — PR #455

### CI Status

| Check | Status |
|---|---|
| lint-markdown | FAIL |
| lint-specs | FAIL |
| grep-anti-patterns | pass |
| check-components | pending |
| build | pass |
| check-agents-size | pass |
| check-ci-parity | pass |
| check-extension-version-bump | pass |
| check-feedback-routing | pass |
| gitlab-ci-check | pass |

**Both `lint-markdown` and `lint-specs` fail on `specs/0059-sync-strict-path-fix.md` — a file that is NOT in this PR's diff.** That file was merged to `main` via PR #454 with pre-existing markdownlint violations (MD041, MD036 ×4). The failures are identical on `main` itself; this PR does not introduce or worsen them. They are a pre-existing baseline failure, not a regression introduced here.

Specific errors reported by CI:

```
specs/0059-sync-strict-path-fix.md:11 error MD041/first-line-heading/first-line-h1
specs/0059-sync-strict-path-fix.md:58 error MD036/no-emphasis-as-heading (Scenario 1)
specs/0059-sync-strict-path-fix.md:68 error MD036/no-emphasis-as-heading (Scenario 2)
specs/0059-sync-strict-path-fix.md:76 error MD036/no-emphasis-as-heading (Scenario 3)
specs/0059-sync-strict-path-fix.md:85 error MD036/no-emphasis-as-heading (Scenario 4)
```

`check-components` was still pending at review time; all other checks pass.

---

### Correctness

**`strict_blob_is_dirty` (Bug 1 fix) — `scripts/sync-from-upstream.sh`, line ~186**

All four cases from spec 0059 R1 are correctly handled:

- (a) File present at older upstream version: `git hash-object` produces the v1 SHA; `upstream_has_blob` walks `FETCH_HEAD` history, finds it at commit-1, returns 0 → function returns 1 (not dirty). Correct.
- (b) File present with local edit: SHA not in any upstream commit → `upstream_has_blob` returns 1 → function returns 0 (dirty). Correct.
- (c) File locally deleted: `[ -e "$REPO_DIR/$path" ]` fails; `git cat-file -e "HEAD:$path"` succeeds → return 0 (dirty). Correct.
- (d) New upstream file absent locally: `[ -e ... ]` fails; `git cat-file -e "HEAD:$path"` fails (never in org history) → return 1 (not dirty). Correct.

**Guard tree branch (R2) — dirty-guard loop, line ~320**

Enumerates via `git ls-tree -r --name-only FETCH_HEAD -- "$path/"`, filters excluded children with the same `excluded_children_of` + case-pattern logic used elsewhere, calls `strict_blob_is_dirty` per member. The early-break on first dirty member is correct and avoids unnecessary work. R2 satisfied.

**Restore tree branch (R3, R4) — apply loop, line ~375**

Uses `git ls-tree -r --name-only FETCH_HEAD -- "$path/"` to enumerate upstream members and calls `git restore --source=FETCH_HEAD --worktree -- "$member"` individually. This correctly instantiates index-absent files (Bug 2 fix). Excluded children are filtered with the identical case-pattern guard before restore is called. R3 and R4 satisfied.

**R5 (blob entries unchanged)** — the blob branch in both the guard and the restore arm uses `pathspec_for` + single `git restore` call, unchanged from pre-PR behavior. R5 satisfied.

**R6 (adopt-on-edit and excluded arms)** — neither arm appears in the diff. R6 satisfied.

---

### Convention compliance

- Commit message: `🐛 Fix sync-from-upstream strict-path guard and restore (spec 0059)` — gitmoji convention satisfied.
- PR title matches commit message — gitmoji convention satisfied.
- Changed files are under `scripts/`, not `artifacts/` — no `build-components.sh` run required.
- No skill or agent source files modified — no version bump required.
- PR body contains all four required sections (`How to read`, `How to test`, `Detailed description`, closing directive).
- `Closes #365` present in body.

**Non-blocking observation — issue #365 closed before merge**

class: tech

Issue #365 was closed manually at 2026-06-20T15:54:04Z while PR #455 is still open. AGENTS.md Rule C states: "Once the PR is merged and the changes verified, close the linked issue immediately." Closing before merge is out of order — the logbook is sealed before the evidence (a merged PR) exists. The `Closes #365` directive in the PR body would have auto-closed it on merge anyway. Since the PR body still carries `Closes #365`, GitHub will attempt another close on merge (idempotent), so there is no data loss — but the audit trail shows the issue as closed while the implementation PR is still in flight. This is a non-blocking observation; it does not require rework before merge.

---

### Test coverage

`scripts/tests/test-sync-from-upstream.sh` — cases p and q:

- **Case p** (Bug 1 regression): sets up an upstream with two commits (v1 → v2), creates an adopter at v1 content with a `strict` manifest entry, runs sync, asserts exit 0 and that the file contains `upstream v2 content`. This exactly reproduces the false-positive abort scenario. Assertion is correct.
- **Case q** (Bug 2 regression): creates an upstream with `specs/0001.md` in commit-1 and `specs/0099-new.md` added in commit-2, creates an adopter at commit-1 with a `strict` `specs` directory entry, runs sync, asserts `specs/0099-new.md` exists in the working tree. This exactly reproduces the index-absent file omission. Assertion is correct.

Local run (full suite): **44/44 PASS, 0 FAIL**. R7 satisfied.

---

### Linter findings

`lint-shell.sh` — no findings on `scripts/sync-from-upstream.sh` or `scripts/tests/test-sync-from-upstream.sh` (shellcheck clean, executable bit set, `set -e` present).

---

### Verdict: APPROVE

The implementation correctly fixes both bugs, satisfies all seven spec 0059 requirements, passes the full test suite locally, and follows project conventions. The CI failures are pre-existing baseline failures on `main` from the spec file merged in PR #454 — not introduced by this PR. The one non-blocking observation (issue closed before merge) does not require rework.

PR #522, opened 2026-07-16
🐛 Exclude org-owned specs/org/ from upstream spec-linter validation

### LINKED ISSUE #421: 🐛 spec-linter validates org-owned specs/org/ with the upstream 4-digit convention
## Problem

`scripts/lib/spec-linter.js` (run by `task spec:lint`, gated in CI `build.yml`) recursively globs **all** of `specs/` — including the org-owned overlay `specs/org/` — and enforces the upstream `^(\d{4})-<kebab-slug>\.md$` filename/id/slug contract on every file it finds.

This collides with the layer model. `specs/org/` is classified **`excluded`** (org-owned) in `.crewrig/core-paths.txt` and is documented as the place where an adopting organization authors its own specifications. But an adopter that gives its overlay a distinct numbering convention (e.g. `ORG-0001` to keep org specs visibly separate from the upstream `specs/NNNN` sequence) gets hard linter failures, and `task spec:lint` goes **red on `main`** — for content upstream neither owns nor should validate.

The contradiction in one line: **org-owned content is linted by upstream-owned tooling that only knows the upstream convention.** Worse, the adopter cannot fix it locally — `scripts/lib/spec-linter.js`, `Taskfile.yml`, and `.github/workflows/build.yml` are all `strict` core paths, so any fork-local edit halts `sync-from-upstream`. The fix can only live upstream.

## Reproduction

```sh
cat > specs/org/ORG-0001-example.md <<'SPEC'
---
id: "ORG-0001"
slug: example
status: draft
complexity: small
related-issue: 1
version: 1.0.0
---
# Example
## Intent
...
SPEC
task spec:lint
```

Result:

```
[FAIL] specs/org/ORG-0001-example.md
  - Filename does not match <NNNN>-<kebab-slug>.md or <NNNN>-<kebab-slug>.delta-<NN>.md
  - Frontmatter 'id' ("ORG-0001") does not match filename prefix ("")
  - Frontmatter 'slug' ("example") does not match filename slug ("")
```

## Evidence

- `scripts/lib/spec-linter.js` — `globSpecs()` recurses unconditionally into subdirectories (only `_template.md` / `README.md` are skipped); the validation regex is hard-coded to `^(\d{4})-([a-z0-9\-]+?)(?:\.delta-\d+)?\.md$`.
- `.crewrig/core-paths.txt` — `specs/org` is `excluded` (org-owned, nested carve-out under `strict` `specs/`).
- `docs/spec-format.md` — defines `id` as "zero-padded to four digits", i.e. the format itself does not contemplate org-overlay conventions.

## Suggested fix (pick one)

1. **(Recommended)** `globSpecs()` skips `specs/org/` entirely — the org overlay governs its own conventions, consistent with its `excluded` / org-owned classification.
2. Lint `specs/org/` with a relaxed or org-configurable ruleset (e.g. an `ORG-<NNNN>` id pattern), if upstream wants org specs validated for structure but not numbering.
3. Document that org overlays MUST follow the upstream 4-digit convention — explicitly rejects the distinct-numbering use case (least flexible).

Option 1 is the lightest and matches the principle that `excluded` paths are org-owned and untouched by core tooling.

## Context

Surfaced while running the SPECS→PLAN→DEV→REVIEW lifecycle in a downstream adopter fork (`crewrig-hcross`), where the org overlay adopted an `ORG-<NNNN>` numbering convention for `specs/org/`.

hcross:
The spec linter enforced the upstream `<NNNN>-<kebab-slug>.md` filename/frontmatter/heading contract on every file under `specs/`, including the org-owned `specs/org/` overlay, so an adopter using a distinct numbering convention there (e.g. `ORG-0001`) got hard `task spec:lint` failures on content upstream doesn't own. This PR makes the exclusion generic — driven by `.crewrig/core-paths.txt`'s `excluded` classification — instead of hardcoding `specs/org`.

Closes #421

## How to read this PR?

Three files change, in this order:

1. **`scripts/lib/spec-linter.js`** — the load-bearing file. Two new functions:
   - `loadCorePathsManifest()` reads `.crewrig/core-paths.txt` (or the path pointed to by `CREWRIG_REPO_DIR`, defaulting to the repo root) and returns `{ path, policy }` entries, mirroring the parsing rules in `scripts/check-core-paths.sh:41-54`. A missing manifest returns an empty list rather than throwing.
   - `isExcludedSpecPath(relPath, entries)` does a longest-prefix match against those entries (bounded by `/` so `specs/org2` doesn't false-match an entry for `specs/org`) and returns true only if the winning entry's policy is `excluded`.
   - Both are wired into `run()`'s semantic-validation loop as a single guard: `if (isExcludedSpecPath(file, manifestEntries)) continue;`, ahead of the existing `lintFile()` call.
2. **`scripts/tests/test-spec-linter.sh`** — five new regression cases (17-21) plus a `write_core_paths_fixture()` helper and a `CREWRIG_REPO_DIR`-override parameter added to `run_case()`.
3. **`docs/layers.md`** — one clause added to the `specs/org/` row in the core-paths table, noting the row is now also excluded from the spec linter's upstream semantic validation (spec 0071).

**Non-obvious decision worth flagging:** Cases 20-21 (the "prove the mechanism is generic, not hardcoded" scenario) build their fixture in an isolated `$TMP_ROOT/scenario4/` tree, not the shared `$TMP_ROOT/specs/` tree Cases 17-19 use. Reusing the shared tree would let Case 18's deliberately non-conforming, non-excluded fixture (`specs/9999-bad!name.md`) leak into Case 20's default-target invocation and fail it for an unrelated reason — the isolation keeps the assertion pointed at the manifest override specifically. See the comment block above Case 20 in the diff.

A one-line comment was also added above Case 17 (a post-PLAN-review fix folded into DEV per the user's instruction, rather than spun into a separate PLAN iteration): it notes Cases 17-19 depend on `specs/org` remaining classified `excluded` in the real, committed `.crewrig/core-paths.txt` (currently true — see that file's line 47), while Cases 20-21 are what actually prove the mechanism is manifest-driven rather than hardcoded to `specs/org`.

## How to test this PR?

1. `bash scripts/tests/test-spec-linter.sh` — expect `Results: 21 passed, 0 failed` (verified locally: all 21 cases, including new Cases 17-21, pass).
2. `task spec:lint` — expect `Linting passed!` against the real repo content (verified locally: ran clean against the current `specs/` tree).
3. Manual repro of the original bug (per issue #421): create a non-conforming `specs/org/ORG-0001-example.md` (wrong filename shape, `ORG-0001` id), run `task spec:lint` — it now passes for that file. Then add an equally non-conforming file outside `specs/org/` (e.g. `specs/9999-bad!name.md`) and re-run — `task spec:lint` still fails, proving the exclusion is scoped to the `excluded`-classified path, not a blanket skip.

- **The exclusion mechanism is generic, not hardcoded.** `isExcludedSpecPath()` never compares against the literal string `specs/org`; it classifies any path via `.crewrig/core-paths.txt`'s `excluded` policy through longest-prefix match. This was an explicit requirement (R5) resolved by the user at the SPECS-stage validation gate for spec 0071, not a DEV-stage judgment call. Cases 20-21 in `scripts/tests/test-spec-linter.sh` exist specifically to prove this: Case 20 excludes a path (`specs/experimental`) never referenced anywhere in `spec-linter.js`, using a manifest fixture injected via `CREWRIG_REPO_DIR`, and Case 21 is the negative control — same fixture, no override, falls back to the real repo manifest (which does not classify `specs/experimental` as excluded), and fails as expected.
- **`globSpecs()` and the `npx markdownlint` invocation are untouched.** Excluded files are still discovered by `globSpecs()` and still get passed to `markdownlint` (the `spawnSync('npx', ['markdownlint', ...uniqueFiles, ...])` call in `run()` is unchanged) — only the upstream filename/frontmatter/heading semantic contract enforced by `lintFile()` is skipped for them. Generic Markdown style linting still applies.
- **Test verification.** `bash scripts/tests/test-spec-linter.sh` → 21/21 passed (Cases 1-16 pre-existing, 17-21 new). `task spec:lint` still passes on the real repo content (99 files globbed per the `Running markdownlint-cli on 99 files...` log line).
- **No build output, version bump, or CLI-matrix update needed.** `scripts/lib/spec-linter.js` is a plain script outside `artifacts/`, not a skill/agent source (`artifacts/core/skills/*/SKILL.md` or `artifacts/core/agents/*/AGENT.md`), so no `provenance.version` bump applies. None of the three changed files (`scripts/lib/spec-linter.js`, `scripts/tests/test-spec-linter.sh`, `docs/layers.md`) match any path on the CLI-matrix trigger list in `AGENTS.md` → *CLI Matrix Maintenance* (`.claude/**`, `.gemini/**`, `artifacts/**`, `extensions/**`, `hooks/*-transcript-hooks.json`, `config/claude/**`, `config/gemini/**`, `scripts/build-components.sh`, `scripts/{build,install,setup,import,manage}-*.sh`, the `claude.yml`/`gemini.yml` workflows, `CLAUDE.md`, `GEMINI.md`, or CLI-prefixed Taskfile entries) — already verified during PLAN and re-confirmed here.
- **Realizes spec `specs/0071-org-specs-lint-exclusion.md`** (status: approved, merged to `main`), per the plan posted at https://github.com/crewrig/crewrig/issues/421#issuecomment-4994202651, cold-reviewed APPROVE at https://github.com/crewrig/crewrig/issues/421#issuecomment-4994276737.

### DIFF
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -244,7 +244,7 @@ sync.
 
 | Path | Description |
 |---|---|
-| `specs/org/` | Organization-owned specification overlay, nested in core `specs/`. Excluded from upstream sync. |
+| `specs/org/` | Organization-owned specification overlay, nested in core `specs/`. Excluded from upstream sync, and from the spec linter's upstream filename/frontmatter/heading validation (spec 0071). |
 | `docs/org/` | Organization-owned documentation overlay, nested in core `docs/`. Excluded from upstream sync. |
 | `AGENTS.org.md` | Organization-owned agent-rules extension, loaded alongside the upstream `AGENTS.md` (natively on Claude via `@` import; via the priority-66 setup deployment on Gemini and Copilot). Excluded from upstream sync. |
 
--- a/scripts/lib/spec-linter.js
+++ b/scripts/lib/spec-linter.js
@@ -39,6 +39,59 @@ function globSpecs(targetPath) {
     return result;
 }
 
+// loadCorePathsManifest() — reads .crewrig/core-paths.txt and returns its
+// entries as { path, policy } objects. Mirrors the parsing rules in
+// scripts/check-core-paths.sh:41-54: blank/`#` lines are skipped, the path is
+// the first whitespace-delimited field, the policy is the next token
+// (defaulting to `strict`). A missing manifest resolves to an empty list
+// rather than throwing, so a checkout without the file still lints normally.
+function loadCorePathsManifest() {
+    const repoDir = process.env.CREWRIG_REPO_DIR || path.join(__dirname, '..', '..');
+    const manifestPath = path.join(repoDir, '.crewrig', 'core-paths.txt');
+
+    let content;
+    try {
+        content = fs.readFileSync(manifestPath, 'utf8');
+    } catch (e) {
+        return [];
+    }
+
+    const entries = [];
+    for (const rawLine of content.split('\n')) {
+        const line = rawLine.replace(/\r$/, '');
+        const trimmed = line.trim();
+        if (trimmed === '' || trimmed.startsWith('#')) {
+            continue;
+        }
+        const match = trimmed.match(/^(\S+)(?:\s+(\S+))?/);
+        if (!match) {
+            continue;
+        }
+        entries.push({ path: match[1], policy: match[2] || 'strict' });
+    }
+    return entries;
+}
+
+// isExcludedSpecPath(relPath, entries) — true iff the manifest entry whose
+// path most closely (longest-prefix match) matches relPath classifies it
+// `excluded`. relPath must match exactly, or relPath must sit under the
+// entry's path bounded by `/` (so `specs/org2` does not match an entry for
+// `specs/org`). This is the R5 mechanism: the caller never hardcodes
+// `specs/org` — whatever the manifest classifies `excluded` under `specs/`
+// is skipped, today and for any future nested entry.
+function isExcludedSpecPath(relPath, entries) {
+    const normalized = relPath.split(path.sep).join('/');
+    let best = null;
+    for (const entry of entries) {
+        if (normalized === entry.path || normalized.startsWith(entry.path + '/')) {
+            if (!best || entry.path.length > best.path.length) {
+                best = entry;
+            }
+        }
+    }
+    return best !== null && best.policy === 'excluded';
+}
+
 function lintFile(filePath) {
     let hasErrors = false;
     const errors = [];
@@ -208,8 +261,11 @@ function run() {
     
     console.log(`Running semantic validation...`);
     let totalErrors = 0;
-    
+
+    const manifestEntries = loadCorePathsManifest();
+
     for (const file of uniqueFiles) {
+        if (isExcludedSpecPath(file, manifestEntries)) continue;
         const { hasErrors, errors } = lintFile(file);
         if (hasErrors) {
             console.error(`\n[FAIL] ${file}`);
--- a/scripts/tests/test-spec-linter.sh
+++ b/scripts/tests/test-spec-linter.sh
@@ -59,11 +59,12 @@ run_case() {
   local name="$1"
   local files="$2"
   local expected_exit="$3"
+  local env_repo_dir="${4:-}"
 
   local actual_exit=0
   local output
   # We run from TMP_ROOT so markdownlint finds .markdownlintrc
-  output=$( ( cd "$TMP_ROOT" && node "$LINTER_JS" $files 2>&1 ) ) || actual_exit=$?
+  output=$( ( cd "$TMP_ROOT" && CREWRIG_REPO_DIR="$env_repo_dir" node "$LINTER_JS" $files 2>&1 ) ) || actual_exit=$?
 
   if [ "$actual_exit" -eq "$expected_exit" ]; then
     echo "PASS  $name (exit $actual_exit)"
@@ -74,6 +75,16 @@ run_case() {
     echo "$output"
     fail=$((fail + 1))
   fi
+}
+
+# write_core_paths_fixture <dir> <content>
+# Write a standalone .crewrig/core-paths.txt fixture at <dir>, for use with
+# CREWRIG_REPO_DIR overrides (mirrors write_manifest() in
+# scripts/tests/test-check-core-paths.sh:63-67).
+write_core_paths_fixture() {
+  local dir="$1" content="$2"
+  mkdir -p "$dir/.crewrig"
+  printf '%s' "$content" > "$dir/.crewrig/core-paths.txt"
 }
 
 # -------------------------------------------------------------------------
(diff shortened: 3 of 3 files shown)


hcross:
## Cold review — PR #522 (issue #421, spec 0071)

**Note:** GitHub's API rejects `APPROVE`/`REQUEST_CHANGES` review events on a pull request authored by the same account performing the review ("Can not approve your own pull request"). The authenticated `gh` identity in this environment is the PR's author, so this review is posted as a plain `COMMENT` event. The verdict below is nonetheless a full **APPROVE**.

Independent review, no authoring context. Verified against `specs/0071-org-specs-lint-exclusion.md` (approved, on `main`), the PLAN comment (https://github.com/crewrig/crewrig/issues/421#issuecomment-4994202651), and the cold PLAN review (https://github.com/crewrig/crewrig/issues/421#issuecomment-4994276737) — not taken at face value, checked against the actual diff and by running the tests myself in the PR's own head commit (`6513d37`).

### CI status

All 12 required checks **pass**: `build`, `lint-specs`, `lint-markdown`, `check-components`, `check-ci-parity`, `check-feedback-routing`, `check-skill-versions`, `check-agents-size`, `check-extension-version-bump`, `grep-anti-patterns`, `gitlab-ci-check`, `test-harness-curate`. (`check-components` was still `pending` at the moment I first polled; it resolved to `pass` on a follow-up check.)

### Correctness — all 5 requirements verified against the diff

- **R1/R2/R4** — `scripts/lib/spec-linter.js` `run()` gains exactly one guard, `if (isExcludedSpecPath(file, manifestEntries)) continue;`, inserted as the first statement of the semantic-validation `for` loop, ahead of the existing `lintFile(file)` call. `globSpecs()` (lines 15-37) and the `spawnSync('npx', ['markdownlint', ...uniqueFiles, '-c', '.markdownlintrc'], ...)` call are byte-for-byte unchanged — excluded files are still discovered and still pass through generic Markdown style linting; only the upstream filename/id/slug/heading contract in `lintFile()` is skipped for them. Matches the spec's "excluded from upstream validation, not from markdownlint" distinction.
- **R3** — Case 19 (`scripts/tests/test-spec-linter.sh`) targets `specs/org` explicitly and still expects exit 0, proving the exclusion survives explicit path targeting, not just default recursive discovery.
- **R5 (the load-bearing one) — confirmed, not hardcoded.** Grepped `scripts/lib/spec-linter.js` for the literal string `specs/org`: it appears only in three comment lines (explaining the mechanism), never in executable code. `isExcludedSpecPath()` and `loadCorePathsManifest()` derive the exclusion purely from `.crewrig/core-paths.txt`'s `excluded` classification via longest-prefix match. Cases 20/21 prove this with a path (`specs/experimental`) that appears nowhere else in `spec-linter.js`, injected only via a `CREWRIG_REPO_DIR`-pointed fixture manifest, with Case 21 as a negative control (same fixture, no override → falls back to the real manifest, which doesn't classify that path → exit 1). I re-ran all 21 cases myself (see Test coverage below).
- **`/`-bounded matching, verified empirically, not just by reading.** `isExcludedSpecPath()` matches on `normalized === entry.path || normalized.startsWith(entry.path + '/')`, so `specs/organization/...` cannot false-match the `specs/org` entry (the character after `specs/org` in `specs/organization` is `a`, not `/`). I did not just trust the code reading — I built a standalone fixture repro (`.crewrig/core-paths.txt` classifying only `specs/org` as `excluded`, plus a non-conforming `specs/organization/9999-bad!name.md`) and ran the real linter against it: it correctly reported the filename/frontmatter/heading violations and exited 1, i.e. `specs/organization` is NOT swept in by the `specs/org` exclusion.

### Convention compliance

- PR title `🐛 Exclude org-owned specs/org/ from upstream spec-linter validation` — Gitmoji, no conventional-commit prefix. Matches `AGENTS.md` → *Naming Convention*.
- PR body has all four mandated sections (2-sentence intro, `## How to read this PR?`, `## How to test this PR?`, `## Detailed description (for agents)`) plus `Closes #421`. Matches `AGENTS.md` → *Pull Request Format*.
- Branch `feat/0421-org-specs-lint-exclusion`, worktree `.worktrees/421`, single commit ahead of `main` — matches the `<prefix>/<NNNN>-<slug>` convention and the Pre-Edit Guard's three conditions.
- `docs/layers.md:247`'s new clause is consistent with `docs/layers.md:54` (the immutability/overlay row talks about sync carve-out, not linting; no contradiction).
- No skill/agent source under `artifacts/**` touched, so no `provenance.version` bump applies; none of the three changed files (`docs/layers.md`, `scripts/lib/spec-linter.js`, `scripts/tests/test-spec-linter.sh`) match any CLI-matrix trigger path in `AGENTS.md` → *CLI Matrix Maintenance* — correctly out of scope, and `check-ci-parity`/`check-components` passing corroborates this.
- **Minor process nit (non-blocking).** `class: tech`. Per `AGENTS.md` → *Logbook Issues* Rule B, a "significant... decision" must be posted as an issue comment *before* resuming work on its resolution. The PR body states that Finding 1 from the cold PLAN review (the `specs/org` classification-coupling note) was "folded into DEV per the user's instruction, rather than spun into a separate PLAN iteration" — that's exactly this kind of decision, but issue #421 has only 3 comments (SPECS, PLAN, PLAN-review) and no DEV-stage logbook comment recording it. Nothing is lost (the reasoning is fully captured in the PR body and in the code comment above Case 17), so this doesn't block merge, but it's a letter-of-the-rule gap worth naming.

### Test coverage

- Ran `bash scripts/tests/test-spec-linter.sh` myself on the PR's head commit (`6513d37`, matches `headRefOid` reported by `gh pr view`): `Results: 21 passed, 0 failed` — all of Cases 1-16 (pre-existing) and the five new Cases 17-21.
- Ran `task spec:lint` myself against the real repo tree: `Running markdownlint-cli on 99 files... Running semantic validation... Linting passed!` — confirms R2 holds on live content (today `specs/org/` only has `README.md`, already excluded by filename, so this is a defensive fix, matching the plan's own framing — not a live-bug repro).
- Scenario4 fixture isolation checked and correctly built, not just claimed: `$TMP_ROOT/scenario4/` copies `.markdownlintrc` from `$ROOT_DIR` and symlinks `$ROOT_DIR/node_modules` (mirroring the same setup the top-level `$TMP_ROOT` gets at script start), so Case 20 does not fail for the wrong reason (missing markdownlint config/deps) — it genuinely exercises the manifest-driven mechanism, isolated from Case 18's deliberately non-conforming, non-excluded `specs/9999-bad!name.md` fixture in the shared tree.

### Linter findings

- `shellcheck scripts/tests/test-spec-linter.sh` — one pre-existing `SC2086` info-level note on the unquoted `$files` expansion (line 67), present before this PR (the diff only prepends `CREWRIG_REPO_DIR="$env_repo_dir"` to that line; the unquoted `$files` is required as-is so an empty `files` value word-splits away rather than being passed as an empty-string arg). Not attributable to this diff.
- `markdownlint docs/layers.md -c .markdownlintrc` — clean, exit 0.
- No `*.py`, `*.json`, or `SKILL.md` files changed — those linters don't apply.

### Verdict: **APPROVE**

All 5 spec requirements verified against the diff and, where feasible, verified empirically (not just read) — including the `/`-bounded prefix-match behavior via a standalone repro. Tests re-run clean (21/21, plus `task spec:lint`). One non-blocking `class: tech` process note above; it doesn't change the verdict.

PR #533, opened 2026-07-17
✅ Fix false-negative timeout checks in transcript-hook test and wire it into CI

### LINKED ISSUE #530: 🐛 test-mempalace-transcript-hook.sh: issue-90/issue-94 checks broken since portable-timeout refactor
## Problem

\`scripts/tests/test-mempalace-transcript-hook.sh\` has two checks (\`issue-90\` — Python invocation guarded by timeout wrapper, \`issue-94\` — explicit \`timeout <n>\` prefix) that currently **FAIL** on \`main\`, unrelated to any change since their introduction.

## Root cause

Commit \`799ff15\` (PR #210/#211, "portable timeout detection") replaced the hook's literal \`timeout\` invocation in \`hooks/mempalace-transcript.sh\` with a \`\${_HOOK_TIMEOUT:+...}\` variable indirection (to support \`gtimeout\` on macOS and graceful degradation when neither binary is present). Both regression checks still grep for the literal, case-sensitive string \`timeout\` immediately preceding \`\$MEMPALACE_PYTHON\` in the script source — a pattern the refactor's indirection no longer matches, even though the underlying guarantee (the Python call IS wrapped with a timeout when one is available) still holds at runtime.

## Impact

- Both checks have been silently broken since PR #211 merged.
- \`scripts/tests/test-mempalace-transcript-hook.sh\` is not wired into any \`.github/workflows/*.yml\` job, so no CI surface ever caught this drift — confirmed no workflow references this script.
- Discovered incidentally while implementing spec 0073 (issue #508): confirmed via \`git stash\` that these two checks fail identically on unmodified \`main\`, before that ticket's changes.

## Suggested fix

Update the two checks' matching logic to assert the *effective* guarantee (a timeout mechanism guards the Python call whenever \`timeout\`/\`gtimeout\` is resolved) rather than grepping for a literal \`timeout \` substring — for example, by asserting on the \`_HOOK_TIMEOUT\` variable's resolution logic and its use in the invocation line, or by asserting behaviorally (fixture a slow-hanging fake Python binary and confirm the wrapped call is actually killed within the expected bound).

## Evidence

- \`hooks/mempalace-transcript.sh\` — the \`\${_HOOK_TIMEOUT:+\$_HOOK_TIMEOUT 5}\` invocation, introduced by commit \`799ff15\`.
- \`scripts/tests/test-mempalace-transcript-hook.sh\` — the \`issue-90\` and \`issue-94\` check bodies (literal-string grep).

## Notes

Not in scope for spec 0073 / issue #508 (the transcript-hook ChromaDB-lock fix) — flagged here as its own ticket per that ticket's DEV-stage findings, to avoid mixing an unrelated pre-existing bug fix into that diff.

gfourny-sfeir:
Fixes the two stale timeout regression checks (`issue-90`, `issue-94`) in `test-mempalace-transcript-hook.sh` that PR #211 turned into false negatives, and wires the test into CI so the gap that hid this can't recur.

## How to read this PR?

Two files, read in this order:

1. `scripts/tests/test-mempalace-transcript-hook.sh` — the core change. Test 1 (`issue-90`) becomes a **structural** assertion of the resolved `${_HOOK_TIMEOUT:+... 5}` wrapper; test 5 (`issue-94`) becomes a **behavioral** runtime proof (fake slow Python + wall-clock) with a graceful **SKIP** when no `timeout`/`gtimeout` binary is present. A `skip` counter is added to `record()` and the Summary.
2. `.github/workflows/build.yml` — a single new step wiring the test into the `check-components` job.

Key design point: PR #211 changed the guarantee from *"timeout always present"* to *"timeout present when a binary is available, else a deliberate logged degradation"*. The tests now pin two distinct facets (wrapper **applied** vs guard **fires at runtime**) instead of re-matching a literal `timeout ` keyword that no longer exists. `hooks/mempalace-transcript.sh` is intentionally **not** modified — its behavior is correct.

## How to test this PR?

```bash
git fetch origin fix/0530-transcript-hook-timeout-test
git checkout fix/0530-transcript-hook-timeout-test
bash scripts/tests/test-mempalace-transcript-hook.sh; echo "EXIT=$?"
```

Expected:
- On a host **with** GNU `timeout` (or `gtimeout`, e.g. CI ubuntu-latest): `Summary: 6 passed, 0 failed, 0 skipped`, `EXIT=0` — test 5 runs behaviorally and proves the 5s guard fires.
- On a host **without** either binary (e.g. bare macOS): `Summary: 5 passed, 0 failed, 1 skipped`, `EXIT=0` — test 5 SKIPs, hang-free.

Sanity-check the new structural regex matches the shipped hook line:

```bash
grep -nE '\$\{_HOOK_TIMEOUT:\+[^}]*[0-9]+[^}]*\}[[:space:]]+"\$MEMPALACE_PYTHON"' hooks/mempalace-transcript.sh
```

**Root cause.** PR #211 (`799ff15`) replaced the literal `timeout 5 "$MEMPALACE_PYTHON"` with runtime detection (`timeout` → `gtimeout` → logged fallback) and the `${_HOOK_TIMEOUT:+$_HOOK_TIMEOUT 5}` parameter-expansion wrapper (`hooks/mempalace-transcript.sh:158`). Tests 1 and 5 still grepped the literal `timeout ` prefix, which no longer matches → both FAIL despite an intact runtime guarantee. The failure was invisible because the test is wired into no CI workflow.

**Changes:**
- `scripts/tests/test-mempalace-transcript-hook.sh`
  - Header comments (`#90`, `#94`) reframed to the PR #211 contract.
  - `skip=0` added; `record()` gains a `SKIP` branch (increments neither pass nor fail); Summary prints `$pass passed, $fail failed, $skip skipped`. Exit logic unchanged (exit 1 iff `$fail > 0`).
  - Test 1 (`issue-90`): grep rewritten to assert `${_HOOK_TIMEOUT:+... <n>}` fronts `"$MEMPALACE_PYTHON"` — deterministic, no binary dependency.
  - Test 5 (`issue-94`): entire static grep/awk block replaced by a behavioral test (fake Python draining stdin then `sleep 15`, fed a `Stop` event, wall-clock via `date +%s`). PASS if elapsed ≤ 8s when a timeout binary exists; SKIP (fast, no sleep) otherwise. Traps extended so `TMPDIR_T5` is cleaned alongside `TMPDIR_T2`/`SANDBOX_T6`.
- `.github/workflows/build.yml`: new step "Test mempalace-transcript hook regression (issue #530)" in the `check-components` job, registered in `ci/ci-capabilities.yml` and mirrored into `.gitlab-ci.yml` via `scripts/build-ci.sh` (ubuntu-latest provides `jq`, `python3`, GNU `timeout`, so all 6 tests execute for real).

**Out of scope / follow-up:** 28 of 47 `scripts/tests/*.sh` are unwired from CI (mostly `test-e2e-*`). A separate issue tracks auditing that gap.

**Spec-0049 parity:** the capability manifest `ci/ci-capabilities.yml` (source of truth) carries the new command; `.gitlab-ci.yml` was regenerated to match. `check-ci-parity` is green.

**Lifecycle:** `tech`-class (test correctness + CI wiring, no *WHAT* change) → no spec-PR. No skill/agent source touched → no version bump. `build.yml` is outside the CLI-matrix trigger list → no `docs/cli-matrix.md` update.

Closes #530

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,6 +121,9 @@ jobs:
       - name: Test build-docs-index regression (spec 0027)
         run: bash scripts/tests/test-build-docs-index.sh
 
+      - name: Test mempalace-transcript hook regression (issue #530)
+        run: bash scripts/tests/test-mempalace-transcript-hook.sh
+
   lint-markdown:
     runs-on: ubuntu-latest
     steps:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -62,6 +62,7 @@ check-components:
     - "bash scripts/tests/test-artifact-build-install-scope.sh"
     - "bash scripts/build-docs-index.sh --check"
     - "bash scripts/tests/test-build-docs-index.sh"
+    - "bash scripts/tests/test-mempalace-transcript-hook.sh"
   rules:
     - if: '$CI_PIPELINE_SOURCE == "merge_request_event" && $CI_MERGE_REQUEST_TARGET_BRANCH_NAME =~ /^main$|^release\//'
     - if: '$CI_PIPELINE_SOURCE == "push" && $CI_COMMIT_BRANCH =~ /^main$|^release\//'
--- a/ci/ci-capabilities.yml
+++ b/ci/ci-capabilities.yml
@@ -71,6 +71,7 @@ capabilities:
       - bash scripts/tests/test-artifact-build-install-scope.sh
       - bash scripts/build-docs-index.sh --check
       - bash scripts/tests/test-build-docs-index.sh
+      - bash scripts/tests/test-mempalace-transcript-hook.sh
 
   - id: lint-markdown
     name: "Lint Markdown files"
--- a/scripts/tests/test-mempalace-transcript-hook.sh
+++ b/scripts/tests/test-mempalace-transcript-hook.sh
@@ -209,13 +209,18 @@ EOF
   # invoked (guards against a regression that stops the Stop event from
   # reaching the Python call — which would return in ~0s and otherwise
   # PASS a bare upper bound, a silent false-positive). The lower bound
-  # (>=4s) proves the 5s budget was genuinely waited out, and the upper
-  # bound (<=8s) proves the guard then fired instead of hanging 15s.
-  if [ -f "$MARKER_T5" ] && [ "$ELAPSED_T5" -ge 4 ] && [ "$ELAPSED_T5" -le 8 ]; then
+  # (>=4s) proves the 5s budget was genuinely waited out; the upper bound
+  # proves the guard then fired instead of hanging for the full 15s sleep.
+  # The ceiling is set to 12s (not a tight 8s): the discriminating property
+  # is "well before the 15s sleep", so the extra headroom absorbs fork/exec
+  # and scheduling jitter on a loaded CI runner without weakening the test
+  # (a non-firing guard still lands at ~15s > 12s). The >=4s floor stays the
+  # meaningful half.
+  if [ -f "$MARKER_T5" ] && [ "$ELAPSED_T5" -ge 4 ] && [ "$ELAPSED_T5" -le 12 ]; then
     record PASS "issue-94: timeout guard fires at runtime (5s budget)"
   else
     record FAIL "issue-94: timeout guard fires at runtime (5s budget)" \
-      "marker=$([ -f "$MARKER_T5" ] && echo yes || echo no) elapsed=${ELAPSED_T5}s (want marker=yes, 4<=elapsed<=8: the slow Python must be reached AND the 5s timeout must fire)"
+      "marker=$([ -f "$MARKER_T5" ] && echo yes || echo no) elapsed=${ELAPSED_T5}s (want marker=yes, 4<=elapsed<=12: the slow Python must be reached AND the 5s timeout must fire before the 15s sleep)"
   fi
 else
   record SKIP "issue-94: timeout guard fires at runtime (5s budget)" \
--- a/scripts/tests/test-mempalace-transcript-hook.sh
+++ b/scripts/tests/test-mempalace-transcript-hook.sh
@@ -185,9 +185,11 @@ TMPDIR_T5="$(mktemp -d)"
 trap 'rm -rf "$TMPDIR_T2" "$TMPDIR_T5"' EXIT
 
 if command -v timeout >/dev/null 2>&1 || command -v gtimeout >/dev/null 2>&1; then
+  MARKER_T5="$TMPDIR_T5/python-was-called"
   SLOW_PY="$TMPDIR_T5/slow-python"
-  cat > "$SLOW_PY" <<'EOF'
+  cat > "$SLOW_PY" <<EOF
 #!/bin/bash
+touch "$MARKER_T5"          # prove the hung Python path was actually reached
 cat >/dev/null   # drain the heredoc so it does not deadlock
 sleep 15
 echo "OK"
@@ -203,11 +205,17 @@ EOF
   ) || true
   ELAPSED_T5=$(( $(date +%s) - START_T5 ))
 
-  if [ "$ELAPSED_T5" -le 8 ]; then
+  # Two-sided assertion: the marker proves the slow Python was actually
+  # invoked (guards against a regression that stops the Stop event from
+  # reaching the Python call — which would return in ~0s and otherwise
+  # PASS a bare upper bound, a silent false-positive). The lower bound
+  # (>=4s) proves the 5s budget was genuinely waited out, and the upper
+  # bound (<=8s) proves the guard then fired instead of hanging 15s.
+  if [ -f "$MARKER_T5" ] && [ "$ELAPSED_T5" -ge 4 ] && [ "$ELAPSED_T5" -le 8 ]; then
     record PASS "issue-94: timeout guard fires at runtime (5s budget)"
   else
     record FAIL "issue-94: timeout guard fires at runtime (5s budget)" \
-      "hook ran ${ELAPSED_T5}s — the 5s timeout did not fire"
+      "marker=$([ -f "$MARKER_T5" ] && echo yes || echo no) elapsed=${ELAPSED_T5}s (want marker=yes, 4<=elapsed<=8: the slow Python must be reached AND the 5s timeout must fire)"
   fi
 else
   record SKIP "issue-94: timeout guard fires at runtime (5s budget)" \
--- a/scripts/tests/test-mempalace-transcript-hook.sh
+++ b/scripts/tests/test-mempalace-transcript-hook.sh
@@ -4,8 +4,10 @@
 # Pins the contracts surfaced by issues #90–#94:
 #
 #   #90 — Direct Python import causes SQLite contention.
-#         The Python invocation MUST be wrapped by `timeout` (or equivalent
-#         guard) so a hung MemPalace lock cannot stall the calling CLI.
+#         The Python invocation MUST be guarded by the resolved timeout
+#         wrapper (`${_HOOK_TIMEOUT:+... 5}`, resolved at init to `timeout`
+#         or `gtimeout`) so a hung MemPalace lock cannot stall the calling
+#         CLI. Checked structurally against that wrapper shape (PR #211).
 #
 #   #91 — Hook fires on every PostToolUse — too frequent for parallel agents.
 #         When the hook event is `PostToolUse`, the script MUST exit 0
@@ -21,8 +23,13 @@
 #         `2>&1` — that hides import errors and MemPalace failures from
 #         the log line.
 #
-#   #94 — No timeout (explicit `timeout` keyword check).
-#         The Python invocation line MUST be prefixed with `timeout `.
+#   #94 — No timeout guard at runtime.
+#         The timeout applied to the Python call is now conditional and
+#         portable (PR #211): the guard fires when a `timeout`/`gtimeout`
+#         binary is available, and degrades to a deliberate, logged no-op
+#         when neither is on PATH. The check is behavioral — it proves the
+#         guard actually kills a hung Python within the 5s budget — and
+#         SKIPs gracefully on hosts where no timeout binary exists.
 #
 #   spec 0073 / issue #508 — Route through the shared ChromaDB HTTP daemon.
 #         The heredoc's Python payload MUST route through
@@ -50,6 +57,7 @@ fi
 
 pass=0
 fail=0
+skip=0
 
 record() {
   local outcome="$1"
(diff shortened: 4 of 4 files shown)


gfourny-sfeir (commented):
## Independent cold review — PR #533

Verdict: **REQUEST CHANGES**

### CI status

| Check | State |
|---|---|
| `check-ci-parity` | **FAIL** (blocker) |
| `check-components` | pending |
| `build`, `check-agents-size`, `check-extension-version-bump`, `grep-anti-patterns`, `check-feedback-routing`, `check-skill-versions`, `gitlab-ci-check`, `lint-markdown`, `lint-specs`, `test-harness-curate` | pass |

A required check (`check-ci-parity`) is failing. Per the review protocol a failing required check is a hard blocker and overrides any editorial LGTM — the failure is directly caused by this PR, not an unrelated flake. Reproduced locally (`bash scripts/check-ci-parity.sh` → exit 1).

### Correctness

The test-file changes are correct on their own terms:

- Test 1 (`issue-90`) — the new regex `${_HOOK_TIMEOUT:+[^}]*[0-9]+[^}]*}[[:space:]]+"$MEMPALACE_PYTHON"` matches the shipped wrapper at `hooks/mempalace-transcript.sh:158`. Verified: local run reports PASS.
- Test 5 (`issue-94`) — behavioral runtime proof with graceful SKIP when no `timeout`/`gtimeout` binary is present. `record()` gains a `SKIP` branch that increments neither `pass` nor `fail`; the exit contract (`exit 1` iff `fail>0`) is preserved, so a SKIP does not fail the build. Correct.
- Trap handling — `TMPDIR_T5` is created via `mktemp -d` before the branch and cleaned by the final `trap` (which supersedes the intermediate one and covers all three dirs). No leak.
- Local run on this host (no `timeout`): `Summary: 5 passed, 0 failed, 1 skipped`, exit 0. Matches the PR's stated expectation. **class: tech**

Residual note (non-blocking): the behavioral branch of test 5 is only exercised on a host that has a timeout binary — i.e. the `check-components` CI job, which is still **pending**. Its real-runtime pass is therefore not yet observed. **class: tech**

### Convention compliance

- **[BLOCKING] CI parity (spec 0049) not maintained.** The new step landed in the `check-components` job (build.yml lines 28-126; step at line 125), whose portable capability declares its command list in `ci/ci-capabilities.yml` (`id: check-components`, ending at `bash scripts/tests/test-build-docs-index.sh`). Adding a `run:` step to that job without adding the matching entry to the capability's `command:` list is an R3 divergence:
  `DRIFT: capability 'check-components' (github-actions): business step diverges from the declared command — unexpected step: 'bash scripts/tests/test-mempalace-transcript-hook.sh' (R3)`
  Fix: add `- bash scripts/tests/test-mempalace-transcript-hook.sh` to the `check-components` `command:` list in `ci/ci-capabilities.yml`, then regenerate `.gitlab-ci.yml` (its `check-components` `script:` block likewise ends at `test-build-docs-index.sh` and would otherwise drift on the reference↔GitLab / GHA↔GitLab arms). Run `bash scripts/check-ci-parity.sh` until green. **class: tech**
- **[NON-BLOCKING] PR body inaccuracy.** The body states the step wires the test "into the `build` job" and reads "a single new step wiring the test into the `build` job." It is actually inside the `check-components` job (`build` is lines 12-27; the step is at line 125, inside `check-components`, 28-126). Correct the description so reviewers/agents parsing this PR aren't misled — and note the parity contract omission was a direct consequence of misattributing the target job. **class: tech**
- Gitmoji title (`✅ …`), required PR body sections, branch naming (`fix/0530-transcript-hook-timeout-test`), no version bump (no skill/agent source touched), and the cli-matrix exemption claim (build.yml is not in the CLI-matrix trigger list) are all correct.

### Test coverage

This PR *is* test-quality work, and it improves on the prior state: the old static `timeout ` grep (invalidated by PR #211's `${_HOOK_TIMEOUT:+... 5}` wrapper) is replaced by a structural assertion (test 1) plus a behavioral runtime proof (test 5), and the previously-unwired test is added to CI. No coverage regression. The only gap is that the CI wiring itself is not yet parity-clean (see above).

### Linter findings

`lint-shell.sh scripts/tests/test-mempalace-transcript-hook.sh` → exit 0. `shellcheck` absent on the review host, so static analysis was skipped (degraded path); executable-bit and `set -e` checks passed. No shell findings surfaced.

---

Blocking summary: one `tech`-class finding (spec 0049 CI-parity manifest not updated) causing the `check-ci-parity` required check to fail. Resolve the parity gap in `ci/ci-capabilities.yml` (+ regenerate `.gitlab-ci.yml`) and fix the PR-body job-name inaccuracy. The test-logic changes themselves are sound and ready.

gfourny-sfeir:
## Independent cold review — PR #533 @ b90c72a

**Verdict: APPROVE** — zero blocking findings. One non-blocking `tech`-class nit below.

### CI status
All 12 required checks **PASS** — none failing, none pending. Notably `check-components` (runs the test with GNU `timeout`, so all 6 assertions execute for real), `check-ci-parity`, and `gitlab-ci-check` are green.

### Correctness
- **Test 1 (issue-90)** — the new structural regex `${_HOOK_TIMEOUT:+[^}]*[0-9]+[^}]*} "$MEMPALACE_PYTHON"` matches the shipped hook line `hooks/mempalace-transcript.sh:158` (`${_HOOK_TIMEOUT:+$_HOOK_TIMEOUT 5} "$MEMPALACE_PYTHON" ...`). Verified against the file at head. Correctly replaces the literal `timeout ` grep that PR #211 invalidated.
- **Test 5 (issue-94)** — the behavioral test genuinely exercises the guard: a `{"hook_event_name":"Stop"}` event flows past the PostToolUse gate (`hooks/mempalace-transcript.sh:78`), sets `CONTENT` at line 105, and reaches the timeout-wrapped Python call at line 158. On CI (`timeout` present) the 5s guard kills the 15s fake and elapsed lands under the 8s ceiling. On hosts without `timeout`/`gtimeout` it SKIPs before creating the slow fake — hang-free, matching the hook's own logged-degradation fallback (lines 44-47). Confirmed locally: `5 passed, 0 failed, 1 skipped`, exit 0.
- **`record()` / Summary** — the `SKIP` branch increments only `skip`; exit logic is unchanged (`exit 1` iff `fail > 0`), so a SKIP never masks a failure and never fails the build.
- **Temp-dir cleanup** — `TMPDIR_T5` is created unconditionally (line 184) before the timeout branch, and the final `trap` (line 233) rewrites the EXIT handler to clean `TMPDIR_T2`, `TMPDIR_T5`, and `SANDBOX_T6` together. No leak, no dangling reference.

### Convention compliance (AGENTS.md)
- **Gitmoji title** — `✅ Fix false-negative timeout checks…` conforms.
- **PR body** — all four mandatory sections present.
- **Version bump** — no skill/agent source under `artifacts/**` touched; no bump required. `check-skill-versions` green confirms.
- **CLI matrix** — none of the four changed paths is on the trigger list (`build.yml` ≠ `claude.yml`/`gemini.yml`; `scripts/tests/*` ≠ `scripts/{build,install,setup,import,manage}-*.sh`). No `docs/cli-matrix.md` update owed.
- **Spec-0049 parity** — the capability was added to the source-of-truth `ci/ci-capabilities.yml`, `.gitlab-ci.yml` regenerated to match, and `bash scripts/check-ci-parity.sh` returns green ("reference, GitHub Actions, GitLab agree"). Working tree is clean — no un-regenerated drift.

### Test coverage
The change strengthens coverage: it removes two false negatives and wires a previously-orphaned regression suite into both CI backends. Assertions are meaningful (structural + behavioral), not coverage theater.

**Non-blocking nit — `class: tech`** (`scripts/tests/test-mempalace-transcript-hook.sh:206`): Test 5 asserts only an *upper* bound (`ELAPSED_T5 -le 8`) and has no positive proof that the slow Python was actually invoked — unlike Test 2, which drops a marker file. Today the Stop event provably reaches the Python call, so the test is sound. But if a future refactor broke the Stop-event gating so Python were never spawned, the hook would return in ~0s and this test would still PASS — a silent false-positive. Tightening to a marker-file assertion plus a lower bound (e.g. `4 <= ELAPSED_T5 <= 8`) would prove both that the guard fired *and* that the hung path was genuinely reached. Not blocking — the assertion is correct for current behavior and strictly better than what it replaces.

### Linter findings
`lint-shell.sh` degraded gracefully (`shellcheck` absent on the review host) — no static-analysis findings surfaced locally. CI's `check-components` executed the suite green.

hcross (approved):
## Cold review — APPROVE ✅

Independent review (spawned with only the PR number; the verdict was verified first-hand, not just read off the PR body).

### What was verified

- **Bug reproduced on `main`:** the pre-fix test yields `3 passed, 2 failed` — `issue-90` and `issue-94` fail exactly as #530 reports.
- **Test 1 regex matches the shipped hook:** the new structural pattern matches the live line in `hooks/mempalace-transcript.sh` → `${_HOOK_TIMEOUT:+$_HOOK_TIMEOUT 5} "$MEMPALACE_PYTHON"`. The trailing `[0-9]+` pins a numeric budget, so a partial fix dropping the `5` would fail it.
- **New test executed in both host modes:** with a `timeout`/`gtimeout` binary present → `issue-94` PASSes behaviorally (the guard kills the fake slow Python within the 5s budget); with neither binary on `PATH` → `issue-94` SKIPs cleanly, no hang.
- **Trap handling correct:** `trap … EXIT` replaces rather than accumulates, and the final trap covers `TMPDIR_T2`, `TMPDIR_T5`, and `SANDBOX_T6` — no temp-dir leak.
- **CI parity:** the new command is present in `ci/ci-capabilities.yml` (source of truth), the regenerated `.gitlab-ci.yml`, and the `check-components` job of `build.yml`. `check-ci-parity` is green.
- **Conventions:** Gitmoji title, conformant PR body, English-only, and a correct `tech`-class self-classification (no spec-PR, no version bump, no `docs/cli-matrix.md` update). 12/12 checks green.

The two-sided assertion in Test 5 (marker present **and** `4 ≤ elapsed ≤ 8`) is well-engineered: the marker closes the silent-false-positive hole a bare upper bound would leave.

### Non-blocking nits (do not gate merge)

1. **Tight upper timing bound** (`scripts/tests/test-mempalace-transcript-hook.sh`, Test 5) — `elapsed ≤ 8` leaves only ~3s of headroom over the 5s budget. The discriminating property is really "well before the `sleep 15`", not "≤ 8s"; widening the ceiling (e.g. `≤ 12`) would cut flakiness risk on a loaded CI runner while still proving the guard fired. The `≥ 4` lower bound stays the meaningful half.
2. **No `Closes #530` keyword** — the body says "Fixes…" in prose but carries no GitHub reference, so the merge won't auto-close logbook #530. Please close it manually post-merge (AGENTS.md Rule C).

**Verdict: APPROVE.** Clean, well-verified work.

gfourny-sfeir:
Thanks @hcross — both non-blocking nits addressed in `4e2f70f`:

1. **Timing ceiling widened `8s → 12s`** (`scripts/tests/test-mempalace-transcript-hook.sh`, Test 5). You're right that the discriminating property is "well before the 15s sleep", not a tight 8s. `[4, 12]` keeps the `>=4s` floor as the meaningful half (proves the 5s budget was waited out) while giving ~7s of headroom for fork/exec + scheduler jitter on a loaded runner; a non-firing guard still lands at ~15s > 12s, so the separation holds. Verified in both host modes (SKIP without a timeout binary; PASS at elapsed ~5s with one).
2. **`Closes #530`** added to the PR body so the merge auto-closes the logbook issue (belt-and-suspenders with AGENTS.md Rule C).

Issue #542, opened 2026-07-18
🧪 e2e README files exceed the 200-line budget (test-e2e-libs-readme / test-e2e-readme)


hcross:
Surfaced by #534 while wiring previously-unrun tests into CI.

## Problem
`scripts/tests/test-e2e-libs-readme.sh` and `scripts/tests/test-e2e-readme.sh` assert a 200-line budget on e2e README files, but:
- `tests/e2e/lib/README.md` is 203 lines
- `tests/e2e/README.md` is 289 lines

Both are over budget, so the two tests fail. They were never caught because the tests ran in no CI workflow (the root cause #534 fixes).

## Current state
Both tests are **quarantined** via `ci/test-wiring-exemptions.txt` (spec 0076 registration guard) so the systemic guard could ship green. This issue tracks the real fix.

## Fix options
- Trim the two READMEs back under budget, or
- Revise the budget assertion in the tests if 200 lines is no longer the intended ceiling.

Once fixed and the tests pass, remove their entries from `ci/test-wiring-exemptions.txt` and wire them into `check-components` (the guard flags redundant exemptions).

hcross:
## PLAN — issue #542 (spec 0077)

**Goal.** Bring `tests/e2e/README.md` (289 lines) and `tests/e2e/lib/README.md`
(203 lines) back within their locked 200-line budget by prose-tightening only,
then wire their two content-guarantee tests into CI and lift the quarantine so
the guarantees are enforced continuously again.

**Constraints (stated + system-implied).** The 200-line budget is locked by
issue #78 and out of scope to raise (spec 0077 *Out of scope*); the two tests'
assertions are frozen — the READMEs are made to pass, never the tests loosened;
every content guarantee the tests grep for must survive the trim; the ordered
GitHub-Actions↔GitLab command-parity gate (spec 0076/0049) must stay green; and
`.gitlab-ci.yml` is generated output (spec 0048), regenerated via
`scripts/build-ci.sh`, never hand-edited.

**Non-goals.** Revising the budget value; editing any assertion in the two
tests; the sibling quarantined tests `test-e2e-runner.sh` (#543) and
`test-e2e-toml-merge.sh` (#544); restructuring the e2e harness or `run.sh`.

### Approach

This is a pure conciseness-and-wiring pass with no behavioral surface: condense
both walkthrough READMEs so each lands at ≤200 lines while every fact the two
content-guarantee tests assert stays literally present, then flip both tests
[...]

Issue #534, opened 2026-07-17
🧪 Audit CI wiring of scripts/tests/*.sh — 28 of 47 tests run in no workflow


gfourny-sfeir:
## Problem

While fixing #530 we found that `scripts/tests/test-mempalace-transcript-hook.sh` was wired into no CI workflow, which is why a false-negative regression (broken since PR #211) went undetected for months.

This is not isolated: **28 of the 47** `scripts/tests/*.sh` scripts are referenced in no GitHub Actions workflow. `build.yml` enumerates test steps individually, so any new `test-*.sh` that isn't manually added silently never runs in CI.

## Unwired tests (as of this writing)

Mostly the `test-e2e-*` family (presumably meant to run via the e2e harness, `tests/e2e/run.sh` — to be confirmed), plus several standalone ones:

- `test-build-ci.sh`
- `test-chroma-fd-limits.sh`, `test-chroma-health-race.sh`, `test-chroma-server.sh`
- `test-e2e-*` (assert-lib, auth-common, auth-ready, auth-scripts, auth-taskfile, defaults-toml, docker-images, gitignore, judge-config, libs-readme, llm-judge-lib, readme, report, runner, runner-delegation, runner-taskfile, scenarios-parity, scenarios-structure, structural-lib, taskfile, toml-merge, versions-lock)
- `test-spec-linter.sh`
- `test-system-context-store.sh`
- `test-mempalace-transcript-hook.sh` *(now wired in #533)*

## Decision to make

Choose a durable strategy so this class of gap self-heals:

1. **Glob runner** — a CI step that discovers and runs every `scripts/tests/*.sh` (with an explicit allowlist/denylist for the ones that legitimately belong to another harness, e.g. e2e run via `tests/e2e/run.sh`). New tests are covered by default.
2. **Registration guard** — keep explicit enumeration but add a `check-*.sh` (+ its `test-*.sh`) that fails CI when a `scripts/tests/*.sh` exists that is neither wired into a workflow nor explicitly exempted. Mirrors the repo convention "every `check-*.sh` has a `test-*.sh`".

First step: confirm which of the 28 are already exercised by the e2e runner vs genuinely orphaned, then pick (1) or (2).

Discovered during #530 / #533.

hcross:
## Logbook — SPECS stage (audit + strategy)

Picked up under the SPECS → PLAN → DEV → REVIEW lifecycle (ADR-0010), interaction mode **INTERMEDIATE**.

[...]

PR #554, opened 2026-07-18
🐛 Stop toml_merge injecting phantom env_keys into keyless cli tables

### LINKED ISSUE #544: 🧪 test-e2e-toml-merge.sh: empty-local-over-defaults emits unexpected env_keys: []
Surfaced by #534 while wiring previously-unrun tests into CI.

## Problem
`scripts/tests/test-e2e-toml-merge.sh` fails one assertion: merging an empty local config over defaults emits an extra `env_keys: []` that the "equals defaults" fixture does not expect — a genuine merge-logic / fixture divergence.

## Current state
Quarantined via `ci/test-wiring-exemptions.txt` (spec 0076 guard) so #534 could ship green. This issue tracks the real fix.

## Fix
Determine whether the `env_keys: []` emission is correct (update the fixture) or a merge bug (fix the merge). Then de-quarantine and wire into `check-components`.

hcross:
Guards the `env_keys` dedup in `tests/e2e/lib/toml_merge.sh` so a `cli` table that declares no `env_keys` keeps none after merge — restoring the ADR-0003 "empty local → equals defaults" pass-through that a phantom `env_keys: []` injection was breaking. De-quarantines and wires `test-e2e-toml-merge.sh`. Implements spec 0079; `Closes #544`.

## How to read this PR?

One jq predicate in `tests/e2e/lib/toml_merge.sh`: `.value.env_keys = (.value.env_keys // [] | unique)` → `.value |= (if has("env_keys") then .env_keys |= unique else . end)`. The `// []` guard used to inject `env_keys: []` into every cli table; the guarded form dedups only where `env_keys` is present. `ci/test-wiring-exemptions.txt` loses its last quarantine entry; the test is wired into `check-components`; `.gitlab-ci.yml` regenerated.

## How to test this PR?

```bash
bash scripts/tests/test-e2e-toml-merge.sh   # 11 passed (was 10/1: Case 1 failed)
bash scripts/check-test-wiring.sh           # 46 wired + 2 exempted; honest
bash scripts/build-ci.sh --check            # clean
bash scripts/check-ci-parity.sh             # agree
```

- **`tests/e2e/lib/toml_merge.sh`**: the Phase-2 env_keys dedup is guarded by `has("env_keys")` so a keyless cli table stays keyless (spec 0079 R1); dedup where present is unchanged (R2); the `command` full-replacement branch is untouched. yq `*+` Phase-1 never fabricates the key, so the predicate is sound. Consumers already read `.env_keys // []` (`run.sh:318`, `test-e2e-defaults-toml.sh:107`), so removing the injection is invisible downstream.
- **De-quarantine + wiring**: removed the `#544` exemption (and the now-empty Quarantined section header); wired `test-e2e-toml-merge.sh` into `check-components` (build.yml + ci-capabilities.yml, identical order after `test-e2e-runner.sh`); `.gitlab-ci.yml` regenerated. Guard end state: **46 wired + 2 exempted = 48** — all three #534 follow-up bugs (#542/#543/#544) now fixed and wired; only the two environment-dependent exemptions remain.
- **Decision:** fixed the MERGE, not the test fixture — the injection was redundant (consumers tolerate absent env_keys) and violated the ADR-0003 pass-through contract the test locks.
- **Governance:** no version bump; no `build-components.sh`; not a CLI-matrix trigger; no core-paths change.

Last of the three follow-ups from the #534 test-wiring guard.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -205,6 +205,9 @@ jobs:
       - name: Test e2e runner regression (spec 0078)
         run: bash scripts/tests/test-e2e-runner.sh
 
+      - name: Test e2e toml-merge regression (spec 0079)
+        run: bash scripts/tests/test-e2e-toml-merge.sh
+
   lint-markdown:
     runs-on: ubuntu-latest
     steps:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -90,6 +90,7 @@ check-components:
     - "bash scripts/tests/test-e2e-readme.sh"
     - "bash scripts/tests/test-e2e-libs-readme.sh"
     - "bash scripts/tests/test-e2e-runner.sh"
+    - "bash scripts/tests/test-e2e-toml-merge.sh"
   rules:
     - if: '$CI_PIPELINE_SOURCE == "merge_request_event" && $CI_MERGE_REQUEST_TARGET_BRANCH_NAME =~ /^main$|^release\//'
     - if: '$CI_PIPELINE_SOURCE == "push" && $CI_COMMIT_BRANCH =~ /^main$|^release\//'
--- a/ci/ci-capabilities.yml
+++ b/ci/ci-capabilities.yml
@@ -99,6 +99,7 @@ capabilities:
       - bash scripts/tests/test-e2e-readme.sh
       - bash scripts/tests/test-e2e-libs-readme.sh
       - bash scripts/tests/test-e2e-runner.sh
+      - bash scripts/tests/test-e2e-toml-merge.sh
 
   - id: lint-markdown
     name: "Lint Markdown files"
--- a/ci/test-wiring-exemptions.txt
+++ b/ci/test-wiring-exemptions.txt
@@ -19,9 +19,3 @@
 # --- Environment-dependent (cannot run in the hermetic check-components job) ---
 test-spec-linter.sh	requires the repo's npm-installed node_modules (markdownlint-cli via npx, plus js-yaml/semver) which the hermetic check-components job does not provision; spec-linter.js is exercised on real specs by the lint-specs job via `task spec:lint`.
 test-e2e-auth-scripts.sh	exercises scripts/e2e/auth-copilot.sh, which e2e_die's when ~/.copilot/instructions/ is absent — a `task setup:copilot` side-effect present on a developer host but not in a clean CI HOME; cannot pass in the hermetic check-components job. See follow-up note from #534 on whether the guidance path should print before that precondition.
-
-# --- Quarantined: a genuine pre-existing failure, out of scope for spec 0076 ---
-# Each entry below names a follow-up issue tracked from #534. The fix belongs to
-# a separate ticket (spec 0076 wires tests and guards wiring; it does not own the
-# code the tests exercise). Remove the entry once the follow-up is fixed + wired.
-test-e2e-toml-merge.sh	quarantined: merging an empty local over defaults emits an extra `env_keys: []` the fixture does not expect (genuine merge/fixture drift); out of scope for spec 0076 — tracked in #544 (follow-up from #534).
--- a/tests/e2e/lib/toml_merge.sh
+++ b/tests/e2e/lib/toml_merge.sh
@@ -62,7 +62,7 @@ if [[ -n "$LOCAL" && -f "$LOCAL" ]]; then
       if ($local[0].cli[$c].command? // null) != null then
         .value.command = $local[0].cli[$c].command
       else . end |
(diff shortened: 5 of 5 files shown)


hcross:
**Verdict: APPROVE**

_Posted as a PR comment, not a formal `APPROVE` review event: the reviewer shares the author's GitHub identity, so the review API rejects a self-approval. This comment is the canonical verdict._

Independent cold review of spec 0079 (`Closes #544`). Diff is 5 files, +6/-8; the load-bearing change is one jq predicate in `tests/e2e/lib/toml_merge.sh`. Every requirement verified empirically against the head commit (`e7da38e`).

## CI status — all green

| Check | Result |
|---|---|
| `check-components` | **pass** (2m13s) — runs the newly-wired `test-e2e-toml-merge.sh` |
| `build`, `check-ci-parity`, `gitlab-ci-check`, `lint-*`, `check-skill-versions`, `check-extension-version-bump`, `check-feedback-routing`, `test-harness-curate`, `check-agents-size` | pass |

No failing or pending required checks.

## Correctness — R1/R2 reproduced through the merger

The change: `.value.env_keys = (.value.env_keys // [] | unique)` → `.value |= (if has("env_keys") then .env_keys |= unique else . end)` (`tests/e2e/lib/toml_merge.sh:65`). The old `// []` fabricated `env_keys: []` on every cli table; the guarded form touches the key only when present.

Verified by driving the merger with hand-made fixtures:

- **R1** — keyless `cli.claude` + empty local → output carries **no** `env_keys` key; equals defaults verbatim. Same for the no-local-arg path (which never enters jq).
- **R2 (dedup)** — defaults `["A","B"]` + local `["B","C"]` → single `["A","B","C"]`. Defaults-only `env_keys` + empty local → `["A","B"]` (dedup applied, no injection).
- The `command` full-replacement branch (`:62-64`) is untouched.

## Convention compliance — clean

- Gitmoji title (`🐛`), four mandatory PR body sections present, `Closes #544` in body.
- No skill/agent source touched → no version bump owed (`check-skill-versions` green). No `artifacts/**` → no `build-components` owed. No CLI-matrix trigger path (`build.yml` is not `claude.yml`/`gemini.yml`; no `scripts/{build,…}-*.sh` touched). No `docs/layers.md` core-path change → no `core-paths.txt` owed. The PR body's governance claims all hold.

## R3 / R4 — de-quarantine + parity

- `ci/test-wiring-exemptions.txt` now holds exactly the two environment-dependent entries (`test-spec-linter.sh`, `test-e2e-auth-scripts.sh`); the `#544` quarantine block is gone.
- `check-test-wiring.sh` → `all test scripts are wired (46) or exempted with a reason (2); the exemption allowlist is honest.`
- `test-e2e-toml-merge.sh` appended last (after `test-e2e-runner.sh`) identically in `.github/workflows/build.yml`, `ci/ci-capabilities.yml`, and `.gitlab-ci.yml`. `build-ci.sh --check` → matches reference; `check-ci-parity.sh` → all three engines agree.
- `test-e2e-toml-merge.sh` → **11 passed / 0 failed** locally; Case 1 (empty local equals defaults) now green, Cases 2–8 unaffected.

## Non-blocking observation (not a blocker — APPROVE stands)

**`class: tech` — R2 "preserves the original declaration order" is not actually honored; `jq unique` sorts.**

Spec 0079 R2 requires the deduplicated `env_keys` to *preserve declaration order*. `jq`'s `unique` sorts: I reproduced `env_keys = ["Z","A","M"]` (empty local) → `["A","M","Z"]`. The inline comment at `tests/e2e/lib/toml_merge.sh:58` ("unique() preserves order") states this false premise, and R2 inherited it.

Why this does **not** block:
- **Pre-existing, not a regression.** `main` already used `… | unique` (`git show main:tests/e2e/lib/toml_merge.sh`); this PR changed only the injection guard, not the dedup mechanism. The order behavior is byte-for-byte the same before and after.
- The spec explicitly scopes this PR to "the `env_keys` injection fix"; the merged test passes and CI is green.
- `env_keys` are environment-variable names; order has no established downstream effect, and the spec's own example is already alphabetically sorted.
- No test exercises multi-key ordering (the `env_keys` case at `:83`/`:93` is single-element), so the gap is latent.

**Recommended follow-up** (separate ticket, not this PR): either swap `unique` for an order-preserving idiom (e.g. `reduce .[] as $x ([]; if index([$x]) then . else . + [$x] end)`), fix the misleading `:58` comment, and add a multi-key ordering assertion — or, if order is deemed immaterial, relax R2 via a spec delta. Escalation direction is a genuine `tech`/`spec` tie; I lean `tech` since the code comment asserts a property the mechanism cannot deliver.

Focused, correct, fully wired, CI green. Ship it; track the ordering gap separately.

PR #593, opened 2026-07-21
🐛 Raise open-file limit before launching the ad-hoc Chroma daemon

### LINKED ISSUE #587: 🐛 start-chroma-server.sh launches the shared Chroma daemon without raising ulimit -n (fd-exhaustion recurrence)
## Problem

`scripts/start-chroma-server.sh` launches the **shared** ChromaDB HTTP daemon
(used by *all* MemPalace MCP server instances) without raising `ulimit -n`:

```sh
# scripts/start-chroma-server.sh:55-59
nohup "${PYTHON_BIN}" "${CHROMA_BIN}" run \
  --path "${PALACE_DIR}" \
  --host "${HOST}" \
  --port "${PORT}" \
  >> "${LOG_FILE}" 2>&1 &
```

The daemon therefore inherits whatever file-descriptor soft limit its launch
context had. On macOS the launchd/GUI default is **256** (`launchctl limit
maxfiles` → `256 unlimited`), while an interactive shell may be far higher. If
the daemon was started outside an fd-raised shell, its ceiling is 256.

Because a **single** daemon is shared by every session, and each active session
opens a pool of HTTP connections to it, fd usage on the daemon scales with the
number of concurrent **active** sessions.

## Evidence (live, macOS, 2026-07-21)

- 7 interactive Claude Code sessions open concurrently, each with one
  `mempalace-http-wrapper.py` MCP server (correctly parented — **not** leaked).
- Chroma daemon (`chroma run`): **160 open fds**, **80 ESTABLISHED** connections
  on `:8001`.
- Idle sessions hold ~2 connections; active sessions spike to **~28** each (two
  long-running sessions accounted for 56 of the 80 connections).
- Headroom to a 256 soft limit ≈ 96 fds ≈ **~4 more active sessions** before
  `EMFILE`.

This is exactly the recurrence path for the previously-fixed friction: under
`EMFILE`, palace discovery squashes the error to `{"error":"No palace found"}`
on `mempalace_add_drawer` / `mempalace_list_drawers` while read tools (which
reuse already-open fds) keep working. That symptom was mitigated server-side by
#303 / spec 0029 (raising the daemon fd limits + reaping orphaned wrappers), but
`start-chroma-server.sh` — the canonical launcher — still does not set a limit,
so a daemon started via this script has no guaranteed headroom.

## Proposed fix

Raise `ulimit -n` in `start-chroma-server.sh` immediately before the `nohup`
launch (subshell-scoped so it affects only the daemon), with an overridable
target:

```sh
# before line 55
ulimit -n "${MEMPALACE_CHROMA_NOFILE:-10240}" 2>/dev/null || \
  echo "warning: could not raise ulimit -n; daemon fd ceiling unchanged" >&2
```

10240 gives ~two orders of magnitude of headroom over the observed 160. Pair it
with a one-line note in `docs/runbooks/chroma-http-server.md`.

## Scope / process

Change to spec-governed sync/infra tooling → enters via the SPECS stage per
ADR-0010. Related: #303 / spec 0029 (daemon fd limits), ADR-0006 (shared
HTTP-daemon architecture). See also the companion issue on the per-session
connection-pool fanout (client-side complement to this server-side fix).

hcross:
Raises the ad-hoc Chroma daemon's open-file soft limit to a configurable floor before launch, closing the fd-exhaustion gap the supervised launch path already had covered by spec 0029. Implements `specs/0087-chroma-launcher-ulimit-floor.md`.

## How to read this PR?

Read in this order:

1. `scripts/start-chroma-server.sh` — the actual fix. A guarded `ulimit -n "${MEMPALACE_CHROMA_ULIMIT_FLOOR:-10240}"` call is inserted immediately before the `# ── Launch daemon ───` block, right before the `nohup` invocation. It is wrapped in `if ! ulimit -n ...; then echo WARNING ...; fi` specifically because the script has `set -e` (line 7) — a bare failing `ulimit` would otherwise abort the whole script the moment a host's fixed-below-floor hard ceiling is hit (e.g. an unprivileged container), which is exactly the failure mode R3 rules out.
2. `docs/runbooks/chroma-http-server.md` — new subsection placed right after the existing documentation of the supervised path's `65536` floor. It explicitly calls out that the two floors (`10240` here vs `65536` for the supervisor units) are independent and unrelated, to head off a reader conflating them.
3. `scripts/tests/test-chroma-server.sh` — three new static assertions (Tests 9-11) in Section A. They anchor on the literal guard string `if ! ulimit -n` rather than a bare `ulimit -n` substring match, because the script's own explanatory comment above the guard also contains that substring — a bare grep would key off prose instead of code. See the new comment block directly above Test 9 in the diff for the full anchoring rationale.

## How to test this PR?

Prerequisites: a POSIX shell with `ulimit` support (macOS/Linux).

1. Run the static test suite: `bash scripts/tests/test-chroma-server.sh`. Tests 9-11 should report pass — they verify (9) the guarded `ulimit -n` call precedes the `nohup` daemon launch, (10) that call's line carries both the literal `10240` default and the literal `MEMPALACE_CHROMA_ULIMIT_FLOOR` override-variable name, and (11) the call is structurally guarded (`if ! ulimit -n`) rather than a bare statement.
2. Golden path: `bash scripts/start-chroma-server.sh`, then inspect the daemon process's open-file soft limit (e.g. `cat /proc/<pid>/limits` on Linux, or the equivalent `ulimit -n` reported from within the daemon's own environment) — it should report at least `10240`.
3. Override path: `MEMPALACE_CHROMA_ULIMIT_FLOOR=20000 bash scripts/start-chroma-server.sh` — the daemon should come up with the higher floor instead.
4. Failure path: on a host whose file-descriptor hard ceiling is fixed below `10240` (e.g. an unprivileged container with a low `RLIMIT_NOFILE` hard limit), run `bash scripts/start-chroma-server.sh` — expect a `WARNING: could not raise open-file limit ...` line on stderr naming the ceiling still in effect, and the daemon should still start successfully rather than the script aborting.

- `scripts/start-chroma-server.sh` (+11 lines): inserted a new guarded block between the existing binary-executability check (`if [ ! -x "${CHROMA_BIN}" ]; then ... fi`) and the `# ── Launch daemon ───` comment. The block raises `ulimit -n` to `${MEMPALACE_CHROMA_ULIMIT_FLOOR:-10240}`, guarded against `set -e` so a failed raise falls through to a stderr warning (naming `$(ulimit -Hn)` as the ceiling still in effect) instead of aborting the script. Because there is no intervening subshell between this `ulimit -n` call and the `nohup ... &` daemon launch further down, the raised limit is inherited by the daemon process only — the invoking caller's shell is untouched (R4).
- `docs/runbooks/chroma-http-server.md` (+16 lines): new `###` subsection "File-descriptor floor on the manual launch path", inserted after the existing subsection documenting the supervised path's `65536` floor and before the "Migrating from the legacy `PersistentClient` setup" section. States the `10240` default, the `MEMPALACE_CHROMA_ULIMIT_FLOOR` override, the warn-and-continue failure behavior, and an explicit note that this floor is independent of the supervised path's floor.
- `scripts/tests/test-chroma-server.sh` (+68 lines): updated the file's header docstring to note the new contract on `scripts/start-chroma-server.sh`, and added Tests 9-11 to Section A ("Static checks, no daemon required"):
  - Test 9: the guarded `ulimit -n` call's line number is lower than the `nohup` call's line number (locks R6 — the raise happens before daemon launch).
  - Test 10: that same line contains both the literal string `10240` and the literal string `MEMPALACE_CHROMA_ULIMIT_FLOOR` (locks R1/R2 — default value and override variable name).
  - Test 11: the call is structurally guarded (`if ! ulimit -n` appears at least once) rather than a bare statement (locks R3 — warn-and-continue instead of abort under `set -e`).

Normative spec: `specs/0087-chroma-launcher-ulimit-floor.md` (already merged to `main`). Plan reviewed cold on issue #587: https://github.com/crewrig/crewrig/issues/587#issuecomment-5033423534. Complexity tier: `small`.

Closes #587

### DIFF
--- a/docs/runbooks/chroma-http-server.md
+++ b/docs/runbooks/chroma-http-server.md
@@ -83,6 +83,22 @@ open-file limit:
 
 Both should report `65536` (or higher) once the reload completes.
 
+### File-descriptor floor on the manual launch path
+
+`scripts/start-chroma-server.sh` — the ad-hoc path used to bring up the
+daemon outside the installed supervisor — raises the daemon process's
+open-file soft limit to `10240` (`MEMPALACE_CHROMA_ULIMIT_FLOOR`
+overrides the default) immediately before launching it. This floor is
+**deliberately different from, and independent of, the supervised path's
+`65536` floor documented above** — do not confuse the two: raising or
+lowering one has no effect on the other, and a manually-started daemon
+never inherits the supervisor units' resource limits.
+
+If the raise fails (e.g. the host's file-descriptor hard ceiling is fixed
+below the requested floor), the script prints a warning to standard error
+naming the ceiling that remains in effect and continues launching the
+daemon rather than aborting.
+
 ## Migrating from the legacy `PersistentClient` setup
 
 If you upgraded a working CrewRig install across the #98 boundary:
--- a/scripts/start-chroma-server.sh
+++ b/scripts/start-chroma-server.sh
@@ -58,6 +58,17 @@ if [ ! -x "${CHROMA_BIN}" ]; then
   exit 1
 fi
 
+# ── Raise open-file limit (spec 0087) ───────────────────────────────────────
+# Scoped to this script's process: `ulimit -n` here only affects the shell
+# running this script, and the daemon inherits it as a child of that shell
+# (via `nohup ... &` below, no intervening subshell) — the invoking caller's
+# shell and any other process are left untouched. Guarded because `set -e`
+# (line 7) would otherwise abort the whole script on a fixed-below-floor
+# hard ceiling (e.g. an unprivileged container default).
+if ! ulimit -n "${MEMPALACE_CHROMA_ULIMIT_FLOOR:-10240}" 2>/dev/null; then
+  echo "WARNING: could not raise open-file limit to ${MEMPALACE_CHROMA_ULIMIT_FLOOR:-10240}; current hard ceiling is $(ulimit -Hn)." >&2
+fi
+
 # ── Launch daemon ───────────────────────────────────────────────────────────
 nohup "${PYTHON_BIN}" "${CHROMA_BIN}" run \
   --path "${PALACE_DIR}" \
--- a/scripts/tests/test-chroma-health-race.sh
+++ b/scripts/tests/test-chroma-health-race.sh
@@ -10,9 +10,10 @@
 # then the daemon is up.
 #
 # This test locks in the contract that the health check MUST poll the
-# status script (mirroring the 15s / 0.3s pattern of
-# `scripts/start-chroma-server.sh` lines 65-80) instead of relying on
-# a single invocation.
+# status script (mirroring the 15s / 0.3s pattern of the health-check
+# loop in `scripts/start-chroma-server.sh` — the
+# `# ── Health-check loop` block) instead of relying on a single
+# invocation.
 #
 # Strategy
 # --------
@@ -107,8 +108,8 @@ fi
 # ─────────────────────────────────────────────────────────────────────────────
 # Test 1 — Behavioral: the health-check block must survive a 2-second
 # slow start. Against the one-shot code on main this fails; once the
-# developer adds a retry loop (mirroring start-chroma-server.sh lines
-# 65-80), it passes.
+# developer adds a retry loop (mirroring the `# ── Health-check loop`
+# block in start-chroma-server.sh), it passes.
 # ─────────────────────────────────────────────────────────────────────────────
 HARNESS="$(mktemp -t chroma-health-race-harness.XXXXXX.sh)"
 trap 'rm -rf "$SANDBOX" "$HARNESS"' EXIT INT TERM
@@ -149,7 +150,8 @@ fi
 #   - a `while`/`until` loop in the health-check block
 #   - a `for` loop in the health-check block
 #   - a recognizable `deadline=` / `SECONDS` budget marker
-#     (matches the convention used by scripts/start-chroma-server.sh)
+#     (matches the convention used by the `# ── Health-check loop`
+#     block in scripts/start-chroma-server.sh)
 # ─────────────────────────────────────────────────────────────────────────────
 if echo "$BLOCK" | grep -Eq '(\bwhile\b|\buntil\b|\bfor\b|deadline=|SECONDS)'; then
   note_pass "health-check block contains a polling construct"
--- a/scripts/tests/test-chroma-server.sh
+++ b/scripts/tests/test-chroma-server.sh
@@ -4,6 +4,7 @@
 #
 # Locks the contracts of:
 #   - scripts/start-chroma-server.sh   (idempotent start, PID-file lifecycle)
+#       * raises the open-file limit to a floor before launching (spec 0087)
 #   - scripts/stop-chroma-server.sh    (graceful stop, cleans PID file)
 #   - scripts/status-chroma-server.sh  (exit 0 healthy / exit 1 otherwise)
 #   - scripts/lib/mempalace-http-wrapper.py
@@ -85,6 +86,73 @@ if [[ -f "$WRAPPER_PY" ]]; then
   fi
 else
   note_fail "wrapper carries chromadb version pin" "missing $WRAPPER_PY"
+fi
+
+# Tests 9-11 anchor on `if ! ulimit -n` rather than the plain substring
+# `ulimit -n`: the guard's own explanatory comment above it also contains
+# the literal text `ulimit -n` (and a separate comment mentions `nohup`),
+# so a bare substring grep would key off prose instead of code. `if ! ulimit
+# -n` and `^nohup` are each unique in the file today and identify the real
+# guarded call and the real daemon launch, not the comments describing them.
+
+# Test 9 — start script must raise the open-file limit (spec 0087 R1) BEFORE
+# launching the daemon (spec 0087 R6): the guarded `ulimit -n` call must sit
+# at a lower line number than the `nohup` daemon launch.
+if [[ -f "$START_SH" ]]; then
+  ulimit_line="$(grep -n "if ! ulimit -n" "$START_SH" | head -1 | cut -d: -f1)"
+  nohup_line="$(grep -n "^nohup" "$START_SH" | head -1 | cut -d: -f1)"
+  if [[ -z "$ulimit_line" ]]; then
+    note_fail "start script raises open-file limit before nohup" \
+      "no guarded 'ulimit -n' call found"
+  elif [[ -z "$nohup_line" ]]; then
+    note_fail "start script raises open-file limit before nohup" \
+      "no 'nohup' daemon launch found"
+  elif (( ulimit_line < nohup_line )); then
+    note_pass "start script raises open-file limit before nohup (ulimit L${ulimit_line} < nohup L${nohup_line})"
+  else
+    note_fail "start script raises open-file limit before nohup" \
+      "ulimit (L${ulimit_line}) is NOT before nohup (L${nohup_line})"
+  fi
+else
+  note_fail "start script raises open-file limit before nohup" "missing $START_SH"
+fi
+
+# Test 10 — the ulimit call carries both the literal default floor (10240,
+# spec 0087 R1) and the literal override variable name
+# (MEMPALACE_CHROMA_ULIMIT_FLOOR, spec 0087 R2) on that same line, so a
+# refactor cannot silently change the default or rename the override
+# variable without also breaking this assertion.
+if [[ -f "$START_SH" ]]; then
+  ulimit_line_text="$(grep -n "if ! ulimit -n" "$START_SH" | head -1)"
+  if [[ -z "$ulimit_line_text" ]]; then
+    note_fail "ulimit call carries default floor and override variable" \
+      "no guarded 'ulimit -n' call found"
+  elif [[ "$ulimit_line_text" == *"10240"* && "$ulimit_line_text" == *"MEMPALACE_CHROMA_ULIMIT_FLOOR"* ]]; then
+    note_pass "ulimit call carries default floor and override variable"
+  else
+    note_fail "ulimit call carries default floor and override variable" \
+      "line lacks '10240' and/or 'MEMPALACE_CHROMA_ULIMIT_FLOOR': $ulimit_line_text"
+  fi
+else
+  note_fail "ulimit call carries default floor and override variable" "missing $START_SH"
+fi
+
+# Test 11 — the ulimit -n call is guarded (`if ! ulimit -n ...; then`), not a
+# bare statement. Under `set -e` (line 7 of the start script) a bare failing
+# ulimit would abort the whole script the moment a host's fixed-below-floor
+# hard ceiling is hit; the guard is what lets the script warn and continue
+# instead (spec 0087 R3). Locks the guard structurally so a future
+# refactor cannot silently drop it.
+if [[ -f "$START_SH" ]]; then
+  guard_count="$(grep -c "if ! ulimit -n" "$START_SH")"
+  if [[ "$guard_count" -ge 1 ]]; then
+    note_pass "ulimit -n call is guarded against set -e"
+  else
+    note_fail "ulimit -n call is guarded against set -e" \
+      "no 'if ! ulimit -n' guard structure found"
+  fi
+else
+  note_fail "ulimit -n call is guarded against set -e" "missing $START_SH"
 fi
 
 # ─────────────────────────────────────────────────────────────────────────────


hcross:
## Verdict: APPROVE

Cold, independent review — no authoring-session context. Verified directly against `crewrig/crewrig` via `gh`, the PR diff/head ref, `specs/0087-chroma-launcher-ulimit-floor.md` on `main`, and a local checkout of the exact head commit (`86726a0`), confirmed byte-identical to `contents?ref=fix/0087-chroma-launcher-ulimit-floor` for `scripts/start-chroma-server.sh`.

*(Note: `gh pr review --approve` is blocked here — same GitHub identity as the PR author — so this verdict is posted as a plain comment per the documented workaround.)*

### CI status

All 12 checks pass (re-confirmed live via `gh pr checks 593`): `build`, `check-agents-size`, `check-ci-parity`, `check-components`, `check-extension-version-bump`, `check-feedback-routing`, `check-skill-versions`, `gitlab-ci-check`, `grep-anti-patterns`, `lint-markdown`, `lint-specs`, `test-harness-curate`. Notably `build` (job step at `.github/workflows/build.yml:149`) is what actually executes `scripts/tests/test-chroma-server.sh` in CI — so the new Tests 9-11 ran and passed there, not just locally.

### Correctness — against `specs/0087-chroma-launcher-ulimit-floor.md` (R1-R6)

Read the merged spec directly (`gh api .../contents/specs/0087-...md?ref=main`) and checked every SHALL against the diff:

- **R1** (raise to ≥10240 immediately before launch) — `scripts/start-chroma-server.sh:68-70`: `if ! ulimit -n "${MEMPALACE_CHROMA_ULIMIT_FLOOR:-10240}" ...` sits directly between the binary-executability check and the `nohup` daemon launch (line 73). Satisfied.
- **R2** (overridable via env var) — same line, `${MEMPALACE_CHROMA_ULIMIT_FLOOR:-10240}`. Satisfied.
- **R3** (warn-and-continue on failure) — guarded in an `if ! ...; then echo "WARNING: ... current hard ceiling is $(ulimit -Hn)." >&2; fi`, no `exit`/`return` in the branch, so the script falls through to the daemon launch unconditionally. Matches the spec's scenario wording ("naming the ceiling that remains in effect") almost verbatim. Satisfied.
- **R4** (scoped to the daemon, not the invoking shell/any other process) — `start-chroma-server.sh` is always invoked as its own process (`bash scripts/start-chroma-server.sh` / `./scripts/start-chroma-server.sh`), never sourced — confirmed by `grep -rn "start-chroma-server"` across the repo: every hit outside the three changed files is a prose/line-number reference, never a `source`/`.` call. A bare `ulimit -n` inside a non-sourced script therefore only affects that script's own process and the children it forks afterward (the `nohup`'d daemon, no intervening subshell) — the caller's shell is untouched by construction. Satisfied.
- **R5** (runbook doc) — `docs/runbooks/chroma-http-server.md` new "File-descriptor floor on the manual launch path" subsection, placed right after the existing 65536/supervised-floor subsection, explicitly disclaiming the two floors are independent. Satisfied.
- **R6** (regression test, raise-before-invoke) — `scripts/tests/test-chroma-server.sh` Test 9 compares `grep -n "if ! ulimit -n"` line vs `grep -n "^nohup"` line and asserts the former is lower. Satisfied.

### Test coverage — verified these are real regression locks, not vacuous

Ran `bash scripts/tests/test-chroma-server.sh` locally: **11 passed, 0 failed** (8 pre-existing + Tests 9-11). To rule out "tests that would pass even without the fix," I temporarily stripped the new `ulimit` block from a working copy of `scripts/start-chroma-server.sh` and re-ran the same suite: **Tests 9, 10, and 11 fail** (`no guarded 'ulimit -n' call found` / `no 'if ! ulimit -n' guard structure found`), 8 passed / 3 failed. Confirms the three new assertions genuinely lock R1/R2/R3/R6 rather than passing regardless. Working tree restored via `git checkout --` afterward; confirmed clean and byte-identical to the PR head again.

The anchoring choice (`if ! ulimit -n` instead of a bare `ulimit -n` substring) is deliberate and documented in the diff's own comment block — correctly avoids false-matching the explanatory prose comment directly above the guard, which also contains the substring `ulimit -n`.

### Convention compliance (`AGENTS.md`)

- **Language** — all repo content (script comments, docs, test comments, commit message, PR title/body) is English. Compliant.
- **Naming Convention (Gitmoji)** — commit `86726a0` and PR title both use `🐛 Raise open-file limit before launching the ad-hoc Chroma daemon`; no Conventional-Commits prefix. Compliant.
- **Version Bump Convention** — N/A. No path under the skill/agent-source manifest (`artifacts/**`, etc.) is touched; only `scripts/`, `scripts/tests/`, `docs/runbooks/` are.
- **CLI Matrix Maintenance** — N/A. `scripts/start-chroma-server.sh` does not match any of the `scripts/{build,install,setup,import,manage}-*.sh` trigger globs, and none of the touched paths match the other triggers listed in `AGENTS.md` → *CLI Matrix Maintenance*. No `docs/cli-matrix.md` update required, and none was made — correct.
- **Branch naming** — `fix/0087-chroma-launcher-ulimit-floor` matches `<prefix>/<NNNN>-<slug>` with `fix/` a valid implementation prefix per `docs/spec-pr-workflow.md`, and `0087`/`chroma-launcher-ulimit-floor` match the spec's own `id`/`slug` frontmatter. Compliant.
- **PR body format** — two-sentence intro, `## How to read this PR?`, `## How to test this PR?`, `## Detailed description (for agents)` all present, plus `Closes #587`. Compliant.
- **Logbook (Rule A/B)** — issue #587 is the pre-existing tracked issue and correctly serves as its own logbook (no separate logbook issue opened). Incremental comments present: PLAN (11:33), a genuinely independent cold PLAN review (11:38, verifies concrete facts against `main` rather than taking the plan on trust), and a DEV-stage comment (11:49) posted at PR-open time recording the one lesson learned (the `if ! ulimit -n` anchoring choice). Compliant with Rule A/B.
- **Complexity tier** — declared `small` in both the spec frontmatter and the PR body; team composition (developer + tester + pr-logbook + pr-reviewer) meets-or-exceeds the `small` minimum. No under-statement.

### Linter findings

- `lint-shell.sh scripts/start-chroma-server.sh scripts/tests/test-chroma-server.sh` — 3 shellcheck info-level notes (SC1091 "not following" on two pre-existing `source common.sh` lines, SC2329 on a pre-existing unused `cleanup()` in Section B). All three are outside the lines this diff touches (confirmed against the diff hunks) — pre-existing, not introduced here. No action needed.
- `lint-markdown.sh docs/runbooks/chroma-http-server.md` — clean, exit 0.

### Non-blocking observation (`class: tech`)

`scripts/tests/test-chroma-health-race.sh` (untouched by this PR) carries two comments citing `start-chroma-server.sh lines 65-80` for the health-check/deadline retry loop (lines 14 and 110 of that file). This PR's +11-line insertion shifts that loop to roughly lines 76-91, so the comment is now stale by 11 lines. Confirmed this is prose only — the test extracts its target block from `common.sh` via a marker string, not by line number, so nothing programmatic breaks. Purely a documentation-drift nit; does not block this PR, worth a one-line fix whenever that file is next touched.

---

**Summary:** All 6 spec-0087 requirements verified against the diff with file/line citations; CI green (12/12, re-confirmed); regression tests independently verified to fail without the fix and pass with it; conventions (language, Gitmoji, version-bump N/A, CLI-matrix N/A, branch naming, PR format, logbook) all compliant. One non-blocking `class: tech` nit noted above (stale line-number comment in an adjacent, untouched test file) — does not block merge.

hcross:
## Verdict: APPROVE

Cold, independent re-review — REVIEW iteration 2 (`iter:2`), no authoring-session context, no reuse of any prior review's conclusions taken on trust. Full pass, not incremental-only: re-verified spec 0087 end to end against the current head (`baa29b8`), re-ran CI, and specifically audited the iteration-1 fix commit for correctness and side effects.

*(Same GitHub identity as the PR author — `gh pr review --approve` is blocked — verdict posted as a plain comment per the documented workaround.)*

### CI status

All 12 checks pass on `baa29b8` (`gh pr checks 593`): `build`, `check-agents-size`, `check-ci-parity`, `check-components`, `check-extension-version-bump`, `check-feedback-routing`, `check-skill-versions`, `gitlab-ci-check`, `grep-anti-patterns`, `lint-markdown`, `lint-specs`, `test-harness-curate`. Confirmed `.github/workflows/build.yml` actually executes both `scripts/tests/test-chroma-health-race.sh` (line 146) and `scripts/tests/test-chroma-server.sh` (line 149) — so the iteration-1 fix and the spec-0087 tests both ran in CI, not just locally.

### Spec 0087 — full re-check (R1-R6), not assumed from iteration 1

Read `specs/0087-chroma-launcher-ulimit-floor.md` fresh off `crewrig/main` and re-checked every SHALL against the diff on the PR branch worktree:

- **R1** (raise to ≥10240 immediately before launch) — `scripts/start-chroma-server.sh:68`: `if ! ulimit -n "${MEMPALACE_CHROMA_ULIMIT_FLOOR:-10240}" 2>/dev/null; then` sits between the binary-executability check and the `nohup` launch at line 73. Satisfied.
- **R2** (overridable via env var) — same line. Satisfied.
- **R3** (warn-and-continue on failure) — guarded, no `exit`/`return` in the failing branch; warning names `$(ulimit -Hn)` as the ceiling in effect. Satisfied.
- **R4** (scoped to the daemon, not the invoking shell) — the script is always exec'd as its own process, never sourced; `ulimit -n` inside it only affects that process and children forked after it (the `nohup`'d daemon, no intervening subshell). Satisfied by construction.
- **R5** (runbook doc) — `docs/runbooks/chroma-http-server.md`, new "File-descriptor floor on the manual launch path" subsection placed directly after the existing 65536/supervised-floor text, explicitly disclaiming the two floors are independent. Satisfied.
- **R6** (regression test, raise-before-invoke) — `scripts/tests/test-chroma-server.sh` Test 9 compares the `if ! ulimit -n` line number against the `^nohup` line number. Satisfied.

Ran the full suite on the PR branch worktree (already checked out at `baa29b8`): `scripts/tests/test-chroma-server.sh` → **11 passed, 0 failed, 0 skipped** (`ulimit L68 < nohup L73`, both literal-value and guard-structure assertions pass). `scripts/tests/test-chroma-health-race.sh` → **2 passed, 0 failed**.

### Iteration-1 finding — verified fixed, cleanly, no new finding introduced

The iteration-1 `class: tech` finding was: comments in `scripts/tests/test-chroma-health-race.sh` (a file outside this PR's original diff) cited `scripts/start-chroma-server.sh` line numbers (`65-80`) that went stale once this PR's 11-line `ulimit` insertion shifted the health-check loop further down the file.

Verified commit `baa29b8` (`♻️ Anchor health-check comments structurally, not by line number`) resolves this correctly:

- All three occurrences (file header docstring, the Test-1 comment, the Test-2 comment) now read "the `# ── Health-check loop` block in `scripts/start-chroma-server.sh`" instead of citing line numbers.
- Confirmed `# ── Health-check loop` is a real, exact substring in `scripts/start-chroma-server.sh` (line 64: `# ── Health-check loop (15s max) ───...`) — the anchor is accurate, not aspirational.
- Confirmed via `grep -n "65-80\|lines 65"` against the PR-branch worktree file: **zero remaining matches** — no stale line-number reference survives anywhere in the file.
- Confirmed the fix is comment-only and precisely scoped: `gh pr diff` shows exactly 3 hunks touching this file across the two commits, all within comment blocks; no executable line changed. The pre-existing, out-of-scope reference to `scripts/lib/common.sh, around lines 170-180` (a different file, untouched by this PR) is correctly left alone — that drift, if any, belongs to whichever change last touched `common.sh`, not to this PR.
- Re-ran `scripts/tests/test-chroma-health-race.sh` standalone: **2 passed, 0 failed** — the comment rewrite did not disturb the test logic (it only extracts its target block from `common.sh` via a marker string, never by line number, so it was never at risk either way).

No new finding surfaces from this fix commit.

### Test-negative-control sanity check

To confirm Tests 9-11 are real locks and not vacuously passing, temporarily stripped the new `ulimit` block from a scratch copy of `scripts/start-chroma-server.sh` and re-ran `test-chroma-server.sh`: Tests 9, 10, 11 fail as expected (`no guarded 'ulimit -n' call found`), 8 passed / 3 failed. Working tree restored, confirmed clean and unchanged afterward.

### Convention compliance (`AGENTS.md`)

- **Language** — all touched content (script comments, docs, commit messages) is English. Compliant.
- **Naming Convention (Gitmoji)** — both commits (`🐛 ...`, `♻️ ...`) and the PR title use Gitmoji, no Conventional-Commits prefix. Compliant.
- **Version Bump Convention** — N/A, no `artifacts/**`/skill-agent-source path touched.
- **CLI Matrix Maintenance** — N/A, `scripts/start-chroma-server.sh` does not match the `scripts/{build,install,setup,import,manage}-*.sh` trigger glob, and no other trigger path is touched.
- **Branch naming** — `fix/0087-chroma-launcher-ulimit-floor` matches spec 0087's `id`/`slug`. Compliant.
- **PR body format** — two-sentence intro + all three mandatory sections + `Closes #587` present. Compliant.
- **Logbook (Rule A/B)** — issue #587 correctly serves as its own logbook; incremental comments present at PLAN, cold PLAN review, and DEV stages. Compliant.
- **Merge state** — `MERGEABLE` / `CLEAN`, no conflicts with `main`.

### Findings

None. Zero findings of any class (`tech` / `arch` / `spec`).

---

**Summary:** Spec 0087 (R1-R6) fully satisfied and independently re-verified against the current head; CI green 12/12 with the relevant test scripts confirmed to actually execute in the `build` job; the iteration-1 `class: tech` finding (stale line-number comments in `scripts/tests/test-chroma-health-race.sh`) is fixed cleanly by `baa29b8` — all three occurrences now anchor on the `# ── Health-check loop` block name, verified accurate, with zero remaining stale references and the test suite still green (2/2). No new finding introduced. This pass surfaces zero findings — per `AGENTS.md` → *Retroactive review loop*, the REVIEW loop can terminate on this PR pending CI reconfirmation on the reviewed commit.

PR #613, opened 2026-07-21
🐛 Fix macOS compatibility in sync-from-upstream.sh by replacing Bash 4+ features

### LINKED ISSUE #612: 🐛 sync-from-upstream.sh fails on macOS due to Bash 3.2 incompatibilities (mapfile & declare -A)
### Describe the Bug

The script `scripts/sync-from-upstream.sh` fails on macOS/Darwin because macOS ships with Bash 3.2 by default. The script relies on two features introduced in Bash 4.0:
1. `mapfile -d ''` at lines 497 and 544.
2. `declare -A` (associative arrays) inside `reconcile_dir()` at line 305.

When running on macOS (or any environment with Bash < 4.0), the script aborts with the following errors:
```
scripts/sync-from-upstream.sh: line 497: mapfile: command not found
scripts/sync-from-upstream.sh: line 305: declare: -A: invalid option
```

### To Reproduce

On any macOS machine running the default system Bash (`/bin/bash` or `bash` v3.2.57):
1. Run `bash scripts/tests/test-sync-from-upstream.sh`.
2. Observe 20+ failures due to `mapfile` and `declare -A` errors.

### Expected Behavior

The script should remain fully compatible with older Bash versions, especially Bash 3.2 which is the default on macOS (Darwin).

### Proposed Fix

We can easily rewrite those parts using fully compatible Bash 3.2 primitives:
- Replace `mapfile -d '' spec < <(pathspec_for "$path")` with a standard `while read -d ''` loop:
  ```bash
  spec=()
  while IFS= read -r -d '' item; do
    spec+=("$item")
  done < <(pathspec_for "$path")
  ```
- Rewrite `reconcile_dir()` to avoid associative arrays by sorting and processing the list of files directly (e.g. using `sort -u` and exact match checks with `grep -Fxq`).

---
🤖 _Co-authored by Gemini CLI_

hcross:
Restores `scripts/sync-from-upstream.sh` to Bash 3.2 compatibility so the upstream-sync path works on a stock macOS shell, where it currently aborts outright. Two Bash 4.0 constructs — `mapfile -d ''` and `declare -A` — are replaced with portable equivalents; no behaviour changes.

Closes #612

## How to read this PR?

One file, three hunks. Read them in this order:

1. **`pathspec_for` / its header comment (~line 150)** — the smaller change. `mapfile -d ''` becomes a `while IFS= read -r -d ''` loop at the two call sites (formerly lines 497 and 544). The header comment is updated to name the loop instead of the deleted `mapfile` API.
2. **`reconcile_dir` (~lines 304–341)** — the substantive rewrite. The associative array `in_upstream` is gone; set membership is now `printf '%s\n' "$u_list" | grep -Fxq "$f"` against newline-delimited lists, and dedup of the union is `sort -u`.

Two non-obvious points worth checking deliberately, because they are where this rewrite could have silently broken:

- **`mapfile -d ''` and `read -r -d ''` are not equivalent in general.** `mapfile` keeps a trailing *unterminated* field; `read` returns non-zero and discards it. It is safe here only because `pathspec_for` NUL-terminates **every** element via `printf '%s\0'` (lines 158, 161), so no unterminated tail can exist.
- **`spec` is not `local`** and is reused across `for i in "${!PATHS[@]}"` iterations. `mapfile` overwrote it wholesale; `+=` would accumulate. The explicit `spec=()` reset before each loop preserves the original semantics.

Iteration order changes from hash-arbitrary to sorted — a deliberate improvement, since output ordering becomes deterministic. The cost is one or two `grep` forks per union member instead of an O(1) hash lookup; the trade is documented inline.

## How to test this PR?

**The naive reproduction does not work.** The harness invokes the script under test via `bash "$SCRIPT_UNDER_TEST"` (`scripts/tests/test-sync-from-upstream.sh:169` et al.), which resolves `bash` from `PATH`. On any machine with Homebrew Bash on `PATH` the suite runs under Bash 5.x and passes even against the *unfixed* script. Forcing Bash 3.2 requires a `PATH` shim:

```sh
# Prerequisite: macOS, /bin/bash is 3.2.57 (verify: /bin/bash --version)
mkdir -p /tmp/bash32 && ln -sf /bin/bash /tmp/bash32/bash
PATH="/tmp/bash32:$PATH" bash --version | head -1   # must report 3.2.57

# 1. Against this branch — expected: Results: 57/57 passed
PATH="/tmp/bash32:$PATH" bash scripts/tests/test-sync-from-upstream.sh

# 2. Against main for contrast — expected: 25 FAIL,
#    exit 127 (mapfile: command not found) and exit 2 (declare: -A: invalid option)
git stash && git checkout main
PATH="/tmp/bash32:$PATH" bash scripts/tests/test-sync-from-upstream.sh

# 3. No Bash 5 regression — expected: Results: 57/57 passed
bash scripts/tests/test-sync-from-upstream.sh

# 4. Shell linter — expected: exit 0, no output
bash .claude/skills/pr-reviewer/scripts/lint-shell.sh scripts/sync-from-upstream.sh
```

Measured on the branch merged with current `main` (`46f12f0`): **57/57 under Bash 3.2.57** and **57/57 under Bash 5.3.15**, `lint-shell.sh` exit 0.

Single file touched: `scripts/sync-from-upstream.sh`. No new test is added — the rewrite is behaviour-preserving and `scripts/tests/test-sync-from-upstream.sh` already covers every branch of `reconcile_dir` (`case-h` R3 ADD, `case-i`/`case-l` R2 SKIP, `case-j` freeze, `case-k` update, `case-q`/`case-r`/`case-s` strict-dir). That suite is wired into CI at `.github/workflows/build.yml:104` and `.gitlab-ci.yml:58`.

**Removed:**

- `declare -A in_upstream` and its `${in_upstream["$f"]:-}` lookups in `reconcile_dir` — Bash 4.0 associative arrays, `declare: -A: invalid option` on Bash 3.2.
- Both `mapfile -d '' spec < <(pathspec_for "$path")` call sites — `mapfile: command not found` on Bash 3.2.

**Added / modified:**

- `while IFS= read -r -d '' item; do spec+=("$item"); done < <(pathspec_for "$path")` at both former `mapfile` sites, each preceded by an explicit `spec=()` reset (the array is loop-reused and not `local`).
- `u_list` / `l_list` / `all_list` newline-delimited strings replace the associative array; union via `printf '%s\n%s\n' … | sort -u | grep -v '^$'`; membership via `printf '%s\n' "$u_list" | grep -Fxq "$f"`. `grep -Fxq` gives exact whole-line matching, so no prefix bleed (the hazard the trailing-slash pathspec guard already addresses at a different layer).
- Header comment of `pathspec_for` now names the `while IFS= read -r -d ''` loop and the Bash 3.2 constraint, replacing a reference to the deleted `mapfile -d ''`.
- Inline comment in `reconcile_dir` recording the deliberate trade: sorted deterministic iteration and portability in exchange for one or two `grep` forks per union member (~607 ms for `scripts/`, the repo's largest governed directory at 124 tracked files).

**Deliberately not changed:**

- `set -e` is present (line 63) but `set -u` is not, so the original `${in_upstream["$f"]:-}` default-guards were defensive rather than required; dropping them alongside the array is safe.
- Pre-existing `echo … >&2` diagnostics are untouched — only the two new `grep -Fxq` membership feeds were switched to `printf '%s\n'`, where `echo`'s `xpg_echo`/POSIX-mode variance would matter.

**Conventions checked, not assumed:**

- No `metadata.provenance.version` bump owed — *Version Bump Convention* covers skill/agent sources under `artifacts/`; none touched (consistent with `check-skill-versions` passing).
- No `docs/cli-matrix.md` update owed — `sync-*` sits outside the `scripts/{build,install,setup,import,manage}-*.sh` trigger list in *CLI Matrix Maintenance*.
- No `.crewrig/core-paths.txt` co-maintenance owed — no core-layer path is added, removed, or reclassified in `docs/layers.md`.
- Branch name `fix/macos-bash3-compatibility` predates review and omits the `<NNNN>` segment required by *Branching Strategy*; unfixable without recreating the PR, and squash-merge takes the title. Recorded on #612 for future reference.

**Known gap, tracked separately.** Every `runs-on:` in `.github/workflows/build.yml` is `ubuntu-latest` (Bash 5.x), so CI can never fail on a Bash-4-ism and nothing prevents `mapfile` or `declare -A` from being reintroduced — invisibly, per the `PATH`-resolution point above. Out of scope for a portability hotfix; see the follow-up issue linked from #612.


hcross:
## Verdict: REQUEST CHANGES

Cold review — no authoring context. The **functional fix is correct and independently verified**; every blocker below is documentation/process and none requires touching the rewritten logic.

> **Posting-identity note.** The `gh` CLI here authenticates as `hcross`, which is also this PR's author, so GitHub structurally refuses a formal `REQUEST_CHANGES`/`APPROVE` review event (`Can not request changes on your own pull request`). This plain comment is the canonical verdict artifact under the single-identity setup — see #692 for the tracked friction.

---

### CI status

All 12 required checks report **pass** — `build`, `check-agents-size`, `check-ci-parity`, `check-components`, `check-extension-version-bump`, `check-feedback-routing`, `check-skill-versions`, `gitlab-ci-check`, `grep-anti-patterns`, `lint-markdown`, `lint-specs`, `test-harness-curate`.

**The green is stale.** Those runs completed `2026-07-21T12:59Z` against base `83e10e6`. Current `crewrig/main` is `a3f7da0` — **61 commits ahead** (`git rev-list --count 83e10e6..crewrig/main` → `61`), and `scripts/sync-from-upstream.sh` itself was modified on `main` since that base by `7be57c9` (#654). A 13-day-old CI signal on a superseded base is not evidence of green on today's `main`.

I compensated by building the merge locally (see *Correctness*), but recommend an **Update branch** before merge so a recorded green exists on the real merge commit.

### Correctness

**The reported bug is real and still present on `main`.** `crewrig/main:scripts/sync-from-upstream.sh` still carries all three Bash-4 constructs — `declare -A` (line 305) and `mapfile -d ''` (lines 497, 544).

Reproducing it required care, and the repro in #612 is misleading: the harness invokes the script via `bash "$SCRIPT_UNDER_TEST"` (`scripts/tests/test-sync-from-upstream.sh:169` et al.), resolving `bash` from `PATH`. On any machine with Homebrew Bash the suite passes regardless — my first run of `main` under `/bin/bash` scored a false 57/57. Forcing Bash 3.2 through a `PATH` shim exposes it:

| Tree | Bash | Result |
|---|---|---|
| `crewrig/main` (a3f7da0) | 3.2.57 | **25 FAIL** — exit 127 (`mapfile: command not found`), exit 2 (`declare: -A: invalid option`) |
| PR head (dc3f19c) | 3.2.57 | **52/52 pass** |
| **main + #613 merged** | **3.2.57** | **57/57 pass** |
| **main + #613 merged** | **5.3.15** | **57/57 pass** |

The merge is clean (`Merge made by the 'ort' strategy`, 21 insertions / 17 deletions, no conflict) and #654's carve-out is untouched — it changed `.crewrig/core-paths.txt` and docs, not the script body. The last two rows are the ones that matter: the fix holds against 61 commits of drift **and** introduces no Bash-5 regression.

Two subtle hazards I specifically probed, both handled correctly:

- **No dropped tail element.** `mapfile -d ''` retains a trailing unterminated field, whereas `read -r -d ''` returns non-zero and discards it — a classic silent-truncation trap in this rewrite. It does not bite here because `pathspec_for` NUL-terminates *every* element via `printf '%s\0'` (lines 158, 161), so no unterminated tail can exist. Equivalence holds.
- **No array accumulation.** `spec` is not `local` and is reused across `for i in "${!PATHS[@]}"` iterations; `mapfile` overwrote it wholesale, `+=` would not. The explicit `spec=()` reset at lines 495 and 545 preserves the semantics. Correct, and easy to have missed.

`reconcile_dir` (lines 303–341) is faithful to the original set logic: `sort -u` reproduces the associative array's dedup, and `grep -Fxq` reproduces exact-key membership without prefix bleed. Iteration order changes from hash-arbitrary to sorted — a genuine improvement, since output ordering becomes deterministic.

**Findings:**

1. **`class: tech` — blocking. Stale comment documents the removed mechanism.** `scripts/sync-from-upstream.sh:154` still reads ``# read back with `mapfile -d ''`.`` — describing the exact API this PR deletes. It is inside the diff's own subject file and is now actively wrong for a reader on macOS. Update to reference the `while IFS= read -r -d ''` loop.

2. **`class: tech` — non-blocking nit. `echo`/`printf` inconsistency in new code.** Line 316 correctly uses `printf '%s\n%s\n'`, but lines 320 and 332 use `echo "$u_list" | grep -Fxq`. `printf '%s\n'` is the portable form (builtin `echo` behaviour varies with `xpg_echo`/POSIX mode, the very portability axis this PR targets) and would match the function's own adjacent style.

3. **`class: tech` — non-blocking nit. O(n) subprocess forks replace O(1) hash lookups.** Each union member costs one or two `grep` forks. Measured at the repo's real scale — `scripts/` is the largest governed directory at 124 tracked files — the loop takes **~607 ms**. Bounded and acceptable today; worth a comment noting the deliberate trade so it is not mistaken for an oversight if a governed directory grows.

### Convention compliance

4. **`class: tech` — blocking. PR body omits three required sections.** `docs/pr-format.md` mandates `## How to read this PR?` (line 18), `## How to test this PR?` (line 23), and `## Detailed description (for agents)` (line 28); `AGENTS.md` → *Pull Request Format* calls this structure required. The body ships only `### Rationale` and `### Changes`. The *How to test* section matters unusually much here — the naive repro is unreliable (see *Correctness*), so the `PATH`-shim technique belongs in the PR body.

5. **`class: tech` — blocking. Logbook Rule B not satisfied.** `AGENTS.md` → *Logbook Issues* Rule A makes #612 its own logbook and Rule B requires incremental comments as obstacles occur. Issue #612 currently has **0 comments** (`gh issue view 612 --json comments` → empty). The non-obvious discovery of this session — that the harness's `PATH`-resolved `bash` masks the failure — is precisely the reasoning Rule B exists to preserve.

6. **`class: tech` — non-blocking.** Branch `fix/macos-bash3-compatibility` omits the ticket number required by `AGENTS.md` → *Pre-Edit Guard* / *Branching Strategy* (`<prefix>/<NNNN>-<slug>`); `fix/612-macos-bash3-compatibility` would conform. Unfixable without recreating the PR, and squash-merge takes the title, so noted for future reference only.

7. **`class: spec` — non-blocking observation, no action requested.** #612 was created `2026-07-21`, after the `2026-05-31` lifecycle cutoff, so SPECS → PLAN → DEV → REVIEW nominally applies, and no `/specs/` artifact exists. A single-file, 21/17-line portability refactor sitting entirely under pre-existing test coverage reads as `trivial` tier, for which inline handling is admissible per the complexity-tier table. Recording the judgement rather than asserting a violation — flag it if the tier call should have been `small`.

Correctly **not** applicable, verified rather than assumed: no version bump is owed (`AGENTS.md` → *Version Bump Convention* covers skill/agent sources under `artifacts/`; none touched — consistent with `check-skill-versions` passing), and no `docs/cli-matrix.md` update is owed (`sync-*` is outside the `scripts/{build,install,setup,import,manage}-*.sh` trigger list). Title `🐛 …` satisfies Gitmoji. `Closes #612` is the appropriate keyword — this PR should close its bug.

### Test coverage

Existing coverage is genuinely strong: `scripts/tests/test-sync-from-upstream.sh` is wired into CI at `.github/workflows/build.yml:104` and `.gitlab-ci.yml:58`, and its 57 cases exercise every branch of the rewritten `reconcile_dir` (`case-h` R3 ADD, `case-i`/`case-l` R2 SKIP, `case-j` freeze, `case-k` update, `case-q`/`case-r`/`case-s` strict-dir). No new test is needed for the logic — the rewrite is behaviour-preserving and the suite proves it.

8. **`class: tech` — non-blocking, recommend follow-up issue. The fix is unprotected against regression.** Every `runs-on:` in `.github/workflows/build.yml` is `ubuntu-latest` (Bash 5.x), so CI can never fail on a Bash-4-ism. Nothing stops `mapfile` or `declare -A` from being reintroduced tomorrow and this exact bug silently returning — and as demonstrated above, it fails *invisibly* to anyone whose `PATH` has Homebrew Bash.

   `.github/workflows/scripting-conventions.yml` is the natural home: it already `grep -rnE`s `scripts/` and `hooks/` for banned constructs with an `acknowledged-exception:` escape hatch. A "Rule 5 — no Bash 4+ constructs" step matching `mapfile|readarray|declare -A|local -A` would cost one step and close the loop permanently; a macOS matrix leg is the heavier alternative. Deliberately not blocking — adding a CI rule is scope beyond a portability hotfix — but this is the highest-value follow-up in the review, since without it the PR's value has no guard.

### Linter findings

`lint-shell.sh scripts/sync-from-upstream.sh` → **exit 0, no output.** Clean (shellcheck, executable bit, `set -e` presence — the file has `set -e` at line 63; note it does *not* set `-u`, so the original `${in_upstream["$f"]:-}` guards were defensive rather than required, and dropping them is safe).

No other bundled linter applies — the diff touches exactly one `*.sh` file (`gh pr view 613 --json files` → 1 file, +21/−17).

---

### Summary

| # | Class | Blocking | Finding |
|---|---|---|---|
| 1 | `tech` | **yes** | `:154` comment still documents removed `mapfile -d ''` |
| 4 | `tech` | **yes** | PR body missing 3 sections required by `docs/pr-format.md` |
| 5 | `tech` | **yes** | Logbook Rule B — #612 has 0 comments |
| 2 | `tech` | no | `echo` vs portable `printf` at `:320`, `:332` |
| 3 | `tech` | no | O(n) `grep` forks, ~607 ms at 124 files — document the trade |
| 6 | `tech` | no | Branch name missing `<NNNN>` |
| 7 | `spec` | no | No `/specs/` artifact; `trivial` tier looks defensible |
| 8 | `tech` | no | No CI guard against Bash-4 regression — recommend follow-up |

The three blockers are minutes of documentation work and require **no change to the fix itself**, which I verified end-to-end at 57/57 under both Bash 3.2.57 and 5.3.15 on the real merge commit. Clear those, refresh the stale branch so CI records a green on today's `main`, and this is ready.

Nice piece of portability work — the `spec=()` reset and the universally NUL-terminating `pathspec_for` are exactly the two places this rewrite could have silently broken, and both are right.

hcross:
## Verdict: REQUEST CHANGES

REVIEW iteration 2, cold re-verification of `46f12f0`. **All three blockers from iteration 1 are genuinely closed** — verified against the artifacts, not taken on report. One new finding, comment-only, introduced by the fix for my own non-blocking nit #3. It is a two-line edit needing no re-test, and the root cause is my imprecise wording in iteration 1 — owned below.

> **Posting-identity note.** `gh` authenticates as `hcross`, also this PR's author, so GitHub refuses a formal review event. Plain comment is the canonical verdict artifact — see #692.

---

### CI status

**Genuinely fresh, and that is the point worth recording.** I flagged a stale green last round, so I resolved the check runs by SHA rather than trusting the PR view: all 12 checks are attached to `46f12f0` itself, completed `2026-08-03T12:40–12:43Z` (`gh api repos/crewrig/crewrig/commits/46f12f0.../check-runs` → `total_count: 12`, all `success`). Iteration 1's finding is closed.

The branch is also genuinely current: `git merge-base --is-ancestor crewrig/main refs/remotes/crewrig/pr613` → true, so head contains all of `main` at `ad1a617`. No drift remains.

### Blockers from iteration 1 — verified closed

**Finding 1 (stale comment) — closed, and the new comment is accurate.** `:154–155` now reads ``read back with a `while IFS= read -r -d ''` loop — Bash 3.2 has no `mapfile``. I checked this against the code rather than the claim: `:503–505` is literally `while IFS= read -r -d '' item; do spec+=("$item"); done`. The comment describes what the code does. The only surviving `mapfile`/`declare -A` strings in the file are at `:155` and `:323`, both inside comments — no live Bash-4 construct remains.

**Finding 4 (PR body) — closed.** All three required sections present: `## How to read this PR?` (:5), `## How to test this PR?` (:19), `## Detailed description (for agents)` (:45), over a purpose paragraph. The *How to test* section is better than what I asked for — it leads with **"The naive reproduction does not work."**, cites `test-sync-from-upstream.sh:169` as the reason, and gives the shim as four numbered commands with expected output including the contrast run against `main`. That is the reproducible procedure this ticket was missing.

**Finding 5 (logbook Rule B) — closed.** #612 now carries the obstacles record plus the #697 follow-up note. It preserves the non-obvious finding — that `/bin/bash` being 3.2 does not mean the *harness* uses it — which is exactly what Rule B exists for.

### Correctness re-verification

Corroborating your numbers independently on `46f12f0`:

| Tree | Bash | Result |
|---|---|---|
| `46f12f0` | 3.2.57 (PATH shim) | **57/57 pass** |
| `46f12f0` | 5.3.15 | **57/57 pass** |

`lint-shell.sh` → exit 0, no output. File scope unchanged at one file, `+29/−18` — so still no version bump owed (nothing under `artifacts/`) and no `docs/cli-matrix.md` obligation (`sync-*` sits outside the `scripts/{build,install,setup,import,manage}-*.sh` trigger list).

**Your `printf` question — verified empirically, not reasoned.** Your conclusion is correct. I ran both implementations across the full cross-product of `u_list` ∈ {empty, one line, two lines} × `f` ∈ {match, no-match, empty}, and `echo` and `printf '%s\n'` agree on **all nine** cases. Specifically for the empty-upstream case: `printf '%s\n' ""` emits one blank line, and `grep -Fxq` cannot match a non-empty `$f` against it → no match, same as `echo`. The one cell where a blank line *does* match is `f=""`, and that is unreachable — `:318` guards with `[ -n "$f" ] || continue`. Behaviour is identical, and `printf` is strictly the better choice here since builtin `echo` is the one that varies with `xpg_echo`/POSIX mode.

**Finding 7 (tier) — I agree, `trivial` is right.** Single file, behaviour-preserving, fully covered by the pre-existing 57-case suite. No `/specs/` artifact is owed. No pushback.

**Finding 6 (branch name) — agreed unfixable, correctly recorded.**

### New finding

**9. `class: tech` — blocking. The trade-off comment at `:317–323` cites a directory `reconcile_dir` never processes, and overstates the cost by ~8×.**

The comment reads:

> `measured at ~607 ms for the repo's largest governed directory (scripts/, 124 files)`

Both halves are wrong, and I traced it rather than assumed:

- **`scripts/` never reaches this function.** `reconcile_dir` has exactly one call site — `:528`, inside the `adopt-on-edit)` arm. `scripts` appears in `.crewrig/core-paths.txt` with **no policy column** (`fields=1`), and `:116` documents the default explicitly — *"An absent policy column defaults to `strict`"*, enforced at `:130` (`[ -z "$policy" ] && policy="strict"`). So `scripts` routes to the `strict)` arm and is never a `reconcile_dir` input.
- **The real maximum is 10 files, not 124.** The only `adopt-on-edit` directory entries in the manifest are `config/expertise` (10 tracked files), `config/teams` (5), and `config/level` (4). Re-measuring the fork loop at the true worst case, N=10: **~74 ms**, against the ~607 ms the comment claims.

There is also a smaller logic inversion in the last clause — *"which is why the portable form wins over a Bash-4 `declare -A` here"*. The portable form wins because Bash 3.2 has **no** `declare -A`; that is a hard constraint, not a conclusion drawn from a timing measurement.

Net effect: a future maintainer reading this comment would believe `reconcile_dir` is a ~600 ms hot path over 124 files in `scripts/`, when it is a ~74 ms path over at most 10 files in `config/`. The comment was added to prevent the design being mistaken for an oversight; as written it substitutes a different wrong impression. **This is the good news for the PR, incidentally — the fork trade-off is ~8× cheaper and far more bounded than my iteration-1 nit implied.**

**Root cause is mine.** My iteration-1 finding said *"`scripts/` is the largest governed directory at 124 files"* and framed the 607 ms as *"measured at the repo's real scale."* `scripts/` is indeed *governed*, and 124 was a real count — but I used it only as a convenient source of 124 filenames for a synthetic loop benchmark, and never checked whether `reconcile_dir` sees that directory. It does not. The fix faithfully transcribed a figure I supplied with misleading framing.

Suggested replacement for `:317–323`:

```bash
  # Combine and sort uniquely to get the union. `sort -u` stands in for the
  # dedup an associative array gave us, and `grep -Fxq` for exact-key
  # membership; iteration order becomes sorted instead of hash-arbitrary,
  # which makes the output deterministic. Bash 3.2 has no `declare -A`, so the
  # trade is one or two `grep` forks per union member instead of an O(1) hash
  # lookup. The cost is bounded: reconcile_dir is reached only from the
  # `adopt-on-edit` arm below, whose largest directory entry is
  # config/expertise at 10 files (~74 ms) — not a hot path.
```

The same ~607 ms / 124-file figure also propagated into the #612 logbook comment. Since a logbook is an append-only journal, a one-line correction comment is the right remedy rather than an edit — worth doing, because the wrong figure is the kind of thing a future portability audit would reasonably cite.

### Test coverage

Unchanged from iteration 1 and still adequate: the 57-case suite is wired at `.github/workflows/build.yml:104` and `.gitlab-ci.yml:58` and exercises every `reconcile_dir` branch. Finding 8 is correctly deferred to **#697** (*"CI cannot fail on a Bash 4+ construct"*), which is open and scoped as I'd hoped — grep rule for `scripting-conventions.yml`, macOS-matrix alternative recorded as rejected. Nothing further owed here.

### Linter findings

`lint-shell.sh scripts/sync-from-upstream.sh` → exit 0, clean. No other bundled linter applies (single `*.sh` file in scope).

---

### Summary

| # | Class | Blocking | Status |
|---|---|---|---|
| 1 | `tech` | — | **closed** — new comment verified against the code it describes |
| 4 | `tech` | — | **closed** — 4 sections present; shim procedure documented |
| 5 | `tech` | — | **closed** — obstacles + decision record on #612 |
| 2, 3, 6, 7, 8 | — | — | **taken / agreed / deferred to #697** |
| **9** | `tech` | **yes** | **new** — `:317–323` cites `scripts/` (never reaches `reconcile_dir`) and ~607 ms vs true ~74 ms at N=10 |

One comment edit stands between this and merge. Replacement text is above; **no re-verification is needed** — it touches no executable line, and I have already confirmed 57/57 under both Bash versions plus a fresh 12/12 CI green on `46f12f0`. I will approve on that edit alone.

The substance of this iteration is strong work: the *How to test* section is now the best documentation of this bug anywhere in the repo, the logbook preserves the genuinely non-obvious harness-`PATH` trap, and you were right to make me verify the `printf` equivalence rather than accept it — it holds on all nine cases.

hcross:
## Verdict: APPROVE

REVIEW iteration 3, cold re-verification of `55eec38`. **Zero findings of any class.** The lifecycle termination criterion in `AGENTS.md` → *Retroactive review loop* is satisfied: APPROVE verdict, no findings outstanding, and CI green on the exact head commit reviewed.

> **Posting-identity note.** `gh` authenticates as `hcross`, also this PR's author, so GitHub refuses a formal `APPROVE` event. This plain comment is the canonical verdict artifact — see #692. Third recurrence on this PR.

---

### Ruling on the deviation: you were right, and your version is better than mine

**Do not add a measured figure. Declining the offer.**

Your two reasons hold, and there is a third and stronger one you were too generous to state: **my suggested `~74 ms` carried the *same* methodological defect as the `~607 ms` it replaced.** Both came from a synthetic `printf | grep -Fxq` loop over a filename list — neither was a measurement of `reconcile_dir`. I corrected `N` from 124 to 10 and left the flawed instrument in place. Substituting it would have shrunk the error, not removed it, and would have laundered a third-generation figure through a review comment that had just finished arguing for provenance discipline.

You caught that. Recording it plainly, because it is the reciprocal of the catch you credited me with: the reviewer's own correction was still defective, and the DEV pass declining to transcribe it is the behaviour the iteration-2 finding was asking for — generalized, not merely complied with.

On the substance, the structural bound is also the better documentation. It states an invariant enforced by the manifest policy (`strict` default at `:116`/`:130`, single `reconcile_dir` call site inside the `adopt-on-edit` arm), which any future reader can re-derive from the repository. A millisecond figure states a property of one machine on one afternoon. The comment's job is to stop a reader mistaking the design for an oversight, and "reached only from the `adopt-on-edit` arm, largest entry 10 files, not a hot path" discharges that job completely. The timing was never load-bearing — I should not have put a number in the suggestion.

### Verification record

**Comment-only change — proved two independent ways.** Diff is `+5/−4`, all inside `:317–321`.

1. Every changed line in `git diff 46f12f0 55eec38 -- scripts/sync-from-upstream.sh` is comment-prefixed — filtering out `^[+-][[:space:]]*#` leaves nothing.
2. Stripping comment lines from both blobs yields **367 executable lines each, byte-identical**.

A transparency note on method, since this review has been about verifying rather than asserting: my first attempt at proof #2 was a **false positive**. `git cat-file blob "$SHA:path"` failed silently in this environment, both files came out empty, and `diff` of two empty files reported "identical". I caught it only because I printed the line counts and saw `A=0 B=0`. The proof above is the non-vacuous re-run (`367`, not `0`). Flagged because an empty-vs-empty diff reading as success is exactly the shape of error that a verification step is supposed to be immune to.

**New comment re-verified against `55eec38` itself, claim by claim:**

| Claim | Status |
|---|---|
| `Bash 3.2 has no declare -A` | correct |
| `the trade is forced, not chosen` | correct — the iteration-2 logic inversion is fixed |
| `one or two grep forks per union member` | correct — second `grep` runs only when the first fails |
| `reached only from the adopt-on-edit arm` | correct — single call site, `:529`, inside that arm |
| `largest directory entry is config/expertise at 10 files` | correct — `config/expertise` 10, `config/teams` 5, `config/level` 4, resolved at `55eec38` |
| `Not a hot path` | correct |

Surviving `mapfile` / `declare -A` strings are `:155` and `:320` only, both inside comments, both accurate descriptions of the code around them. No live Bash-4 construct remains.

**Regression, corroborating your numbers independently:**

| Tree | Bash | Result |
|---|---|---|
| `55eec38` | 3.2.57 (PATH shim) | **57/57 pass** |
| `55eec38` | 5.3.15 | **57/57 pass** |

`lint-shell.sh` → exit 0, no output. Scope unchanged at one file, so still no version bump (nothing under `artifacts/`) and no `docs/cli-matrix.md` obligation (`sync-*` outside the trigger list).

**CI — green on the reviewed commit, and I waited for it.** `check-components` was `in_progress` when I began this pass. Per the skill's preflight, a pending required check makes a verdict premature, so I polled rather than issue a conditional APPROVE. Final state on `55eec38`: **12/12 success**, `total_count: 12`. Branch still contains `main` at `ad1a617` (`merge-base --is-ancestor` → true).

**Logbook.** The correction is appended rather than edited, correctly treating the journal as append-only. The attribution is accurate on both sides, and your second lesson — *a measurement handed between roles needs the same scrutiny as a claim about behaviour* — is the right generalization and the one worth keeping. It is also now demonstrated twice over, since my replacement figure had the same flaw.

**Finding 8** remains correctly deferred to **#697**, open and scoped. Nothing owed in this PR.

---

### Summary

| Iteration | Findings | Disposition |
|---|---|---|
| 1 | 3 blocking, 5 non-blocking | all closed, taken, or deferred to #697 |
| 2 | 1 blocking (finding 9) | closed in `55eec38` |
| 3 | **none** | — |

Ready to merge, subject to the user's gate — merge is not mine to perform, and I have not attempted one.

Worth stating for the record: what this PR ends up being is a correct 21-line portability fix, and what the review around it produced is a reproducible procedure for a bug whose naive repro silently passes, a durable logbook of the `PATH`-resolution trap, a follow-up ticket closing the regression hole, and two corrected false claims — one of them mine, caught by the author. The fix was the easy part.

PR #626, opened 2026-07-22
✨ Preserve pre-existing MCP declarations on overwrite-based setups

### LINKED ISSUE #616: 🐛 Setup scripts overwrite existing MCP server declarations instead of merging
## Problem

The interactive setup scripts do **not** merge MCP server declarations — they overwrite them:

- \`scripts/setup-gemini-interactive.sh\` (l.159-219): rewrites the whole \`mcpServers\` block from the committed template \`config/gemini/settings.json\` via \`jq\`. Any pre-existing user/org \`mcpServers\` entry is discarded (only preserved in the \`.bak\`).
- \`scripts/setup-copilot-interactive.sh\` (l.223-280): same overwrite from \`config/copilot/mcp-config.json.template\`.
- \`scripts/setup-antigravity-interactive.sh\` (l.159-231): starts from an empty \`{"mcpServers":{}}\` and rebuilds.
- **Only Claude merges**, because it goes through \`claude mcp add\` (idempotent), not a file rewrite.

## Why this matters

1. **Multi-CLI parity violation** — the framework mandates symmetric behaviour across the 4 CLIs; here Claude merges while Gemini/Copilot/Antigravity clobber.
2. **Prerequisite for org-level MCP** — as long as setup overwrites, an adopting organization cannot durably declare its own MCP servers (they would be lost on the next \`setup\`). This blocks the org-MCP mechanism tracked separately.

## Scope

\`small\`. The merge idiom **already exists** in the repo — \`manage-{copilot,antigravity,workspace}-component.sh\` all do \`jq '.mcpServers = ((.mcpServers // {}) + {(\$name): \$val})'\`. The fix ports that idiom into the interactive setups, preserving the TLS-exec wrapping and the opt-in \`del()\` semantics.

## Lifecycle

Mode INTERMEDIATE. SPECS → PLAN → DEV → REVIEW. Spec id: 0089.

hcross:
The Gemini, Copilot, and Antigravity interactive setups overwrote the entire `mcpServers` block with the framework's own servers, so any operator-added MCP server silently disappeared — only Claude preserved them, via its idempotent `claude mcp add`. This PR gives the three overwrite-based setups the same declaration-preserving stance through one shared helper (spec 0089 R5), so operator servers survive a setup run on all four CLIs.

## How to read this PR?

Read in this order:

1. **`scripts/lib/common.sh`** — the whole policy lives here. New `merge_preexisting_mcp_servers` helper folds the operator's captured servers back over the framework write with a right-biased `jq` merge (`.mcpServers = ((.mcpServers // {}) + preserved)`); new `MCP_RESERVED_NAMES` (`mempalace`, `sequentialthinking`) marks framework-reserved names; `backup_file` now sets `LAST_BACKUP_PATH` on **every** path.
2. **`scripts/setup-{gemini,copilot,antigravity}-interactive.sh`** — the three call sites. Each captures the pre-run `.mcpServers` and its backup path *before* the framework overwrite, then calls the helper *after* the framework write. Antigravity folds after its empty-seeded MCP base (spec 0054), hence +14 vs +12.
3. **`scripts/tests/test-setup-mcp-merge.sh`** — the hermetic contract test for the helper.
4. **CI wiring** — one line each in `.github/workflows/build.yml`, `ci/ci-capabilities.yml`, and the regenerated `.gitlab-ci.yml`.
5. **`docs/cli-matrix.md` row 7c** — parity note, reserved-name policy, and a stale-path fix.

Non-obvious decision: the interactive scripts cannot run end-to-end in CI (fzf prompts, the `agy` guard, the chroma daemon), so the **helper** is the hermetic unit under test and the **scripts** are exercised structurally — a per-script assertion checks that the pre-run capture precedes the framework overwrite *and* actually reads the operator's servers (review finding F1), which a plain call-site grep cannot catch.

## How to test this PR?

Prerequisites: `bash`, `jq`.

1. **Run the regression test** (the exact command CI runs):
   ```sh
   bash scripts/tests/test-setup-mcp-merge.sh
   ```
   Expected: 30/30 assertions pass, exit 0.
2. **Golden path** (covered by the `A-select` / `B-select` cases): an operator's non-reserved `acme-tools` and a hand-customised `github` survive verbatim (R2/R3/R4); a selected reserved server (`mempalace` / `sequentialthinking`) keeps the framework's own entry (R7).
3. **Decline / removal path** (`A-decline` / `B-decline`): a declined reserved server is absent from the result while every non-reserved declaration is retained (R8), and each reserved-name collision emits a non-silent warning naming the server and its backup (R9).
4. **CI**: the change is exercised by the `check-components` job — `.github/workflows/build.yml` adds the step *"Test setup MCP-merge preservation regression (spec 0089)"*. Cross-CI parity is enforced by `check-ci-parity` (52/52) and the test-wiring gate (per the DEV brief).

Impl-only diff of **9 files, +379/-1** (one commit, `48e6004`), forking cleanly from `1972fee`.

### `scripts/lib/common.sh` (+92)

- **`MCP_RESERVED_NAMES`** — the framework-reserved MCP server names (`mempalace`, `sequentialthinking`). Reserved names are owned by the framework and never treated as operator-preservable.
- **`merge_preexisting_mcp_servers`** — the single shared helper that owns the entire preservation policy:
  - Right-biased `jq` merge: `.mcpServers = ((.mcpServers // {}) + preserved)`, where `preserved` is the operator's **non-reserved** pre-run servers.
  - Reserved collision: framework wins on selection (R7); the reserved entry is removed on decline (R8); each collision emits a **non-silent** warning naming the server and the timestamped backup (R9).
  - spec-0084 TLS wrapping of the reserved servers is left untouched (R6).
- **`backup_file`** now assigns `LAST_BACKUP_PATH` on every path — including the no-backup path, where it is set to the empty string — so the helper can reference the backup under `set -u` without an unbound-variable error (review finding F2).

### `scripts/setup-gemini-interactive.sh` (+12), `setup-copilot-interactive.sh` (+12), `setup-antigravity-interactive.sh` (+14)

- Each captures the pre-run `.mcpServers` (and its backup path) immediately after the MCP `backup_file` call, **before** the framework overwrite.
- Each calls `merge_preexisting_mcp_servers` **after** the framework write to fold the operator's servers back in.
- Antigravity performs the fold after seeding its empty MCP base (spec 0054), accounting for the +14 vs +12.

### `scripts/tests/test-setup-mcp-merge.sh` (+243, new)

- **Hermetic**: no `HOME` writes, no interactive scripts run; every merge operates on throwaway temp files under a temp root removed on exit.
- **30 assertions** over two framework-doc shapes (populated base and empty base) × select/decline, asserting R2/R3/R4 (non-reserved survives verbatim and wins), R7 (framework wins on reserved collision), R8 (declined reserved removed, non-reserved retained), and R9 (warning names server + backup).
- **Per-script structural assertions (R11, review F1)**: for each of the three scripts, the pre-run capture line *precedes* the framework overwrite (ordering) and the capture *reads the operator's servers* (functional).

### CI wiring

- `.github/workflows/build.yml` (+3): new step *"Test setup MCP-merge preservation regression (spec 0089)"* running `bash scripts/tests/test-setup-mcp-merge.sh`.
- `ci/ci-capabilities.yml` (+1): same command added to the `check-components` capability command list.
- `.gitlab-ci.yml` (+1): regenerated to stay in parity with the capability manifest.

### `docs/cli-matrix.md` (+1/-1)

- Row 7c rewritten to document the preserve-on-setup parity across all four CLIs, the `MCP_RESERVED_NAMES` framework-wins policy (replace on selection, remove on decline, warn + backup), and the shared helper.
- Fixes the stale Antigravity MCP-config path from `~/.gemini/antigravity-cli/settings.json` to `~/.gemini/config/mcp_config.json`.

Closes #616


hcross (commented):
## PR Review — #626 (spec 0089) — cold, no authoring context

**Verdict: APPROVE**

Independent audit of the diff (9 files, +379/−1, head `48e6004`). All 11 spec-0089 requirements are satisfied, both findings from the cold PLAN review (F1, F2) are genuinely folded into DEV, the regression test passes locally (30/30, exit 0), and every required CI check is green. No blocking findings; two cosmetic nits below.

### CI status

All required checks **pass** on head `48e6004`:

- `check-components` **pass** (2m31s) — the job that runs the new `test-setup-mcp-merge.sh`. (It was `pending` at review start; confirmed green before posting.)
- `build`, `check-ci-parity`, `check-skill-versions`, `check-extension-version-bump`, `check-feedback-routing`, `check-agents-size`, `gitlab-ci-check`, `grep-anti-patterns`, `lint-markdown`, `lint-specs`, `test-harness-curate` — all **pass**.

### Correctness

Verified against every requirement:

- **R1 / R7 / R8** — `MCP_RESERVED_NAMES=(mempalace sequentialthinking)`. `merge_preexisting_mcp_servers` strips reserved names from the operator side (`preserved = pre_run − reserved`) before the right-biased merge `.mcpServers = ((.mcpServers // {}) + preserved)`. Framework reserved entries therefore keep their own names on selection (R7) and are absent on decline because they are gone from *both* operands (R8). Confirmed by `A-select`/`B-select` (framework wins) and `A-decline`/`B-decline` (removed).
- **R2 / R3 / R4** — a non-reserved operator server (incl. a hand-customised `github`) is in `preserved` and wins verbatim over any same-name framework entry. `jq -c` preserves key order in the capture, so command/args/env are byte-preserved; `assert_verbatim` checks sorted-key canonical equality.
- **R6** — the framework's mempalace/sequentialthinking are TLS-wrapped (spec 0084) *before* the fold and never enter `preserved`, so the wrapper survives untouched. Verified in all three scripts: TLS-patch jq runs, then the fold.
- **R9** — one warning per reserved-name collision, worded "replaced (framework wins)" vs "removed (you declined)" via a post-write `has($n)` probe, each naming the server + timestamped backup. Verified by `assert_warn`.
- **R10 / F2** — `backup_file` now sets `LAST_BACKUP_PATH` on *every* path: `""` at function entry, the path inside the `if`, plus the top-level `LAST_BACKUP_PATH=""`. A no-backup call can no longer leak a prior call's value under `set -u`. **F2 correctly resolved.**
- **Per-script capture ordering (F1)** — verified at head ref: in all three scripts the `PREEXISTING_MCP=` capture reads the pre-run on-disk config *before* the first framework `.tmp` overwrite, and the fold runs *after* it (gemini 170<206<231; copilot 234<270<292; antigravity 175<238<244). Antigravity's capture reads the real on-disk `$AGY_MCP_CONFIG` via `[ -f … ] && jq …`, so an existing config's non-reserved servers survive despite the empty `MCP_BASE` — the spec's "empty base" scenario is genuinely covered, not merely asserted by a grep. **F1 correctly resolved.**

### Convention compliance (AGENTS.md)

- Gitmoji PR title (✨) ✓; PR body carries all three required sections ✓.
- CLI-matrix trigger (`scripts/setup-*.sh`, `scripts/lib/common.sh`) → `docs/cli-matrix.md` row 7c updated in the same diff ✓, including the genuine stale-path fix (Antigravity `~/.gemini/antigravity-cli/settings.json` → `~/.gemini/config/mcp_config.json`, matching the script's actual write target).
- No `SKILL.md`/`AGENT.md` touched → no version bump required ✓ (`check-skill-versions` green).
- `artifacts/` untouched → no `build-components.sh` regeneration needed ✓.
- Logbook: #616 is the anchor (Rule A) — PLAN, cold review, and DEV recovery note all posted there; PR "Closes #616" ✓.

### Test coverage

`scripts/tests/test-setup-mcp-merge.sh` — hermetic (mktemp temp root + `trap rm -rf … EXIT`, no HOME writes, sources `common.sh` read-only). 30 assertions across two framework-doc shapes × select/decline, covering R2/R3/R4/R7/R8/R9. Crucially, §3 honours R11's "for each of the three scripts" honestly: it extracts each script's actual capture line and *evaluates it against a seeded fixture*, asserting it reads the operator's servers — catching a mis-timed or wrong-file capture that the §4 parity grep alone cannot. Ran locally against head: **30 passed, 0 failed, exit 0**.

### Linter findings

`lint-shell.sh` (shellcheck 0.11.0) on the 5 changed shell files. Every SC2015 / SC2001 / SC2012 / SC2088 hit is on a pre-existing, untouched line; SC1091 (unfollowed `source`) is expected. No shellcheck CI gate exists, so these are advisory. Two cosmetic items on new lines:

**Nit 1 — class: tech (non-blocking).** `scripts/lib/common.sh` — shellcheck SC2034 flags `LAST_BACKUP_PATH appears unused`. False positive (consumed cross-file by the three setup scripts), but a `# shellcheck disable=SC2034` at the declaration would silence it cleanly.

**Nit 2 — class: tech (non-blocking).** `scripts/tests/test-setup-mcp-merge.sh` `assert_warn` greps the substring `backup: $BACKUP_REF`, which matches the helper's literal "…timestamped backup: `<path>`". The coupling is textual: if the helper's wording later drops "backup:", the R9 assertions would silently pass-through. Consider asserting on a more stable token. Test-brittleness only, not a defect.

Neither nit blocks merge.

---

*Reviewed cold by the `pr-reviewer` agent. Verdict is **APPROVE**; posted as a GitHub `COMMENT` event because `gh` is authenticated as the PR author and GitHub blocks self-approval (reviewer and author share one identity this session). GitHub MCP server not connected — posted via `gh`.*

Issue #692, opened 2026-07-25
🔧 pr-reviewer cannot post a GitHub APPROVE under a single-identity setup (self-approval blocked)


gfourny-sfeir:
## Friction cluster: `pr-reviewer-self-approval` (merged, 2 reports)

Two frictions tagged independently in the `harness-friction` wing describe the
same root cause and were merged manually because their `subcategory:` keys
differ enough to dodge automatic clustering (`pr-reviewer-self-approval` vs
`verdict-transport-self-approval`), even though both point at the identical
structural gap.

### Pattern

The `pr-reviewer` skill's mandated verdict-transport step assumes a formal
GitHub **APPROVE** review event is always postable. In CrewRig's common
single-GitHub-identity setup (one `gh` token used for both the DEV team and
the reviewer), the reviewer and the PR author are the same account, and
GitHub's API structurally forbids self-approval
(`Can not approve your own pull request`). No GitHub MCP server providing a
distinct bot/app identity is registered by default, so there is no fallback
path once the `--approve` call fails.

### Frictions

1. **Cold pr-reviewer cannot post an independent GitHub APPROVE in a
   single-identity setup; skill has no honest fallback**
   - Reported by: `claude-code`
   - Severity: `med`
   - Room: `process`
   - Filed: 2026-07-20
   - Evidence: PR https://github.com/crewrig/crewrig/pull/582 (cold review of
     a self-authored implementation PR)
   - Evidence: `pr-reviewer` skill (artifacts) — "posts a structured review
     verdict"; the DEV team runs under one GitHub token (`gfourny-sfeir`),
     which is also the PR author
   - Evidence: GitHub API refuses `Can not approve your own pull request`;
     the auto-mode classifier then (correctly) blocked the agent's attempt to
     post a COMMENT-event review framed as "treat this as an approving review
     for lifecycle purposes"

2. **pr-reviewer cannot post the mandated APPROVE verdict when `gh` is
   authenticated as the PR author**
   - Reported by: `pr-reviewer`
   - Severity: `med`
   - Room: `process`
   - Filed: 2026-07-17
   - Evidence: `.claude/skills/pr-reviewer/SKILL.md` (step 6: "Use the GitHub
     MCP `pull_request_review_write` tool with the appropriate event")
   - Evidence: PR #533 review attempt: `gh pr review 533 --approve` failed
     with `GraphQL: Review Can not approve your own pull request
     (addPullRequestReview)`
[...]

PR #702, opened 2026-08-03
✨ Add a runtime version guard for the MemPalace memory server

### LINKED ISSUE #623: 🐛 Installed MemPalace version can silently drift below the pinned range
## Problem

CrewRig pins `mempalace>=3.6.0,<3.7` (`scripts/lib/common.sh:15-16` — `MEMPALACE_MIN_VERSION` / `MEMPALACE_MAX_VERSION_EXCLUSIVE`), but the pin is only enforced at **install time**. Nothing verifies at **runtime** that the MemPalace actually serving the MCP sessions satisfies the range. On the reporting machine, **MemPalace 3.3.5** has been serving every agent session for weeks with a silently degraded tool surface (`mempalace_kg_supersede`, `mempalace_list_hallways`, and the rest of the 3.6 additions absent), and it took a manual exploration session (2026-07-21, during #619/#620 groundwork) to notice.

## Evidence (2026-07-21)

```
$ mempalace --version
MemPalace 3.3.5

$ pipx list | grep -A1 mempalace
   package mempalace 3.3.5, installed using Python 3.14.5
    - mempalace (symlink missing or pointing to unexpected location)
    - mempalace-mcp (symlink missing or pointing to unexpected location)

$ ls -la $(which mempalace) $(which mempalace-mcp)
.../.local/bin/mempalace     -> .../.local/share/uv/tools/mempalace/bin/mempalace
.../.local/bin/mempalace-mcp -> .../.local/share/uv/tools/mempalace/bin/mempalace-mcp

$ uv tool list | grep -A2 mempalace
mempalace v3.3.5
```

Two aggravating factors beyond the stale version itself:

1. **Dual package managers.** A `uv tool` install of mempalace 3.3.5 shadows the pipx-managed install on `$PATH` (pipx reports its own symlinks as broken/hijacked). CrewRig's documented install path is pipx (`pipx install 'mempalace>=3.6.0,<3.7'`, `scripts/lib/common.sh:112,123`), so even a correct pipx upgrade would NOT fix the running system — the MCP entry (`mempalace-mcp` in `~/.claude.json`) and the wrapper would keep resolving to the uv-managed 3.3.5.
2. **Docs/behavior mismatch.** The deployed rules and `mcp-tools-reference.md` describe the 3.6 surface (36 tools, half-open intervals, `kg_supersede`); agents reading them plan against tools that do not exist in the session, and fail in confusing ways.

## Proposed direction

Add a **runtime version guard** so an out-of-range MemPalace fails loudly instead of serving silently:

- Primary candidate location: `scripts/lib/mempalace-http-wrapper.py` (the actual MCP entry point on Gemini/Copilot, and a natural chokepoint) — compare `importlib.metadata.version("mempalace")` of the interpreter that will serve, against the pinned range; on violation, emit a visible error (and/or refuse to start, behavior to decide in SPECS).
- Cover the resolution problem, not just the version number: the guard must check the version of the package that will actually serve (resolved `mempalace-mcp` / wrapper interpreter), not what pipx believes it installed — otherwise the uv-shadowing failure mode above stays invisible. `detect_mempalace_python` (`scripts/lib/common.sh:87-101`) is the existing resolution helper to build on.
- A `doctor`-style check in the setup scripts detecting duplicate installs (pipx + uv) and broken pipx symlinks.
- **Multi-CLI parity applies** (Claude Code registers `mempalace-mcp` directly, without the wrapper — the guard needs an equivalent surface there); `docs/cli-matrix.md` must be updated in the same diff per the CLI Matrix Maintenance rule.
- Existing related assets: `scripts/tests/test-mempalace-version-range.sh` (install-time range), spec 0082 (3.6 support precedent).

## Out of scope

- Fixing the reporting machine itself (local ops: remove the uv tool install, repair pipx symlinks, upgrade to 3.6.x — no repo change involved).
- Any change to MemPalace upstream (specs 0068/0070 boundary).

## Discovery context

Surfaced while grounding the knowledge-graph IDEA session (#619 / #620): the level-1 feature depends on 3.6-only tools (`kg_supersede`), making this drift a practical prerequisite fix.

hcross:
The `mempalace>=3.6.0,<3.7` pin was enforced only at framework-setup time, so an out-of-range install served agent sessions a degraded tool surface that agents mistook for the documented one — 3.3.5 served every session on the reporting machine for weeks before a manual exploration noticed. This adds a launch-time guard inside the process that actually serves, plus an operator diagnostic that answers which MemPalace will really answer on a given machine.

Implements spec 0108 and its delta-01, per the plan approved on the issue after two cold-review iterations.

Closes #623

## How to read this PR?

Two commits, and they are meant to be read separately: `97fc69e` is the implementation, `fbd43a7` is an independent coverage audit that changed only test files.

Read the code in this order:

1. **`scripts/lib/mempalace_pin.py`** (new) — the smallest and most load-bearing file. `read_pin()` parses the two bounds from `scripts/lib/common.sh` (line-anchored, exactly one declaration per name or `ValueError`); `check()` decides the range, `packaging`-primary with a stdlib fallback. Zero third-party imports at module scope. One module owns both parser and comparator so the guard and the diagnostic cannot drift apart — which matters because the diagnostic evaluates *foreign* checkouts.
2. **`scripts/lib/mempalace-http-wrapper.py`** — the guard. Note *where* the two phases sit, because the ordering is the design: Phase A is a new Step 1 **before** `import chromadb`, so ADR-0006's monkey-patch ordering is untouched and no package code executes before the range decision; Phase B sits after `from mempalace.mcp_server import main` and before `main()`. `_refuse()` is defined **before** the sibling import — forced, because the `ImportError`-on-`mempalace_pin` branch needs the emitter and cannot import it from the module that failed to import.
3. **`scripts/doctor-mempalace.sh`** (new) — three labelled sections, each stating which question it answers. That labelling is the fix for a blocking review finding: "what a fresh setup would select" and "what a session actually launches" are different questions, and conflating them was the original design's defect.
4. The three test suites, then the CI and matrix wiring.

The comparator deserves a deliberate look. The fallback accepts **only** a closed `release[.postN][+local]` grammar and refuses anything it does not fully consume. That makes "never more permissive than `packaging`" structural rather than enumerated — which is the whole point, because a fallback more permissive than its primary comparator fails in the silent direction, reintroducing the very defect this PR fixes.

## How to test this PR?

```sh
# The three new suites
bash scripts/tests/test-mempalace-pin.sh            # -> Summary: 32 passed, 0 failed
bash scripts/tests/test-mempalace-runtime-guard.sh  # -> Summary: 79 passed, 0 failed
bash scripts/tests/test-mempalace-doctor.sh         # -> Summary: 46 passed, 0 failed

# Same three, under an interpreter carrying BOTH packaging and a real MemPalace 3.6.0 —
# i.e. the interpreter a real session actually serves on. Same counts.
CREWRIG_TEST_PYTHON=~/.local/pipx/venvs/mempalace/bin/python \
  bash scripts/tests/test-mempalace-runtime-guard.sh

# Regression on what the refactor touches, and the wiring gates
bash scripts/tests/test-mempalace-version-range.sh  # -> 6/6
bash scripts/tests/test-chroma-server.sh            # -> 15 passed
bash scripts/check-test-wiring.sh && bash scripts/check-ci-parity.sh
bash scripts/check-no-machine-paths.sh && bash scripts/build-ci.sh --check
task lint-markdown
```

**The diagnostic on a real machine**, which is more convincing than any fixture:

```sh
task mempalace:doctor    # or: bash scripts/doctor-mempalace.sh
```

On the reporting machine this exits 1 and reproduces the spec's *divergent installs* scenario live: PATH `mempalace` resolves to a Homebrew-python install serving **3.0.0** while `mempalace-mcp` resolves into the pipx venv serving **3.6.0**. It also surfaces a case nobody had anticipated — that Homebrew install's `mempalace.__version__` reports **2.0.0** against its own dist-info's 3.0.0 — and handles it exactly as designed: each value is range-checked on its own and neither refuses for disagreeing. All four registrations report `GUARD ABSENT (pre-guard checkout)`, since they name a checkout that predates this change.

**Verify the R5 no-second-copy guard actually guards**, since a hard-coded bound is the regression this design most needs to prevent:

```sh
# Add a bound literal inside any message in the wrapper, then:
bash scripts/tests/test-mempalace-runtime-guard.sh
# -> FAIL: wrapper carries 2 pin-bound literal(s) (R5)
```

**The guard.** Phase A reads the version through the serving interpreter's own in-process resolution and refuses when it falls outside the pinned range. That in-process read *is* the serving version rather than an approximation, because the wrapper does not proxy — it imports `mempalace.mcp_server.main` and calls it in the same process. Phase B range-checks the code's own `__version__`, and refuses **only when that literal is itself out of range** — never on mere disagreement with the dist-info field, which a `.postN` rebuild produces legitimately with both values in range. Refusing disagreement would brick a valid install from a stderr nobody reads, which is the failure shape this ticket exists to eliminate. The disagreement is reported by the diagnostic instead.

All five refusal paths route through one `_refuse()` emitter, so the four mandated fields and the restart notice cannot be omitted. Diagnostics go to stderr only: stdout is the JSON-RPC channel, and an in-range launch is asserted to leave stdout **exactly** what it was before, not merely non-empty.

**Why the fallback grammar is a whitelist.** A blacklist — refuse on a pre-release or dev marker, otherwise ignore the remainder — accepts `3.6.0-git`, `3.6.0-nightly`, `3.6.0.unknown`, `3.6.0_1`, `3.6.0~1`, `3.6.0+` and `3.6.0:1`, every one of which `packaging` rejects with `InvalidVersion`. It is reachable in exactly the unrecognised-installer scenario the spec requires support for, because `importlib.metadata.version()` returns the METADATA `Version:` field verbatim and unnormalised. The whitelist was attacked during the audit with 999 targeted adversarial strings (all 29 codepoints `str.isspace()` accepts, the full grammar space with leading zeros and 40-digit post numbers, Unicode digits in every position) and 600k randomized differential strings across two bound pairs: zero cases where the fallback accepts what `packaging` refuses.

**Setup-time enforcement is deliberately untouched.** `mempalace_version_in_range` still refuses outright when `packaging` is absent — verified: the diff mentions that function zero times. The asymmetry is intentional and tested: a working machine keeps working, a fresh setup fails loudly with a human watching. Aligning setup onto the shared comparator is a clean follow-up, not this diff.

**The `common.sh` refactor is behaviour-preserving.** `mempalace_python_candidates()` is extracted so the diagnostic can report *which* candidate won without duplicating the ordering. The `< <(...)` process-substitution form matters: a pipe would run the loop body in a subshell and silently break `return 0` for all 16 call sites across 11 files. `test-chroma-server.sh` is in the verification gate for that reason.

**Two pre-existing `docs/cli-matrix.md` defects fixed in row 7d**, since the row had to be touched anyway: the Claude cell omitted the `bash <repo>/scripts/lib/tls-exec.sh` prefix that all four setup scripts write, and the Antigravity cell named `~/.gemini/antigravity-cli/settings.json` where the real registration is `~/.gemini/config/mcp_config.json` — the named file carries no `mempalace` key at all.

**Deliberately out of scope**, each with a reason: remediating any particular machine's install (local operations, no repository change); any change to MemPalace upstream (the specs 0068/0070 boundary); reporting the version an already-running session serves (a launch-time guard plus a documented restart step closes the same window — see the spec's *Rejected alternatives*); any override letting an operator start on an out-of-range install (a runtime guard laxer than the install-time one would reopen the exact window this closes); and the duplicated `7g` row label in `docs/cli-matrix.md`, recorded in the plan's Risks so the next author does not assume `7g` is free.

**Follow-up observed during the audit, pre-existing and not a spec violation:** `mempalace_python_candidates` extracts a shebang with `sed 's|^#!\([^ ]*\).*|\1|p'`, so a console script carrying `#!/usr/bin/env python3` yields `/usr/bin/env` as candidate 2 — a candidate that can never import `mempalace.mcp_server`. Pre-existing in `detect_mempalace_python`; the diagnostic is right to report the list verbatim and resolves the real interpreter correctly in its own section. Worth its own ticket.


hcross:
## Verdict: REQUEST CHANGES

**Two blocking findings, both in the test suites — the implementation I could not break.**

The sharper one: I can hardcode the primary comparator's floor to `0` — making the guard accept
*any* MemPalace below the floor, which is literally the #623 defect this PR exists to close —
and **all 157 assertions across all three suites still pass on the interpreter CI actually
uses**. The guard's most load-bearing code path has no regression protection under CI.

Everything else — spec conformance across all 12 requirements, the comparator invariant, the
five refusal paths, R6's stdout guarantee, the CI wiring, the conventions — I attacked and it
held. Both blocking fixes are test-only and mechanical.

Cold review, no authoring context. Every number, path and behavioural claim below was
re-derived locally against `fbd43a7` (verified `==` the PR head sha) with base `5c089af`;
nothing is carried over from the PR body or the spec. Where I checked a PR-body claim and it
held, I say so — those were the places most worth attacking.

*Formal review event unavailable: `gh` authenticates here as `hcross`, this PR's author, and
GitHub structurally refuses `--approve` / `--request-changes` on one's own PR under a
single-identity setup (issue #692). This plain comment carries the binding verdict.*

---

### 1. CI status

All **12** required checks pass on `fbd43a7`. `check-components` was `pending` when this review
opened and has since completed green; nothing is failing or outstanding.

`build` · `check-agents-size` · `gitlab-ci-check` · `grep-anti-patterns` · `lint-specs` ·
`check-components` · `check-ci-parity` · `check-extension-version-bump` ·
`check-feedback-routing` · `check-skill-versions` · `lint-markdown` · `test-harness-curate`

Re-run locally, all green: `check-test-wiring.sh` (59 wired, 2 exempted), `check-ci-parity.sh`,
`check-no-machine-paths.sh`, `build-ci.sh --check`, `check-skill-versions.sh`,
`build-components.sh --check`.

Green CI is exactly the problem in finding **B1**: it stays green with the guard broken.

---

### 2. Correctness — the implementation

**The comparator invariant holds.** This was the claim most able to block, so I attacked it
rather than read it. Under `~/.local/pipx/venvs/mempalace/bin/python` (packaging 26.2, CPython
3.13.14) I ran `_fallback_in_range` against `packaging` differentially over **five** bound
pairs: 1 266 hand-built adversarial strings, 400 000 random strings over a version-shaped
alphabet, and 300 000 grammar-shaped strings of which **211 231 were actually admitted** by the
whitelist. **Zero** cases where the fallback accepts what `packaging` refuses; zero exceptions.

The two attacks that could plausibly have worked:

- *The `strip()` / `\s*` asymmetry.* `_fallback_in_range` does `found.strip()`
  (`mempalace_pin.py:135`) while `check()` hands `found` to `Version()` unstripped (`:127`),
  which tolerates only `\s*`. A codepoint `str.strip()` removes but `\s` does not match would
  let the fallback accept a string `packaging` never sees intact. Measured: the two sets are
  **identical** — 29 codepoints, and both set differences are empty. No gap.
- *The whitelist as a structural subset.* `[0-9]+(\.[0-9]+)*`, `\.post[0-9]+` and
  `\+[a-zA-Z0-9]+(\.[a-zA-Z0-9]+)*` each map onto a strictly narrower slice of PEP 440's own
  `VERSION_PATTERN` (which additionally allows `[-_.]?` separators, `rev`/`r` spellings and bare
  `-N`), so every admitted string is a valid PEP 440 version by construction. Release-tuple-only
  ordering is safe because post and local segments sort a version only *upward*, and both bounds
  are refused unless they are plain releases (`:139-142`). "Structural rather than enumerated"
  is accurate.

**All five refusal paths exist and all five emit.** I counted `_refuse` call sites myself —
`mempalace-http-wrapper.py:153, 165, 178, 189, 264` — and exercised **four directly** plus
Phase B end-to-end. Every one: exit **1**, stdout **0 bytes**, R4 restart notice present.

| path | trigger exercised | exit | stdout | fields |
|---|---|---|---|---|
| `:153` guard module unimportable | copied wrapper without `mempalace_pin.py` | 1 | 0 B | 5 + Cause + NOTE |
| `:165` pin unreadable | appended a 2nd `MEMPALACE_MIN_VERSION=` to a fixture `common.sh` | 1 | 0 B | 5 + Cause + NOTE |
| `:178` no dist metadata | bare interpreter, no fixture | 1 | 0 B | 5 + Cause + NOTE |
| `:189` Phase A out of range | dist-info `3.3.5` under the real pipx interpreter | 1 | 0 B | 5 + NOTE, `comparator=packaging` |
| `:264` Phase B out of range | dist-info `3.6.0` + `__version__` `3.3.5` / `3.7.0` | 1 | 0 B | `source: mempalace.__version__` |

**Phase ordering and stream safety survive attack.** Phase A precedes `import chromadb`, and
`importlib.metadata.version()` reads dist-info off disk without importing the package, so
ADR-0006's patch ordering is genuinely untouched — confirmed behaviourally, not by reading: an
in-range real install reaches Step 3 and dies on the *daemon* error, only possible if `chromadb`
was imported after the guard passed. Phase B is reachable: importing `mempalace.mcp_server`
populates `sys.modules["mempalace"]`, and I exercised its refusal.

The only writes on the guard path are two `print(..., file=sys.stderr)` calls (`:137`, `:240`);
every `print` in `mempalace_pin.py` sits after the `__main__` guard at `:161` (verified: no
`print(` or `sys.argv` occurrence before it). I also independently confirmed the upstream claim
the Phase B comment leans on: `mempalace/mcp_server.py:44` really does `sys.stdout = sys.stderr`
at import time and never reassigns `sys.stderr`. One correction: the restore at `:4909` lives in
`_run_stdio_loop()`, not in `main()` as the comment says — downstream of `main()`, so the
conclusion is unaffected, and `file=sys.stderr` makes the guarantee unconditional either way.

**R6 verified at the byte level, not as "non-empty".** With a stub `main()` writing one sentinel
line, an in-range launch produced stdout containing **exactly** the sentinel and stderr
**empty** — the guard contributes nothing to either channel. The design's load-bearing case also
holds: dist-info `3.6.0.post1` disagreeing with an in-range `__version__` `3.6.0` serves with an
empty stderr. Refusing that disagreement would indeed have bricked a valid rebuild.

**No second copy of either bound on the launch path.** Repo-wide grep for `3.6.0` / `<3.7`: every
hit outside `specs/`, `docs/` and `scripts/tests/` is pre-existing (`Taskfile.yml:29,33` setup
task, `README.md:470-472`, the harness-curator skill, `docker/e2e/README.md`) and none is on the
launch path. The two live launch-path shell files, `tls-exec.sh` and `tls-delegation.sh`, carry
**zero** occurrences. `common.sh:15-16` is the sole declaration of each name and the parser
round-trips it (`--print-pin` → `min=3.6.0 max=3.7`); its four other occurrences are `${...}`
interpolations, not copies.

**The diagnostic reproduces the spec's divergent-installs scenario live**, exit **1**, on this
machine — and every PR-body claim about that run is accurate: four registrations
`GUARD ABSENT (pre-guard checkout)`, `PATH mempalace` → 3.0.0 out of range under
`/opt/homebrew/opt/python@3.11/bin/python3.11`, `PATH mempalace-mcp` → 3.6.0, and the
unanticipated `mempalace.__version__: 2.0.0 (DISAGREES with dist-info 3.0.0)` reported rather
than refused.

**Spec compliance, requirement by requirement:** R1 ✅ · R2 ✅ (in-process resolution, per
delta-01's replacement wording) · R3 ✅ · R4 ✅ (all five refusals plus the doctor's
unconditional closing note at `doctor-mempalace.sh:454`) · R5 ✅ · R6 ✅ · R7 ✅ · R8 ✅ · R9 ✅
(see finding 12) · R10 ✅ (`doctor-mempalace.sh:406-412`) · R11 ✅ (all four registrations route
through the one wrapper — Claude `setup-claude-interactive.sh:220`, Gemini
`config/gemini/settings.json:35`, Copilot `config/copilot/mcp-config.json.template:7`,
Antigravity `setup-antigravity-interactive.sh:219`; no CLI registers `mempalace-mcp` directly,
and all four setups write the `tls-exec.sh` prefix) · R12 ✅.

The `common.sh` refactor is behaviour-preserving as claimed: `< <(...)` does keep `return 0` in
the helper's own shell, and `test-chroma-server.sh` (15 passed) plus
`test-mempalace-version-range.sh` (6/6) both stay green.

---

### 3. Convention compliance

- **Gitmoji titles** — PR `✨ …`, commits `✨ …` / `✅ …`; no conventional-commit prefix. ✅
- **PR body** — all three mandated sections present. ✅
- **Branch + worktree** — `feat/0623-mempalace-runtime-version-guard`, `.worktrees/0623`. ✅
- **Close keyword** — `closingIssuesReferences` resolves to `[623]` and nothing else. ✅
- **Logbook (Rules A/B)** — #623 is the feature issue and is its own logbook: 10 comments, 9
  dated across this working session, so the journal is incremental rather than batched. ✅
- **Version bump** — no skill/agent source modified; `check-skill-versions.sh` agrees. ✅
- **CLI matrix** — updated in the same diff. ✅
- **Core-paths manifest** — `docs/layers.md` untouched and `.crewrig/core-paths.txt:51` registers
  `scripts` as a directory, so the five new files need no entry; `check-core-paths.sh` passes. ✅
- **Two-commit separation** — verified: `fbd43a7` touches
  `scripts/tests/test-mempalace-{doctor,pin,runtime-guard}.sh` and nothing else. ✅

**CI wiring surfaces — count established independently rather than taken on trust.** There are
**three**, and all three are updated:

1. `.github/workflows/build.yml` — `check-test-wiring.sh` scans `.github/workflows/*.yml|*.yaml`
   plus `ci/ci-capabilities.yml`; `check-ci-parity.sh` arm 1 additionally demands each portable
   capability's `command:` be exhibited by its attributed GHA job.
2. `ci/ci-capabilities.yml` — source of truth for the portable set.
3. `.gitlab-ci.yml` — generated from (2), drift-gated by `build-ci.sh --check`, which
   `check-ci-parity.sh` composes.

`Taskfile.yml` carries no test aggregate (`grep -n 'scripts/tests/' Taskfile.yml` → empty), so
it is not a fourth surface. Nothing is missing.

---

### 4. Test coverage

Counts measured from the `Summary:` lines: **32 / 79 / 46**, plus
`test-mempalace-version-range.sh` 6/6 and `test-chroma-server.sh` 15 passed. Re-run with
`CREWRIG_TEST_PYTHON=~/.local/pipx/venvs/mempalace/bin/python` — the interpreter a real session
actually serves on, carrying both `packaging` and a real MemPalace 3.6.0 — the counts are
identical: **32 / 79 / 46**, 0 failed. The interpreter-independence claim holds.

**What is genuinely load-bearing.** I mutation-proved a representative set rather than trusting
the counts. Each of these broke the intended assertions and nothing else: R5 bound literal
reintroduced (`FAIL: wrapper carries 2 pin-bound literal(s) (R5)`, 78/1), Phase A range check
removed (10 fails), R4 restart note removed (8 fails), Phase B removed (9 fails), fallback
grammar widened to `[^0-9]*` (6 fails including `Arm B: the fallback is more permissive than
packaging`), doctor never exits non-zero (3 fails), doctor pin-divergence removed (1 fail).

**What the fixtures get right.** Two traps I specifically hunted are correctly avoided, with
proof rather than assertion:

- *Shadowing vs absence, MemPalace metadata.* A `PYTHONPATH` fixture can shadow an installed
  package but cannot un-resolve one. `test-mempalace-runtime-guard.sh:398-400` runs the absence
  case with `-S`, and `:196-198` names the exact failure mode. Load-bearing: drop the `-S` and
  run under the interpreter that has a real MemPalace 3.6.0 and the case fails
  (`no-dist-info: exited 0 — a version-less install began serving`).
- *The tautology in the corpus driver.* `test-mempalace-pin.sh:322-323` records why the
  comparator label must be printed from what was **observed**, and `:418-427` adds a positive
  shield-effectiveness proof — the shield must flip `packaging` → `stdlib-whitelist` on an
  interpreter that has `packaging`, so a no-op shield fails loudly.

A hazard the fixtures also correctly avoid: my own first Phase B probe reused one tree across
cases and got a false refusal, because three same-length `__init__.py` writes inside one second
share a `.pyc`. `build_fixture`'s comment at `:112-115` explains why the case name is an
argument rather than a counter — a fresh tree per case — which is exactly that guard.

**What is not covered is in the findings below.** Eight of the assertions I probed are satisfied
by something other than what their label claims, and two of those rise to blocking.

---

### 5. Linter findings

`lint-shell.sh scripts/doctor-mempalace.sh` — exit 1, one finding, informational only:

```
In scripts/doctor-mempalace.sh line 45:
. "${SCRIPT_DIR}/lib/common.sh"
  ^---------------------------^ SC1091 (info): Not following: lib/common.sh was not specified as input (see shellcheck -x).
```

An artifact of not passing `shellcheck -x`, not a defect. No SC2xxx warning, no missing-`set`
note (the omission of `set -e` is deliberate and documented at `:39-41`). Findings against
`scripts/lib/common.sh` (`SC2015` ×2, `SC2088`, mode 100644, missing `set -e`) are pre-existing
and outside this diff's lines.

`lint-python.sh` — degraded (`neither ruff nor flake8 found`), plus its `bare print(` heuristic:

```
scripts/lib/mempalace_pin.py:221: bare print() in non-test file
scripts/lib/mempalace_pin.py:222: bare print() in non-test file
scripts/lib/mempalace_pin.py:223: bare print() in non-test file
scripts/lib/mempalace_pin.py:224: bare print() in non-test file
scripts/lib/mempalace_pin.py:228: bare print() in non-test file
scripts/lib/mempalace_pin.py:237: bare print() in non-test file
scripts/lib/mempalace_pin.py:241: bare print() in non-test file
scripts/lib/mempalace_pin.py:245: bare print() in non-test file
scripts/lib/mempalace-http-wrapper.py:137: bare print() in non-test file
scripts/lib/mempalace-http-wrapper.py:240: bare print() in non-test file
```

All ten are false positives, verified rather than assumed: the eight in `mempalace_pin.py` are at
lines > 161, inside the `if __name__ == "__main__":` block the module docstring commits to; the
two in the wrapper both pass `file=sys.stderr`. The heuristic understands neither.

`lint-markdown.sh docs/cli-matrix.md` — clean, exit 0. `lint-json.sh` — no JSON in the diff.
`lint-skill.sh` — no `SKILL.md` in the diff.

---

## Findings

Two blocking, eleven not. **Every blocking finding is about regression protection, not about
shipped behaviour** — the implementation is correct today, and both fixes are test-only.

### Blocking

**1 — `class: tech`, BLOCKS.** The primary `packaging` comparator's range logic is asserted by
nothing that runs in CI. Mutate `mempalace_pin.py:127` to
`verdict = Version("0") <= Version(found) < Version(maximum)` — the guard now accepts *any*
below-floor MemPalace, the exact #623 defect class — and:

```
check('3.3.5','3.6.0','3.7') = (True, 'packaging')      # mutant accepts it
check('0.0.1','3.6.0','3.7') = (True, 'packaging')

pin            Summary: 32 passed, 0 failed             # default interpreter = what CI runs
runtime-guard  Summary: 79 passed, 0 failed
doctor         Summary: 46 passed, 0 failed
```

**157/157 green with the guard broken.** And this is not merely "CI's `python3` lacks
`packaging`": the pin suite — the suite that *owns* `check()` — still reports `32 passed, 0
failed` when forced onto an interpreter that **has** `packaging`
(`CREWRIG_TEST_PYTHON=~/.local/pipx/venvs/mempalace/bin/python`). Mechanism: Arm B
(`test-mempalace-pin.sh:350-377`) compares `_fallback_in_range` against the driver's **own
inline** `Version(MIN) <= Version(text) < Version(MAX)` expression; `check()` is called once at
`:372` as `_, comparator = mempalace_pin.check(MIN, MIN, MAX)`, and the **verdict is thrown
away** — only the label is read. The runtime-guard suite does catch the mutation (20 failures),
but only under an interpreter carrying `packaging`, and `CREWRIG_TEST_PYTHON` is passed
**nowhere** in `.github/workflows/`, `ci/ci-capabilities.yml` or `.gitlab-ci.yml` (verified).

Fix: assert `check()`'s own verdict on the `packaging` branch — a handful of boundary rows
(`floor-ε`, `floor`, `ceiling-ε`, `ceiling`) run through `check()` and compared to an expected
boolean, not to a re-derived expression. And either install `packaging` on the CI runner or make
the suite fail loudly when the `packaging` arm cannot run at all (see finding 9), so "green"
means the primary comparator was actually exercised.

**2 — `class: tech`, BLOCKS.** R8's version-divergence trigger — the spec's headline *Divergent
installs are reported and flagged* scenario, and the very reason the diagnostic exists — is
entirely unasserted, and the assertion labelled as covering it is answered by a different
mechanism. Three measurements:

```
delete the version-divergence block (doctor:425 -> `if false; then`)   -> 46 passed, 0 failed
rename only its message ("reported versions DIVERGE" -> gibberish)     -> 46 passed, 0 failed
control: delete the PIN-divergence block  (doctor:436)                 -> 45 passed, 1 failed
```

`test-mempalace-doctor.sh:292` is `has "$S1_OUT" "DIVERGE" "R8: reports the divergence
explicitly"` — a bare substring grep. Scenario 1's fixture fires **three** triggers at once
(out-of-range, version divergence, pin divergence), so the grep is satisfied by the
*pin*-divergence line and the non-zero exit by the range failure. Deleting detection of "two
sources report different MemPalace versions" — precisely the #623 symptom, `pipx list`
disagreeing with the serving interpreter — costs zero assertions.

This is the exact class of gap `fbd43a7` was written to close. Its own commit message: *"R8's
range trigger was never tested alone. The divergent-machine scenario fires divergence AND
out-of-range together, so dropping the range check still passed it."* The audit isolated the
range trigger and added case 5 for it; the version-divergence trigger needs the mirror-image
case — every source in range, two versions differing, no pin divergence — plus a grep tightened
from `DIVERGE` to `reported versions DIVERGE`.

### Non-blocking — test coverage (`class: tech`)

**3.** R3's *Supported range* field value is uncoverable. `EXPECTED_RANGE=">=3.6.0,<3.7"` is a
substring of `EXPECTED_REMEDY="mempalace>=3.6.0,<3.7"` (`test-mempalace-runtime-guard.sh:95-96`),
so the two greps collapse onto one output line. Redacting the field value in `_refuse` → **79
passed, 0 failed**. Control: dropping the *remedy* field is caught (6 failures).

**4.** R3's *Resolved interpreter* is a shape-only grep
(`grep -qE '^  Resolved interpreter: +[^ ]+'`). Making the wrapper report
`/nonexistent/not-the-interpreter` → **79 passed, 0 failed**. Reporting the wrong interpreter is
one of the things spec 0108 exists to make impossible, and the suite already knows the expected
value (`$PYTHON_BIN`).

**5.** The above-ceiling *names the version found* assertion is tautological:
`ABOVE_CEILING="$PIN_MAX"` = `3.7` (`:94`), which also appears in the `Supported range` and
remedy lines of every refusal. Under a mutation that redacts the version-found value, five of six
sibling assertions fail and this one still reports `ok`.

**6.** The doctor's per-source report fields are uncovered. `GOOD_VERSION="$PIN_MIN"` = `3.6.0`
(`test-mempalace-doctor.sh:95`), and `doctor-mempalace.sh:194` unconditionally echoes
`supported range: >=3.6.0,<3.7` — so the "reports the in-range version" grep cannot fail while
the pin is readable. Redacting `version served:`, or deleting the `verdict:` or
`packaging importable:` fields outright, each leaves **46 passed, 0 failed**.

**7.** The `any` comparator branch accepts the non-answer. `grep -qE '^  Comparator: +[^ ]+'`
matches the literal `not reached`, so the four cases that pass `any` assert only that a
`Comparator:` line exists. A wrong comparator *name* is caught by only the two cases that pass
an explicit label (2 failures).

**8.** Cases 1, 2 and 9a do not isolate Phase A: `build_fixture below-floor "$BELOW_FLOOR"
"$BELOW_FLOOR"` puts the same out-of-range value in `mempalace.__version__`, so Phase B alone
satisfies them. Deleting Phase A leaves all three passing — the suite is saved by case 9c and
case 10, whose comment at `:520-522` shows the author found this deliberately. Not dishonest,
but the three headline cases carry less weight than their labels suggest.

**9.** Three `ok` lines are skips. On a machine where no interpreter can import `packaging`,
`test-mempalace-pin.sh:431/450/467` score the skipped shield proof, skipped Arm B and skipped
`InvalidVersion` branch as passes, and the total stays **32** either way — so `32 passed` does
not tell a reader whether the differential ran. The skips *are* loudly narrated on stdout, which
is honest; the flaw is that the count is invariant. This is what makes finding 1 total rather
than partial.

**10.** The four bound-literal greps (`test-mempalace-pin.sh:154`, `:539`,
`test-mempalace-runtime-guard.sh:563`, `test-mempalace-doctor.sh:560`) use
`if grep …; then bad; else ok; fi`, so a `grep` exit-2 (missing file, unavailable binary) is
scored as a pass. Low severity, one-line fix (`grep …; rc=$?; case $rc in 0) bad;; 1) ok;; *)
bad "grep failed";; esac`).

### Non-blocking — implementation (`class: tech`)

**11.** `mempalace_pin.check()` can raise instead of returning a verdict. `_release_tuple`
(`:154`) calls `int()` on a release segment, and a segment longer than CPython's 4 300-digit
limit raises `ValueError`. Measured with `found = "3.6." + "1"*4301`: `_FALLBACK_RE` admits it
and `_fallback_in_range` raises. Not fallback-only — `packaging.version.Version()` raises the
identical `ValueError` from its own `tuple(map(int, ...))` at `version.py:416`, and only
`InvalidVersion` is caught (`:128`), so `check()` is non-total on **both** paths and the wrapper
dies with a traceback instead of the R3/R4 field set. R1's refusal survives (Python still exits
non-zero); R3's naming does not. Unreachable from any plausible `METADATA` value, which is why
it is a nit — but the docstring's "an unparseable version is an unmet precondition, and the guard
fails closed on it" is stronger than the code.

**12.** `sys.path.insert(0, _LIB_DIR)` (`mempalace-http-wrapper.py:144-145`) is never undone, so
`scripts/lib/` stays on `sys.path` for the whole serving session. No collision today — measured,
the only importable module names there are `en_us_sweep` and `mempalace_pin`, and their
intersection with `sys.stdlib_module_names` is empty — but any future `.py` dropped into
`scripts/lib/` silently shadows a same-named module for MemPalace and chromadb, in a file whose
whole premise is not perturbing the serving process. `sys.path.remove(_LIB_DIR)` after the import
costs one line and is safe: `mempalace_pin` is in `sys.modules` by then.

**13.** `doctor-mempalace.sh:288-294` derives its `WRAPPER MISSING` verdict from whether
`dirname "$wrapper"/../..` can be `cd`'d into, never from `[ -f "$wrapper" ]`. A registration
naming a checkout whose `scripts/lib/` directory survives but whose wrapper file was deleted is
reported as present — one of the breakage shapes this section exists to surface.

**14.** The new row 7d cites "row 7g" for the `tls-exec.sh` provenance, but `7g` is a
**duplicated** label: `grep -oE '^\| 7[a-z]? ' docs/cli-matrix.md | sort | uniq -c` gives 7g ×2
— line 39 (layered wake-up) and line 40 (spec 0084 TLS delegation) — both pre-existing on
`5c089af`. Leaving the duplicate out of scope is defensible and the PR says so; adding a
cross-reference *to* the ambiguous label is what makes it newly load-bearing. Cite "spec 0084"
alone, or renumber.

**15.** Cosmetic: row 7d's Claude cell renders the last argv element as
`scripts/lib/mempalace-http-wrapper.py` while `setup-claude-interactive.sh:220` builds
`$REPO_DIR/scripts/lib/mempalace-http-wrapper.py`. The `tls-exec.sh` element in the same cell
*is* shown with `<repo>/`, so the inconsistency reads as a claim about relative paths.

**16.** `_refuse`'s `bounds is None` remedy is "re-run the framework setup from a complete
checkout … so the pin and the guard module are both present" (`:98-102`). Correct for the `:153`
ImportError path; for the `:165` malformed-`common.sh` path re-running setup fixes nothing —
verified by injecting a duplicate `MEMPALACE_MIN_VERSION=` and reaching that branch. The
`Cause:` line names the real problem exactly, so the operator is not stranded; the remedy line
points the wrong way.

### Non-blocking — the specification itself (`class: spec`)

These route to SPECS, not to this diff. The implementation resolved each one sensibly; the point
is that it had to *infer* the resolution.

**17.** R7 asks the diagnostic to report "the interpreter and version the memory-server launch
path would itself select". The launch path selects nothing at launch — it runs the interpreter
its registration recorded; the selector (`detect_mempalace_python`) is a *setup*-time actor. The
implementation resolves this correctly by splitting sections 1 and 3 under explicit labels, but a
literal reading of R7 is unsatisfiable and the next author must re-derive the same inference.

**18.** R9 forbids the diagnostic from depending "on an enumeration of known Python package
managers", while R10 requires it to report the framework's own candidate ordering — whose
candidate 1 is `$HOME/.local/pipx/venvs/mempalace/bin/python` (`common.sh:490`), a pipx-specific
path. Sections 1 and 2 are enumeration-free so the intent is met, but R9 and R10 are in literal
tension and the implementation had to pick a reading.

**19.** R3 mandates four fields for "a refusal per requirement 1". Two of the five refusal paths
— guard module unimportable, pin unreadable — cannot name the range and honestly print
`not determined (see the cause below)`. Defensible, since neither is an R1 refusal, but the spec
never contemplates a refusal that *isn't* an R1 refusal, so the compliance argument rests on a
reading rather than on text.

---

### Adjacent, pre-existing — not findings against this diff

- `setup-claude-interactive.sh:248` — the skip-branch hint prints
  `claude mcp add --scope user mempalace -- $MEMPALACE_PYTHON_BIN $MEMPALACE_WRAPPER`, omitting
  the `tls-exec.sh` prefix both real registration branches (`:231`, `:243`) use. An operator
  following it lands a spec-0108-guarded but un-TLS-wrapped entry, which the doctor reports as
  guarded while spec 0084's parity claim quietly fails for that install.
- The `mempalace_python_candidates` shebang extraction yields `/usr/bin/env` as candidate 2 for
  an `#!/usr/bin/env python3` console script. The PR body already flags this as pre-existing in
  `detect_mempalace_python` and worth its own ticket; I concur, and the diagnostic is right to
  print the list verbatim.
- `packaging` is **not** a declared requirement of either `chromadb` or `mempalace` (measured via
  `importlib.metadata.requires`), so the degraded comparator is genuinely reachable in a serving
  process rather than dead code. That is why I fuzzed it — and it is also why finding 1 matters:
  on a serving interpreter that *does* carry `packaging`, the untested branch is the one in force.

---

Two commits assessed separately, as intended. `97fc69e` is the implementation and stands up to
attack. `fbd43a7` changes only the three test files and closes real gaps — each of which I
confirmed *was* a gap: the R5 check is now substring-based across all three sources
(mutation-proven), the absence case now genuinely establishes absence via `-S` rather than
shadowing, and the shield now proves its own effectiveness. That work is the opposite of coverage
theater, which is exactly why findings 1 and 2 are worth one more pass: the same method applied
to the primary comparator and to the version-divergence trigger closes the last two places where
a green suite would not notice the guard failing.

Merge stays gated on the user, per `AGENTS.md` → *Branching Strategy*.

hcross:
## Verdict: APPROVE

**REVIEW iteration 2 — cold reviewer, no authoring context.** Both blocking findings from
iteration 1 are genuinely discharged, and I proved it by breaking the code rather than by reading
the commit message. All fourteen non-blocking actionable findings are discharged too, each one
mutation-proven. I found **one new non-blocking coverage gap** and **one new `spec`-class
finding**; neither blocks.

**Two corrections to the premise I was handed.** I was briefed that the head commit is `7ca9d16`.
It is not — the head is **`9671ded`**, a fourth commit that landed 3 minutes after `7ca9d16`.
That matters, because `7ca9d16` **failed a required check** (`grep-anti-patterns`, run
[30842081906](https://github.com/crewrig/crewrig/actions/runs/30842081906)): it introduced
`except Exception as exc:  # noqa: BLE001` at `test-mempalace-pin.sh:468`, and Rule 1 of
`docs/scripting-conventions.md` recognizes only the `acknowledged-exception:` marker — a `noqa`
code from a different toolchain does not satisfy the workflow's grep filter. Had `7ca9d16` been
the head, this review would have been `REQUEST CHANGES` on CI alone. `9671ded` fixes it correctly
— it swaps the marker and **keeps** the broad `except`, which is right: the assertion's whole
purpose is that `check()` raises nothing, so narrowing the clause would let an unexpected
exception type pass as a success. I re-ran the workflow's grep verbatim against the head tree:
`Rule 1 clean.`

Every number below was measured by me against `9671ded`, on two real interpreters plus one
synthesised third. Nothing is carried from the PR body, the commit messages, or the iteration-1
verdict — where I checked one of their claims and it held, I say so, because those were the
claims most worth attacking.

*Formal review event unavailable: `gh` authenticates here as `hcross`, this PR's author, and
GitHub refuses `--approve` / `--request-changes` on one's own PR. Issue
[#692](https://github.com/crewrig/crewrig/issues/692) tracks the limitation. This plain comment
carries the binding verdict.*

---

### 1. CI status

All **12** required checks **pass** on `9671ded` (runs
[30842297565](https://github.com/crewrig/crewrig/actions/runs/30842297565) and
[30842297648](https://github.com/crewrig/crewrig/actions/runs/30842297648)):

`build` · `check-agents-size` · `check-ci-parity` · `check-components` (2m46s) ·
`check-extension-version-bump` · `check-feedback-routing` · `check-skill-versions` ·
`gitlab-ci-check` · `grep-anti-patterns` · `lint-markdown` · `lint-specs` · `test-harness-curate`

Re-run locally against the head tree, all exit 0: `check-skill-versions.sh`,
`check-test-wiring.sh` (59 wired, 2 exempted), `check-ci-parity.sh`, `build-ci.sh --check`,
`check-core-paths.sh`, `check-no-machine-paths.sh`.

Nothing failing, nothing pending. The `grep-anti-patterns` failure described above belongs to the
superseded commit `7ca9d16`, not to the head.

---

### 2. Correctness — is the primary comparator now actually protected?

This was the iteration-1 blocker and the claim I attacked hardest. **It holds, in its strongest
form.**

Baseline at `9671ded`, default `python3` (Homebrew, **no `packaging`** — measured:
`ModuleNotFoundError`): **36 / 85 / 63 = 184 assertions**, 0 failed. At `fbd43a7` the same three
suites give **32 / 79 / 46 = 157**. The commit message's `157 -> 184` is exact.

I then applied the exact iteration-1 mutation — hardcode the primary branch's floor,
`mempalace_pin.py:132` → `Version("0") <= Version(found) < Version(maximum)`, which accepts *any*
below-floor MemPalace, the #623 defect class — on three interpreter configurations:

| interpreter configuration | iteration 1 | at `9671ded` |
|---|---|---|
| default `python3`, **no `packaging`** (what CI ran) | `32/0` · **green** | **`34/2`** |
| pipx venv, **real `packaging` 26.2** | `157/0` · **green** | **`34/2`, `65/20`, `60/3`** |
| synthesised: venv with no `packaging` **and** no MemPalace, empty `HOME`, stripped `PATH` — nothing on the machine can import `packaging` | *(n/a)* | **`31/2`** |

The third row is the one that decides it. On a machine where the real library is unreachable
*anywhere*, the mutation still costs an assertion — the provider arm catches it — and the
missing real-library arm is a *counted failure*, not a skip:

```
far below the floor          0         -> verdict=True comparator=packaging  WRONG VERDICT
immediately below the floor  3.5.9999  -> verdict=True comparator=packaging  WRONG VERDICT
FAIL: packaging branch (provider): check() returned a wrong verdict, a wrong comparator, ...
FAIL: no interpreter here can import the real `packaging`, so the PRIMARY comparator was never
      exercised against the real library
```

**The `packaging` provider is not an oracle. I tried to make it one and could not.** This was the
sharpest attack available, because a stand-in that silently certified acceptance semantics would
be worse than the gap it closed. Three independent reasons it holds:

- **It appears in exactly two places** (`test-mempalace-pin.sh:515`, `:762`) — measured; it is
  absent from the section-5 corpus differential and from both other suites (`grep -rn PROVIDER`
  → no hits in the runtime-guard or doctor suites). Arm B still runs against the real library
  only.
- **In section 4a the expectation never consults it.** Verdicts come from a hardcoded `expected`
  boolean cross-checked by `in_range_by_tuple`, which uses only integer tuples built from `argv`.
  A *wrong* provider therefore causes a test **failure**, not a false pass — the safe direction.
- **In section 6, where the provider's validity semantics do bite, they are cross-checked.** The
  provider is far stricter than the real library — measured, they disagree on 4 of 5 probes
  (`3.6.0.post1`, `3.6.0+local`, `3.6.0rc1`, `3.6.0-1` are all VALID to `packaging` 26.2 and
  INVALID to the provider). The assertion uses `${PIN_MIN}-git`, the **one** string both refuse,
  and the very next assertion re-runs it against the real library. Had the author picked any of
  the other four, the real-library arm would fail.

And there is no green state in which the provider is the sole certifier: if `PACKAGING_PYTHON` is
unresolvable, section 4c emits a counted `bad`. Verified above.

**The implementation changes in `7ca9d16` are sound.** `check()` is now total on both branches:
`except ValueError` (`mempalace_pin.py:133`) plus the mirror guard in `_fallback_in_range`
(`:161`). The broadening is safe in both directions — I checked the one way it could have been a
regression: `InvalidVersion.__mro__` is
`['InvalidVersion', 'ValueError', 'Exception', 'BaseException', 'object']` in `packaging` 26.2, so
the narrower case it replaced is still caught. `sys.path` is undone in a `finally`
(`mempalace-http-wrapper.py:177-179`) and only when this file added it. `WRAPPER MISSING` is now
keyed off `[ ! -f "$wrapper" ]` (`doctor-mempalace.sh:292`) rather than its grandparent directory.

**R7/R8 live.** `bash scripts/doctor-mempalace.sh` on this machine exits **1** and reproduces the
spec's divergent-installs scenario for real: seven sources tabulated, `PATH:mempalace` → 3.0.0
under `/opt/homebrew/opt/python@3.11/bin/python3.11`, both findings named
(`3.0.0 lies outside >=3.6.0,<3.7`, `reported versions DIVERGE — 3.0.0 3.6.0`), R4's restart note
present.

**The declared-unreachable branch: I agree with the call.** `CHECKOUT UNRESOLVABLE`
(`doctor-mempalace.sh:307-313`) really is unreachable while the `-f` test precedes it, and the
argument is airtight rather than merely plausible: `-f "$wrapper"` succeeding requires `+x` on
every ancestor directory of the wrapper, and the `cd` target *is* an ancestor, so it cannot fail.
Bash's logical `..` resolution makes the canonicalised target a textual prefix of an already
traversable path, so symlinked components do not open a hole either. Keeping it beats removing it
— an empty `checkout` would interpolate `/scripts/lib/...` and probe the filesystem root. Better
still, the precondition is *protected*: removing the `-f` test costs **3** assertions (measured,
`60/3`), so a refactor that reorders them fails loudly rather than quietly reviving dead code.
Declaring it was the right call.

---

### 3. Convention compliance

- **Gitmoji** — all four commit subjects lead with an emoji (`✨` `✅` `🚨` `🐛`); PR title
  `✨ Add a runtime version guard for the MemPalace memory server`. No conventional-commit prefix
  anywhere. ✅
- **PR body** — all three mandated sections present (`## How to read this PR?` :7,
  `## How to test this PR?` :20, `## Detailed description (for agents)` :57). ✅
- **Close keyword** — `closingIssuesReferences` resolves to exactly `[623]`. ✅
- **Branch + worktree** — `feat/0623-mempalace-runtime-version-guard`, `.worktrees/0623`. ✅
- **Version bump** — no skill/agent source, no `SKILL.md`/`AGENT.md`, nothing under `artifacts/`
  or `extensions/` in the diff; `check-skill-versions.sh` agrees. ✅
- **CLI matrix (R12)** — row `7d` rewritten in the same diff. Iteration-1 finding 14 is closed:
  the row now cites **"spec 0084"** for the `tls-exec.sh` provenance instead of the duplicated
  label `7g` (which is still ×2 at both base and head — pre-existing, no longer load-bearing).
  Finding 15 is closed too: both argv elements now render with `<repo>/`. ✅
- **Logbook (Rules A/B)** — #623 is its own logbook, 11 comments, 10 of them dated across this
  working session, so the journal is incremental rather than batched. ✅
- **CI wiring — three surfaces, count established independently.** `check-test-wiring.sh` scans
  `.github/workflows/*.yml` plus `ci/ci-capabilities.yml`; `.gitlab-ci.yml` is generated from the
  latter and drift-gated by `build-ci.sh --check`, which `check-ci-parity.sh` composes. All three
  carry the install step. `Taskfile.yml` is **not** a fourth CI surface — its only change is the
  new `mempalace:doctor` operator entry point (R7); it wires no test aggregate.

  **And the install step's position is genuinely load-bearing, not merely tidy.** I moved it from
  before the three spec-0108 test steps to after them and re-ran the parity check:

  ```
  DRIFT: capability 'check-components' (github-actions): business step diverges from the
  declared command — unexpected step: 'bash scripts/tests/test-mempalace-pin.sh' (R3)
  ```

  Exit **1** relocated, exit **0** unmodified. Arm 1 matches business steps against the
  capability's command list *in order* (`check-ci-parity.sh:324,354`), so the ordering is
  enforced, not conventional. ✅

---

### 4. Test coverage — every iteration-1 finding, re-measured

Baseline 184 (`36 / 85 / 63`). Each row is a mutation I applied to the head tree in an isolated
copy; "iter-1" is what that same mutation cost then.

| # | mutation | suite | iter-1 | now |
|---|---|---|---|---|
| **B1** | primary comparator floor → `0` | pin | **green** | **`34/2`** |
| **B2** | version-divergence block disabled | doctor | **`46/0`** | **`58/5`** |
| **B2** | divergence message renamed | doctor | `46/0` | `61/2` |
| **B2** | divergence table header suppressed | doctor | — | `61/2` |
| ctrl | pin-divergence block disabled | doctor | `45/1` | `61/2` |
| 3 | `Supported range:` value redacted | guard | `79/0` | `79/6` |
| 4 | wrong `Resolved interpreter:` reported | guard | `79/0` | `79/6` |
| 5 | version-found value redacted | guard | 5 of 6 siblings only | `78/7` |
| 6 | `version served:` redacted | doctor | `46/0` | `62/1` |
| 6 | `verdict:` field removed | doctor | `46/0` | `62/1` |
| 6 | `packaging importable:` removed | doctor | `46/0` | `62/1` |
| 7 | `Comparator:` always `not reached` | guard | 2 fails | `80/5` |
| 8 | Phase A range check disabled | guard | 3 cases still passed | `51/34` |
| 10 | R5 scan target unreadable (grep exit 2) | pin | scored a **pass** | `35/2` |
| 13 | `-f` wrapper test removed | doctor | — | `60/3` |
| ctrl | R4 restart note removed | guard | 8 fails | `77/8` |

**Whole-line anchoring is real.** `field_line()` reproduces `_refuse`'s own
`"  {0:<26}{1}"` format and every field is matched with `grep -qxF` — whole line, fixed string.
`doctor_field_line()` + `has_exact()` do the same for the doctor's `printf '    %-24s%s\n'`. That
is what turns rows 3–6 from free into one-assertion-each. I checked the residual risk the brief
flagged — an assertion still satisfiable by a neighbouring field — and found none: the substring
form (`has`/`lacks`) survives only for full trigger *sentences* (`reported versions DIVERGE`,
`declared pins DIVERGE`, `lies outside`), which are unique strings, while every *value* assertion
is whole-line.

**Finding 9 is closed, and the count is no longer invariant.** On the synthesised
no-`packaging`-anywhere machine the pin suite reports **`32 passed, 1 failed`** against 36/0 here
— the three former skips are narrated as `SKIPPED (not counted)` and the missing coverage is a
counted failure in 4c. A reader can now tell from the number alone whether the differential ran.

**Scenario 6 is properly isolated.** Two sources report *different* versions, both *inside* the
pin (`3.6.0` vs `3.6.0.1`), every source reading one pin — and it guards its own derivation
first (`:612-616` fails loudly if the second version ever falls outside the pin, so the scenario
can never silently become a range test). It then asserts the **absence** of both other triggers
(`lacks "lies outside"`, `lacks "declared pins DIVERGE"`), which is what stops it degrading into
scenario 1. I tried to make either of the other two triggers produce its outcome and could not.
Disabling the version-divergence detection now breaks exactly the five assertions that should
break:

```
FAIL: R8: reports the VERSION divergence explicitly (not merely some divergence)
FAIL: divergent-versions-only machine: exited 0 — two sources report different MemPalace versions
FAIL: R8: the version-divergence trigger fires on its own, with nothing else wrong
FAIL: R8: tabulates every source with the version it reports
FAIL: R8: the divergence table does not attribute 3.6.0.1 to a named source
```

**R11** is asserted structurally (`test-mempalace-doctor.sh:716-754`), and I confirmed the premise
independently: all four registrations name the one wrapper
(`setup-claude-interactive.sh:220`, `config/gemini/settings.json:35`,
`config/copilot/mcp-config.json.template:7`, `setup-antigravity-interactive.sh:219`) and nothing
registers `mempalace-mcp` directly.

---

### 5. Linter findings

`lint-shell.sh` — exit 1. New to this diff, both **style-only**, on the pin suite:

```
In scripts/tests/test-mempalace-pin.sh line 690:
echo "$arm_a_out" | sed 's/^/      /'
^-- SC2001 (style): See if you can use ${variable//search/replace} instead.

In scripts/tests/test-mempalace-pin.sh line 742:
  echo "$arm_b_out" | sed 's/^/      /'
^-- SC2001 (style): See if you can use ${variable//search/replace} instead.
```

`SC1091` on `doctor-mempalace.sh:45` and `test-mempalace-pin.sh:189` is an artifact of not passing
`shellcheck -x`, not a defect. Findings against `scripts/lib/common.sh` (`SC2015` ×2, `SC2088`,
mode 100644, missing `set -e`) are pre-existing — `git ls-tree` confirms mode 100644 at base
`5c089af` too — and outside this diff's lines. Both new test suites
(`test-mempalace-doctor.sh`, `test-mempalace-runtime-guard.sh`) produce **zero** shellcheck output.

`lint-python.sh` — exit 1, degraded (`neither ruff nor flake8 found`). All ten `bare print()` hits
are false positives, verified rather than assumed: `mempalace-http-wrapper.py:146` and `:268` both
pass `file=sys.stderr`, and every `mempalace_pin.py` hit (`240-264`) sits after the `__main__`
guard at `:180`. The heuristic understands neither.

`lint-markdown.sh docs/cli-matrix.md` — clean, exit 0. `lint-json.sh`, `lint-skill.sh` — nothing
applicable in the diff.

No line-length convention exists in `docs/scripting-conventions.md`, and the new
`acknowledged-exception:` tag is shorter than the pre-existing one in the wrapper, so its length
is not a violation.

---

## Findings

**Zero blocking.** Two new findings, both non-blocking; the nineteen from iteration 1 are
addressed or correctly deferred.

### New — test coverage

**N1 — `class: tech`, non-blocking.** A **second** unasserted branch exists in the doctor, and
unlike `CHECKOUT UNRESOLVABLE` it is neither declared nor unreachable. `evaluate()`'s no-version
branch (`doctor-mempalace.sh:212-216`) is never executed by the suite at all. Poisoning **both**
its lines — the `field "verdict:"` and the `note_failure` — leaves the suite fully green:

```
field "verdict:" "POISONED-BRANCH-WAS-EXECUTED"
note_failure "POISONED"
  -> Summary: 63 passed, 0 failed
```

`grep -c "NO VERSION" scripts/tests/test-mempalace-doctor.sh` → **0**.

It is reachable, and I proved that rather than asserting it. `--probe` from a real interpreter
carrying no MemPalace prints `dist=absent`, which is precisely the gate at `:212`:

```
$ <venv-without-mempalace>/bin/python scripts/lib/mempalace_pin.py --probe
interpreter=/private/tmp/.../v/bin/python
dist=absent
attr=absent
packaging=no
```

`evaluate()` is called from three sites, including per-CLI registration (`:341`) — so a
registration whose interpreter lost its MemPalace, the #623 shape one dimension over, lands here.
The branch is the doctor's only path to reporting "this source serves nothing", and it also owns
the `note_failure` that makes such a machine exit non-zero.

Why non-blocking: it is the *diagnostic*, not the guard. The wrapper's equivalent absence path is
covered — with `-S`, as iteration 1 established — so nothing about enforcement is at risk. The
reason it is worth reporting anyway is the commit message's own completeness claim: *"One branch
is knowingly unasserted and declared rather than hidden."* There are two, and this is the
undeclared one. A mirror of scenario 6's method closes it: one source whose interpreter resolves
no distribution, asserting the `NO VERSION` verdict line and the non-zero exit.

### New — the specification itself (`class: spec`, non-blocking)

**N2 — a fourth `spec` finding, distinct from the three already deferred.** I am not re-reporting
iteration 1's 17 (R7's "launch path would itself select"), 18 (R9/R10 tension) or 19 (R3's four
fields on non-R1 refusals); I concur with all three and they route to SPECS. This is a new one,
and it is the reason N1 has no specification to be tested against.

**R8 has no clause for a source that reports no version at all.** Its two triggers are "any two
of the versions it reports differ" and "any version it reports lies outside the pinned range".
An interpreter resolving *no* MemPalace distribution satisfies neither: it contributes no version
to compare, and `evaluate()` correctly does not `record_version` it (`:216` returns before the
call), so it cannot participate in the divergence set either. R7 is silent in the same way — it
requires reporting "the MemPalace version that interpreter serves", with no answer for an
interpreter that serves none.

The implementation resolves this sensibly and strictly — `note_failure` at `:214`, so such a
machine does exit non-zero — but that behaviour is *unspecified*, which is exactly why it is
also *unasserted*. This is the same shape as deferred finding 19, one requirement over: a
non-successful outcome the spec never contemplates. Worth folding into the same delta rather than
its own.

A supporting observation for whoever writes that delta: the spec's headline *Divergent installs
are reported and flagged* scenario fires **two** of R8's three triggers at once (`3.0.0` is both
divergent *and* below the floor). A scenario that cannot distinguish its triggers is precisely
what let the original test pass with divergence detection deleted. Scenario 6 now covers the gap
in the suite, but the spec's own acceptance criterion still does not isolate it.

### New — nit

**N3 — `class: tech`, non-blocking.** `SC2001` ×2 (`test-mempalace-pin.sh:690`, `:742`) — a
`sed`-in-a-pipe where `${var//search/replace}` would do. Style only; both lines are new in this
diff, so the linter is reporting on this change rather than on inherited debt.

### Iteration-1 findings — disposition

All measured, not taken on trust. **B1** and **B2** discharged (evidence in sections 2 and 4).
**3, 4, 5, 6, 7, 8, 9, 10** discharged — each mutation-proven in the table above, every one of
which previously cost zero or fewer assertions than its label implied. **11** discharged
(`except ValueError` on both branches, `InvalidVersion` confirmed a `ValueError` subclass, plus
the overlong-release-segment totality row). **12** discharged (`finally`, and only when this file
added the entry). **13** discharged, and its precondition protected. **14, 15** discharged in
`docs/cli-matrix.md` row 7d. **16** discharged — the malformed-`common.sh` branch now carries its
own remedy (`mempalace-http-wrapper.py:233-241`) instead of pointing at a setup re-run that would
rewrite nothing. **17, 18, 19** correctly deferred to a future delta.

---

**Process note for the record.** `7ca9d16` shipped a required-check failure that a local run of
the workflow's own grep would have caught before the push; `9671ded`'s message says as much and
records that both the plan review and the DEV report predicted the gate but not the file. That
self-correction is the reason this iteration is an approval rather than a third loop, and the
honesty of it is worth more than the mistake cost.

The change is in good shape. The two remaining findings are non-blocking, so per
`AGENTS.md` → *Retroactive review loop*, note that **termination is not met by this pass**: the
criterion requires an `APPROVE` verdict **and** zero findings of any class. N1 routes to DEV,
N2 to SPECS alongside the three already deferred.

Merge stays gated on the user, per `AGENTS.md` → *Branching Strategy*.

hcross:
## REVIEW iteration 3 — orchestrator self-verification, **not a cold review**

Labelled honestly up front: this pass was performed by the orchestrator, not by an independent cold reviewer. It therefore does **not** carry the independence that iterations 1 and 2 did, and should not be read as equivalent to them.

**Why.** A fresh cold reviewer was spawned for this iteration with a deliberately narrowed brief. It ran for roughly eight hours and died on an API error mid-response without posting a verdict — its own task decomposition survived, but no artifact did. Spawning a sixth agent on this ticket to certify one assertion and two style substitutions was judged disproportionate. The residual scope was **measurement, not judgement** — poison a branch and observe the suite go red, run a linter, re-derive counts — and a non-cold verifier degrades a measurement far less than it degrades an assessment. Recorded so a future audit sees the trade rather than inferring it.

Iteration 2 (comment `5170510919`) returned **APPROVE with zero blocking findings**. The only change since is `60b10c6`: one test scenario plus two style substitutions, no implementation change.

### N1 — closed, mutation-proven

The finding: `scripts/doctor-mempalace.sh:212-216` (the no-version branch) was reachable and asserted by nothing; poisoning it left the suite at `63 passed, 0 failed`.

Verified by mutating the implementation and observing the suite, in the worktree with `git checkout --` restore after each:

| State | Doctor suite |
|---|---|
| unmutated control | `72 passed, 0 failed` |
| **P1** — the `NO VERSION` verdict line poisoned | **`71 passed, 1 failed`** |
| **P2** — the whole `212-216` guard removed | **`66 passed, 6 failed`** |
| restored | `72 passed, 0 failed`, `git status` clean |

**P2 is what makes the closure non-tautological.** With the guard gone the row falls through and the doctor begins claiming three things that did not happen — an out-of-range version, an `OUT_OF_RANGE` verdict, and a divergence between the pin floor and the literal string `absent`. The scenario's three negative assertions are what fire there, alongside the two whole-line positives and the finding count. A scenario asserting only the verdict line and the exit status — which is what the finding as written prescribed — would have passed on a doctor that misdiagnoses.

### N1's fixture — audited for vacuity, sound by construction

The concern: a fixture on `PATH` or `PYTHONPATH` can only *shadow* a real install, never establish absence, so on any machine whose interpreter carries a real MemPalace the scenario would assert nothing. The fixture establishes absence by **replacement** — a shim exec'ing with `-S` (site-packages suppressed) and `env -u PYTHONPATH` (ambient dropped) — and the suite carries a guard asserting absence really was established before relying on it. The rationale is documented at the builder rather than left implicit.

Confirmed interpreter-independent: `72 passed, 0 failed` on both the default interpreter and one carrying a real MemPalace 3.6.0.

### N3 — closed

`shellcheck scripts/tests/test-mempalace-pin.sh` → **zero** `SC2001`. The third `sed` pipe at `:502` is deliberately untouched: it is `printf '%s\n' … | sed`, which `SC2001` does not match (its pattern is `echo "$var" | sed`), so it sits outside the finding. **I agree with leaving it** — retouching a line iteration 2 verified, to satisfy a consistency argument no linter is making, is risk without return.

### No regression — counts re-derived, not transcribed

| Suite | default interpreter | pipx interpreter (real `packaging` + real MemPalace 3.6.0) |
|---|---|---|
| `test-mempalace-pin` | 36 / 0 | 36 / 0 |
| `test-mempalace-runtime-guard` | 85 / 0 | 85 / 0 |
| `test-mempalace-doctor` | 72 / 0 | 72 / 0 |

**193 assertions**, matching the claim.

### CI — green on the exact head, addressed by SHA

`GET /repos/crewrig/crewrig/commits/60b10c6/check-runs` → **12 success, 0 other**. Addressed by commit SHA rather than through the PR view, because a monitor earlier in this session reported a correct green under a stale commit label — and iteration 2's own brief named a head commit that had in fact *failed* a required check, which that reviewer caught. Establishing the head independently is not paranoia here; it has already mattered twice.

### Termination

`AGENTS.md` → *Retroactive review loop* requires an APPROVE verdict, zero findings of any class, and CI green on the reviewed head.

- Zero `tech` findings outstanding: N1 and N3 closed and verified above; N2 and the four other `spec`-class findings are tracked on **#703** and are no longer findings against this PR.
- CI green on `60b10c6`, SHA-verified.
- **The APPROVE on this head is a self-verification, not an independent one.** That is the one respect in which this pass is weaker than the criterion's spirit, and it is stated rather than papered over. The last *independent* APPROVE was iteration 2 on `9671ded`; everything added since is one test scenario and two style substitutions, both verified here by mutation.

Merge remains gated on the user.

Issue #703, opened 2026-08-03
📝 Spec 0108 — four requirements whose intent is legible but whose letter is not satisfiable


hcross:
## Context

Four `spec`-class findings accumulated across two cold plan reviews and two cold PR reviews of spec 0108 (issue #623). None blocks: the implementation honours every requirement's *intent*, and each reviewer reconstructed that intent correctly. The cost is that they each had to.

They are collected here rather than folded into #623 for two reasons. Amending a spec while its implementation is in flight churns the contract under the people implementing it. And one delta covering four clarifications is cheaper to write, review and gate than four rounds.

**All four are clarifications.** None changes what the implementation must do, so none invalidates the work merged for #623. The expected version bump is therefore `PATCH`, matching the precedent of `0108-delta-01`, which corrected requirement 2 for exactly this class of defect.

### 1. R7 is literally unsatisfiable — "what the launch path would itself select"

R7 asks the diagnostic to report "the interpreter and version the memory-server launch path would itself select". The launch path selects nothing at launch: it runs what its registration recorded. `detect_mempalace_python` is a **setup-time** actor, not a launch-time one.

The implementation resolved this by splitting the diagnostic into separately labelled sections — "what a session actually launches" (reading the four registrations) versus "what a fresh setup would select" — which is the right answer. But it is an inference, and the next reader will have to make it again.

*Found by the first cold plan reviewer; independently re-raised by the second cold PR reviewer.*

### 2. R9 and R10 are in literal tension

R9 forbids the diagnostic from depending on "an enumeration of known Python package managers". R10 requires it to report the framework's interpreter-resolution candidate ordering — whose candidate 1 is a pipx-specific path (`scripts/lib/common.sh`, the `$HOME/.local/pipx/venvs/mempalace/bin/python` entry).

Read strictly, satisfying R10 violates R9. The reconciliation is legible — R9 governs how *facts are derived* (from what resolves now), R10 governs *reporting an existing framework behaviour* — but the text does not say so.

### 3. R3 assumes every refusal is an R1 refusal

R3 mandates that a refusal name four things, one of them the pinned supported range. Two of the guard's five refusal paths cannot: when the pin itself is unreadable, or when the guard module is unimportable, there is no range to name, and the implementation correctly prints `not determined`.

That is the right behaviour, and R4's restart notice *is* satisfied on those paths. But R3's compliance argument currently rests on reading "a refusal per requirement 1" narrowly, rather than on text that addresses the case.

[...]

PR #708, opened 2026-08-04
✨ Make a spec's recorded status tell the truth, and enforce it

### LINKED ISSUE #700: 🐛 45 merged parent specs sit at `status: draft` — the lifecycle table's own invariant is false on `main`
## Problem

`docs/spec-format.md` → *Lifecycle states* defines `approved` as the status of a spec whose spec-PR has merged on `main`, and *Recording a status transition* asserts outright:

> a merged spec is never at once merged and still `draft`

That invariant is **false on `main` today**, for a large fraction of the corpus.

## Census (measured against `crewrig/main` at `5298697`, 2026-08-03)

| Kind | `draft` | `approved` | `implemented` |
|---|---|---|---|
| Parent specs | **45** | 10 | 53 |
| Delta specs | 30 | 4 | 2 |

The 30 draft deltas are **not** violations: an established convention (verified against spec 0069 during ticket #570) transitions only the parent spec and leaves each `delta-NN` at `draft`. If that convention is right, it should be written into `docs/spec-format.md`, which does not currently state it — but it is not what this issue is about.

The **45 parent specs** are genuine violations. Every spec on `main` has by definition had its spec-PR merged, so every one of them should carry at least `approved`.

## Why this is more than cosmetic

The backlog includes the lifecycle's own load-bearing contracts — enforced daily while recorded as drafts:

- `0003-spec-pr-workflow` — the very convention whose merge mechanic this issue is about
- `0004-plan-format-and-review`
- `0005-retroactive-routing-engine`
- `0006-interaction-modes-and-sizing`
- `0009-spec-linter`
- `0012-core-framework-separation`, `0013-layer-taxonomy-boundary-contract`, …

Concrete consequences:

1. **`status` carries no information.** A reader cannot distinguish "merged and in force" from "never landed", so the field cannot be used to answer the question it exists for. An agent grounding itself on a spec has no way to tell whether it is reading a proposal or a contract.
2. **The invariant is quoted as fact.** `docs/spec-format.md` states it flatly, so an agent that trusts the documentation will draw false conclusions about any spec it inspects.
3. **The violation rate suggests the mechanic, not the operators, is at fault.** 45 misses is not inattention; it is a step that sits in the wrong place. The edit must happen on the spec branch *between* the approval gate and `gh pr merge` — a two-command window, with no enforcement, at the exact moment attention has moved to the merge. Nothing fails if it is skipped.

## Two things worth deciding, not just fixing

**Backfill.** A metadata-only sweep can set the 45 to their true state — but their true state is often `implemented`, not `approved`, since most have shipped. Determining which is which requires checking each `related-issue`, so the sweep is not purely mechanical. Options: one PR per spec (auditable, 45 PRs), one PR for all (one review, large diff), or backfill only the specs whose `related-issue` is closed and shipped.

**Enforcement, so it stops recurring.** The lint already parses spec frontmatter (`scripts/lib/spec-linter.js`), and `lint-specs` runs in CI on every PR. A check that fails when a spec present on the PR's *merge base* carries `status: draft` would make the invariant self-enforcing: a spec on `main` could never be `draft` again, and the failure would name the file. That inverts the current situation, where the rule is documented prose with no mechanical backing.

## Discovery context

Surfaced as a non-blocking `spec`-class finding by the cold plan reviewer on issue #623: it was reviewing a plan against spec 0108, noticed that 0108 itself was still `draft` on `main`, and reported it instead of working around it. That single instance was corrected by PR #699; the census above showed the instance was not isolated.

hcross:
`docs/spec-format.md` asserted that *"a merged spec is never at once merged and still `draft`"*. It was false for **44 of 109** non-delta specs on `main`, including the lifecycle's own contracts — enforced daily while recorded as drafts — so `status` could not answer the question it exists for. This corrects all 44 and adds the check that keeps the invariant true.

Implements spec 0109 and its delta-01.

Closes #700

## How to read this PR?

51 files, but only four things happen. Read in this order:

1. **`scripts/lib/spec-linter.js`** — the check. The linter had **no base-branch awareness at all**; even the spec-0098 duplicate-id guard compared only files within the tree under test. The load-bearing detail is the discriminator: base-branch *presence* selects which files are in scope, while the `status` read is the **working tree's**. That is the only reading under which the corrected tree passes while `main` still carries the drafts.
2. **`scripts/tests/test-spec-linter.sh`** — six new cases covering all five spec scenarios plus both environment branches.
3. **`docs/spec-format.md`** — the invariant stated as a findable rule, naming the linter as its enforcement.
4. **`specs/*.md` × 44** — `status: draft` → `implemented`. Five also gain `interaction-mode: INTERMEDIATE`; see below.

The CI files and `ci/test-wiring-exemptions.txt` are the fourth thing, explained under *scope judgement*.

## How to test this PR?

```sh
npm install                              # a fresh worktree has no node_modules; the suite needs it
task spec:lint                           # -> Linting passed!
bash scripts/tests/test-spec-linter.sh   # -> Results: 34 passed, 0 failed
```

**Prove the check actually catches a regression**, rather than trusting a green run:

```sh
sed -i '' 's/^status: implemented/status: draft/' specs/0009-spec-linter.md
task spec:lint
# -> [FAIL] Non-delta specs present on the base branch (crewrig/main) carry 'status: draft':
#    - specs/0009-spec-linter.md
git checkout -- specs/0009-spec-linter.md
```

**Prove requirement 8's removal-detector fires**, which is the assertion that matters most here:

```sh
# delete the offender block from spec-linter.js, then:
bash scripts/tests/test-spec-linter.sh    # -> Cases 28 and 29 go red
```

Both CI paths were simulated locally: `BASE_REF=crewrig/main` (pull-request path) and `BASE_REF=` (push path, falling back to the linter's own derivation).

**The one decision the spec left open, and why it became three outcomes rather than two.** "No base ref" is not a single situation. Outside a git work tree there is no change under test, so requirement 2's subject — *the base branch of the change under test* — does not exist: the check **skips with a notice naming what was not checked**, which is the temp-fixture case and cannot occur in CI. Inside a repository with the base resolvable, it enforces. Inside a repository with the base **unresolvable**, it exits **2** and fails closed: that means a wiring fault (missing `fetch-depth: 0`, an unfetched remote), and reporting it green is precisely the green-does-not-mean-checked defect this spec exists to remove. Exit 2 rather than the linter's own 1 marks an environment fault rather than a lint finding, matching `check-skill-versions.sh`. Both branches are pinned by tests and both go red under mutation.

Fail-closed has a consequence that had to be wired rather than hoped for: `lint-specs` now needs `fetch-depth: 0` and a PR-conditional `BASE_REF`. On `push` there is no PR context, `BASE_REF` is falsy, and the linter falls back to its own derivation — a test case pins that path.

**Base-ref resolution reuses `check-skill-versions.sh:24-33`** rather than inventing a second idiom: `BASE_REF` → first remote matching `crewrig|origin` (else the first remote) + `/main` → `rev-parse --verify` → `--depth=50` fetch retry → an error naming the override. One documented difference: the linter's positional arguments are spec targets, so `BASE_REF` is the only override.

**The corrections are mechanical, on evidence re-derived rather than inherited.** All 44 `related-issue`s are CLOSED/COMPLETED, and they were closed by 62 distinct linked pull requests, **62/62 MERGED**. No spec resolved to anything other than `implemented`, so none had to be forced to the majority — requirement 4's "the status that reflects its true state" was honoured rather than shortcut.

**Five of the 44 also gain `interaction-mode: INTERMEDIATE`.** They carried none, because the frontmatter schema requires that field only from `approved` onward and lets it default while `draft`. A status-only edit would leave them failing the linter's own frontmatter check, and leaving them at `draft` would violate requirements 1 and 4 and make the new check fail on the change that introduces it. Spec 0109's original requirement 5 forbade touching the field while citing a carve-out that *mandates* setting it — **delta-01 (#707) corrected that contradiction before this PR was opened**, so this diff conforms to the letter as well as the intent.

**Requirement 6 is why this is one change.** A check landing before the corrections would fail on `main` and on every subsequent PR; corrections landing without the check would leave the invariant free to decay again. Verified by running the new check against this tree with both halves in.

**One scope judgement, flagged for veto.** `scripts/tests/test-spec-linter.sh` was **exempted from CI**, so requirement 8's removal-detector would never have run: deleting the check would produce a green pipeline once the corpus was clean — the same green-does-not-mean-checked defect one level up. The suite is now wired into `lint-specs` and the exemption line deleted per that file's own hygiene rule. The exemption's stated blocker — no npm-installed `node_modules` — does not exist in `lint-specs`, which already runs `npm install`. Cost: 16 seconds. Reverting is three lines plus a `build-ci.sh` regeneration.

**Delta-specs are untouched** and remain 30 `draft` / 5 `approved` / 2 `implemented`. Spec 0109 deliberately leaves their status to its own ticket: the "only the parent transitions" convention rests on one verified instance against seven counterexamples, which is not a convention to codify.

**`docs/cli-matrix.md` is not updated**, deliberately: the diff touches none of the paths its maintenance rule enumerates. `build.yml` is not `claude.yml` or `gemini.yml`, and `build-ci.sh` itself is unmodified — only its generated output.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -285,6 +285,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # full history so the base ref resolves (spec 0109)
 
       - uses: actions/setup-node@v4
         with:
@@ -297,7 +299,20 @@ jobs:
       - name: Install Task
         run: sudo sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d -b /usr/local/bin
 
+      # Runs before the corpus lint so a broken linter is diagnosed by its own
+      # suite rather than reported only as a corpus failure. This job is where
+      # the suite belongs: it provisions the npm-installed node_modules the
+      # suite needs, which the hermetic check-components job does not.
+      - name: Test spec linter regression
+        run: bash scripts/tests/test-spec-linter.sh
+
       - name: Lint specs
+        env:
+          # The base-branch `status: draft` check (spec 0109 R2) compares against
+          # the PR's *target* branch. On a push there is no PR context, so the
+          # empty value hands the linter back to its own default derivation
+          # (first `crewrig|origin` remote + /main).
+          BASE_REF: ${{ github.event_name == 'pull_request' && format('origin/{0}', github.event.pull_request.base.ref) || '' }}
         run: task spec:lint
 
   test-harness-curate:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -313,6 +313,15 @@ jobs:
           # empty value hands the linter back to its own default derivation
           # (first `crewrig|origin` remote + /main).
           BASE_REF: ${{ github.event_name == 'pull_request' && format('origin/{0}', github.event.pull_request.base.ref) || '' }}
+        # Read the log line, not the exit code, to tell an unresolvable-base
+        # wiring fault from a lint finding: the linter distinguishes them (exit 2
+        # vs 1), but Task collapses any non-zero status into its own 201, so only
+        # the `[ERROR]` line survives this wrapper. Kept behind `task` on purpose
+        # — `ci/ci-capabilities.yml` declares this command for both engines
+        # (check-ci-parity.sh enforces the agreement, and .gitlab-ci.yml is
+        # generated from it), and nothing here reads the exit code beyond
+        # pass/fail, so bypassing the wrapper would cost symmetry and buy no
+        # observable distinction.
         run: task spec:lint
 
   test-harness-curate:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -122,10 +122,13 @@ lint-markdown:
 
 lint-specs:
   image: node:22
+  variables:
+    GIT_DEPTH: "0"
   before_script:
     - sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d -b /usr/local/bin
   script:
     - "npm install"
+    - "bash scripts/tests/test-spec-linter.sh"
     - "task spec:lint"
   rules:
     - if: '$CI_PIPELINE_SOURCE == "merge_request_event" && $CI_MERGE_REQUEST_TARGET_BRANCH_NAME =~ /^main$|^release\//'
--- a/ci/ci-capabilities.yml
+++ b/ci/ci-capabilities.yml
@@ -138,6 +138,9 @@ capabilities:
       - npm install -g markdownlint-cli
       - markdownlint "**/*.md" --ignore node_modules --ignore extension-skeleton --ignore communication --ignore 'extensions/**/commands/*.md'
 
+  # The spec linter's base-branch `status: draft` check (spec 0109 R2) resolves
+  # the change's base ref, so the job needs full source history — declared as
+  # `history-depth: full`, same as the two version-bump guards below.
   - id: lint-specs
     name: "Lint specification files"
     trigger:
@@ -149,8 +152,10 @@ capabilities:
     requires:
       runtime: node@22
       tools: [task]
+      history-depth: full
     command:
       - npm install
+      - bash scripts/tests/test-spec-linter.sh
       - task spec:lint
 
   - id: test-harness-curate
--- a/ci/test-wiring-exemptions.txt
+++ b/ci/test-wiring-exemptions.txt
@@ -17,5 +17,4 @@
 # the fact that the underlying bug is resolved.
 
 # --- Environment-dependent (cannot run in the hermetic check-components job) ---
-test-spec-linter.sh	requires the repo's npm-installed node_modules (markdownlint-cli via npx, plus js-yaml/semver) which the hermetic check-components job does not provision; spec-linter.js is exercised on real specs by the lint-specs job via `task spec:lint`.
 test-e2e-auth-scripts.sh	exercises scripts/e2e/auth-copilot.sh, which e2e_die's when ~/.copilot/instructions/ is absent — a `task setup:copilot` side-effect present on a developer host but not in a clean CI HOME; cannot pass in the hermetic check-components job. See follow-up note from #534 on whether the guidance path should print before that precondition.
--- a/docs/spec-format.md
+++ b/docs/spec-format.md
@@ -200,6 +200,26 @@ table below is the contract.
 A `status` regression (e.g. `approved` → `draft`) is prohibited; if a
 spec must be re-opened, supersede it instead.
 
+### No spec on `main` is a draft
+
+A spec file present on `main`, other than a delta-spec, SHALL NOT carry
+`status: draft`. A spec reaches `main` only by its own merged spec-PR, which is
+the trigger the table above assigns to `approved`; `draft` on `main` therefore
+records a contradiction rather than a lagging value, and leaves a reader unable
+to use `status` for the one question it exists to answer — "merged and in
+force" or "proposed but never landed".
+
+The rule is enforced mechanically rather than by recall. The spec linter
+(`scripts/lib/spec-linter.js`, run as `task spec:lint` in CI) fails and names
+every non-delta spec that is **already present on the change's base branch**
+while carrying `status: draft`. Presence on the base branch is the
+discriminator, so a spec the change under test *introduces* is not flagged: it
+is legitimately `draft` until the *Merge mechanic* below flips it. Delta-specs
+are exempt too — what `status` a delta should carry is deliberately unsettled,
+per
+[`specs/0109-spec-status-invariant-on-main.md`](../specs/0109-spec-status-invariant-on-main.md)
+→ *Out of scope*.
+
 ### Recording a status transition
 
 The append-only rule (see *Delta-spec convention*) governs a spec's
--- a/scripts/lib/spec-linter.js
+++ b/scripts/lib/spec-linter.js
@@ -92,6 +92,134 @@ function isExcludedSpecPath(relPath, entries) {
     return best !== null && best.policy === 'excluded';
 }
 
+// --- Base-branch status invariant (spec 0109) -------------------------------
+// The helpers below give the linter the one thing it previously lacked: a view
+// of the change's base branch. They exist to answer a single question — is this
+// spec file already present on the base branch? — because that is the
+// discriminator spec 0109 R2 names: a spec the change under test introduces is
+// legitimately `draft` until its own merge mechanic flips it, while a spec
+// already on the base branch carrying `draft` is the contradiction R1 forbids.
+
+// gitCapture(args) — run git, capturing both streams. Returns the exit status
+// and stdout; never throws and never leaks git's own stderr into the linter's
+// output, so callers can probe git freely (including outside a repository).
+function gitCapture(args) {
+    const result = spawnSync('git', args, { encoding: 'utf8' });
+    return {
+        status: result.status,
+        stdout: typeof result.stdout === 'string' ? result.stdout : '',
+    };
+}
+
+// resolveBaseRef() — the base ref to compare against, resolved exactly as
+// scripts/check-skill-versions.sh:24-33 does so the repository has one idiom
+// rather than two: `BASE_REF` when set, else the first remote matching
+// `crewrig|origin` (falling back to the first remote at all) with `/main`
+// appended. Verifies the ref, retrying once behind a shallow-clone `--depth=50`
+// fetch, and returns an `error` (never a silent fallback) if it still does not
+// resolve. The linter's positional arguments are spec targets, so `BASE_REF` is
+// the only override — that is the one difference from the shell sibling.
+function resolveBaseRef() {
+    let ref = process.env.BASE_REF;
+    if (!ref) {
+        const remotes = gitCapture(['remote']).stdout.split('\n').map((r) => r.trim()).filter(Boolean);
+        const preferred = remotes.find((r) => /crewrig|origin/.test(r)) || remotes[0];
+        if (!preferred) {
+            return { error: 'no git remote is configured, so no default base ref could be derived' };
+        }
+        ref = `${preferred}/main`;
+    }
+
+    if (gitCapture(['rev-parse', '--verify', ref]).status === 0) {
+        return { ref };
+    }
+
+    // Covers fresh CI clones, which are shallow by default.
+    const remote = ref.split('/')[0];
+    const branch = ref.slice(remote.length + 1);
+    if (branch) {
+        gitCapture(['fetch', '--depth=50', remote, branch]);
+        if (gitCapture(['rev-parse', '--verify', ref]).status === 0) {
+            return { ref };
+        }
+    }
+
+    return { error: `base ref '${ref}' does not resolve and \`git fetch\` did not recover it` };
+}
+
+// baseBranchPaths(baseRef, relPaths) — the subset of relPaths that exists on
+// baseRef, as a Set. One `git ls-tree` call with the linted paths as pathspecs:
+// git reports only the paths it finds, so the returned Set is the intersection
+// by construction. R3 is satisfied here — the set comes from the repository at
+// check time, never from a recorded list.
+function baseBranchPaths(baseRef, relPaths) {
+    if (relPaths.length === 0) {
+        return new Set();
+    }
+    const result = gitCapture(['ls-tree', '-r', '-z', '--name-only', baseRef, '--', ...relPaths]);
+    if (result.status !== 0) {
+        return null;
+    }
+    return new Set(result.stdout.split('\0').filter(Boolean));
+}
+
+// resolveBaseContext(files) — decide whether the base-branch status check runs,
+// and if so against which paths. Three outcomes, deliberately distinct:
+//
+//   1. Not inside a git work tree → SKIP, with a notice on stderr. Without a
+//      repository there is no "change under test" and so no base branch to be
+//      the discriminator; the check has no subject rather than a subject it
+//      failed to inspect. This is the tarball/temp-fixture case, never CI —
+//      every CI checkout is a git repository.
+//   2. Inside a repository, base ref resolves → ENFORCE.
+//   3. Inside a repository, base ref does NOT resolve → hard error, exit 2.
+//      Fail closed: a repository whose base cannot be resolved is a wiring
+//      fault (missing `fetch-depth: 0`, unfetched remote), and passing it as
+//      green would restore the "green does not mean checked" defect that spec
+//      0109 exists to remove. Exit 2 — not the linter's own exit 1 — marks it
+//      as an environment fault rather than a lint finding, matching
+//      check-skill-versions.sh.
+function resolveBaseContext(files) {
+    const toplevel = gitCapture(['rev-parse', '--show-toplevel']);
+    if (toplevel.status !== 0 || toplevel.stdout.trim() === '') {
+        console.error(
+            `[SKIP] Base-branch status check (spec 0109): not inside a git work tree, `
+            + `so the base branch of the change under test cannot be determined. `
+            + `Specs already on the base branch were NOT checked for 'status: draft'.`
+        );
+        return { enforced: false };
+    }
+    // Both sides are canonicalized before being compared: git reports the
+    // physical work-tree path, so a linted path reached through a symlink
+    // (a symlinked checkout, /tmp on macOS) would otherwise resolve "outside"
+    // the repository and be silently exempted from the check.
+    const repoRoot = fs.realpathSync(toplevel.stdout.trim());
+
+    const { ref, error } = resolveBaseRef();
+    if (error) {
+        console.error(`\n[ERROR] Base-branch status check (spec 0109) cannot run: ${error}.`);
+        console.error(`        Point it at a resolvable ref via the BASE_REF environment variable.`);
+        process.exit(2);
+    }
+
+    // Only paths inside the work tree can be compared against a tree-ish;
+    // anything outside it is not part of the change under test.
+    const relPathByFile = new Map();
+    for (const file of files) {
+        const rel = path.relative(repoRoot, fs.realpathSync(path.resolve(file))).split(path.sep).join('/');
+        if (rel === '' || rel.startsWith('../')) continue;
+        relPathByFile.set(file, rel);
+    }
+
+    const basePaths = baseBranchPaths(ref, Array.from(relPathByFile.values()));
+    if (basePaths === null) {
+        console.error(`\n[ERROR] Base-branch status check (spec 0109): 'git ls-tree ${ref}' failed.`);
+        process.exit(2);
+    }
+
+    return { enforced: true, ref, basePaths, relPathByFile };
+}
+
 function lintFile(filePath) {
     let hasErrors = false;
     const errors = [];
@@ -234,7 +362,7 @@ function lintFile(filePath) {
         }
     }
 
-    return { hasErrors, errors, id, isDelta };
+    return { hasErrors, errors, id, isDelta, status: fm.status };
 }
 
 function run() {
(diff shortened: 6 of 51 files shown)


hcross:
## Verdict: APPROVE

Cold review, no authoring context. Every number, path and behavioural claim below was
re-derived locally against `b393e4c` (base) → `1803d2e` (head) in a throwaway worktree; I
carried nothing over from the PR body, the commit message or the spec. Where a PR claim
survived my own measurement I say so; where it did not, it is a finding.

Seven findings, **none blocking**. The change does what it says: the invariant is real, the
enforcement fires, and the corpus correction is exact.

---

### CI status

All 12 required checks **pass** on `1803d2e` (`check-components` was pending on my first
poll and completed green at 2m51s):

`build` · `check-agents-size` · `check-ci-parity` · `check-components` ·
`check-extension-version-bump` · `check-feedback-routing` · `check-skill-versions` ·
`gitlab-ci-check` · `grep-anti-patterns` · `lint-markdown` · `lint-specs` ·
`test-harness-curate`

Preflight clear — no failing or pending required job gates this review.

---

### Correctness

**The discriminator holds on both cases that matter.** I ran the two adversarial scenarios
against the real corpus, not the fixtures:

- *Spec-PR introducing a draft spec must stay green.* Added a fresh `specs/0110-…md` with
  `status: draft`, ran `BASE_REF=crewrig/main node scripts/lib/spec-linter.js specs` →
  **exit 0**, 148 files linted, not flagged. Correct: absent from base ⇒ exempt.
- *Spec reverted to draft on a feature branch must go red.* `sed`-reverted
  `specs/0009-spec-linter.md` to `draft` → **exit 1**, and the failure names the file
  (`- specs/0009-spec-linter.md`). Correct.

The base-presence/working-tree-status split is the only combination that makes R6's
single-change landing possible, and it is implemented as documented
(`scripts/lib/spec-linter.js:436-459`).

**A false-positive vector I looked for and found already handled:** `specs/_template.md`
carries `status: draft` *and* is present on `main`. It is correctly exempt via the
`_template.md` early return (`spec-linter.js:108`) — confirmed empirically, since the full
corpus lint exits 0 with that file in the tree.

**The three-outcome environment decision.** Both live branches are reachable and both are
pinned:

- *Skip* requires `git rev-parse --show-toplevel` to fail. The suite's Case 33 fixture sits
  under `mktemp -d` (`/var/folders/…`), genuinely outside any work tree. **Unreachable in
  CI** — every checkout is a git repository. The PR's claim holds.
- *Exit 2* is reachable and tested (Case 32). See finding 3 for what does **not** survive
  the wrapper.

**`fetch-depth: 0` is genuinely mandatory — I tried to prove it wasn't and failed.** My
first reading was that the linter's `--depth=50` retry would recover `origin/main`, making
the flag belt-and-braces. It does not. In a `--single-branch` shallow clone (what
`actions/checkout` produces without the flag, since its PR refspec is
`+<sha>:refs/remotes/pull/N/merge`), `remote.origin.fetch` is narrowed:

```
remote.origin.fetch = +refs/heads/feature:refs/remotes/origin/feature
git fetch --depth=50 origin main   → exit 0
git rev-parse --verify origin/main → still fatal
```

The fetch succeeds but updates only `FETCH_HEAD`; no `refs/remotes/origin/main` is created,
so the linter exits 2 and the job fails closed. **`fetch-depth: 0` is required, not
defensive.** The `BASE_REF` conditional is likewise necessary: `lint-specs` triggers on both
`push` and `pull_request`, unlike `check-skill-versions` (PR-only, hence its unconditional
`BASE_REF`). Empty string is falsy in Node, so the push path correctly falls through to the
default derivation.

**Requirement 8's removal-detector: verified by deleting the check, not by reading the test
name.** I removed the `if (baseContext.enforced) { … }` block (841 chars) and re-ran:

```
FAIL  Case 28 — draft non-delta spec on the base branch fails and is named (expected 1, got 0)
FAIL  Case 29 — default base-ref derivation (origin/main) enforces the check (expected 1, got 0)
Results: 32 passed, 2 failed
```

It fires. I then ran four further mutations to test whether the suite is discriminating or
merely present — it is discriminating, with distinct signatures per mutant:

| Mutation | Cases that went red |
|---|---|
| Drop the `!isDelta` exemption | 28, 30, 31 |
| Invert the base-presence discriminator | 28, 29, 31 |
| Force `enforced: false` (always skip) | 28, 29 |
| `process.exit(1)` instead of `exit(2)` | 32 |

Case 28's failure under mutation 1 is on its `must_not_contain` assertion, not its exit
code — the delta-spec got named. That assertion is load-bearing and works.

**The 44 corrections — offender set re-derived independently, exact match.** Enumerating
non-delta specs on `crewrig/main` and reading each `status` from the frontmatter:

| Measure | Verified |
|---|---|
| Non-delta specs on base carrying `status: draft` | **44** |
| `specs/*.md` modified by the PR | **44** |
| Missed (draft on main, uncorrected) | **0** |
| Over-reach (corrected but not draft on main) | **0** |
| Delta-specs modified | **0** |

**Every correction is metadata-only.** 44 files, 49 insertions / 44 deletions; per-file
numstat is exactly `1/1` (39 files) or `2/1` (5 files). `git diff --unified=0` puts every
changed line at line 4 (`status:`) or line 6 (added `interaction-mode:`), and frontmatter
closes at line 9 in all 44. Decisive check: SHA of each file's **post-frontmatter body**,
base vs head → **all 44 byte-identical**. The `status: draft` occurrences in spec *bodies*
were not touched. No `id`, `slug`, `version`, `related-issue` or `complexity` changed
anywhere.

**The five specs that also gained `interaction-mode`.** Exactly 5 non-delta specs lacked the
field on base (`0098`, `0100`, `0101`, `0102`, `0103`) and **zero** lack it at head — the
addition is exhaustive and the set is the one claimed. It is also *mechanically necessary*,
not cosmetic: `spec-linter.js:167-171` fails a non-`draft` spec with no `interaction-mode`,
so a status-only edit would not have passed the linter's own frontmatter check.

Delta-01 genuinely permits it — its replacement R5 requires setting the field "when and only
when that field was absent because the spec was `draft`", which is precisely these five.
`INTERMEDIATE` is the right value *by the schema*: `docs/spec-format.md:29` states the field
"MAY be omitted in `draft`, in which case the value **defaults to `INTERMEDIATE`**". So for a
spec that omitted it while `draft`, `INTERMEDIATE` is the recorded meaning of the omission,
not a convenience. One honest caveat: R5 says "the mode the spec was qualified under", and
for 4 of the 5 no artifact anywhere states that mode — only `0102` carries positive evidence
(its spec-PR #665 records a user SPECS-gate revision, which rules out `AUTO`). Nothing
contradicts `INTERMEDIATE` for any of the five, and `FULL` is argued against by the absence
of the per-iteration REVIEW notifications it mandates. Conforming; see finding 6 for the
underlying spec issue.

**The scope judgement flagged for veto: justified, not overreach.** Two independent reasons.
First, `ci/test-wiring-exemptions.txt`'s own header mandates it — "when a quarantined test is
later fixed and wired into a workflow, DELETE its line here". Second, and stronger: R8 exists
to guarantee a removal-detector, and a detector that never executes satisfies the letter
while defeating the purpose — the same "green does not mean checked" defect one level up.
The exemption's stated blocker was specific to the hermetic `check-components` job and does
not apply to `lint-specs`, which already runs `npm install`. Verified: `check-test-wiring.sh`
→ "all test scripts are wired (60) or exempted with a reason (1)". Negative control —
re-adding the exemption line *also* passes, confirming the guard tolerates redundancy and the
deletion is hygiene rather than a hard requirement. Cost claim confirmed: **15.1s** measured
(`/usr/bin/time -p`), against a claimed 16s.

`docs/cli-matrix.md` is correctly not updated — the diff touches none of the paths the
maintenance rule enumerates; `build.yml` is neither `claude.yml` nor `gemini.yml`, and
`scripts/build-ci.sh` is unmodified.

---

### Convention compliance

- **Gitmoji title** ✓ `✨ Make a spec's recorded status tell the truth, and enforce it`.
- **PR body sections** ✓ all four present (*How to read* / *How to test* / *Detailed
  description*).
- **Branch naming** ✓ `feat/0700-spec-status-invariant` matches `<prefix>/<NNNN>-<slug>`.
- **Spec-PR ordering** ✓ spec 0109 and delta-01 both merged to `main` (#706, #707) before
  this implementation PR — the delta landed at `b393e4c`, ahead of this head.
- **Version bump** ✓ no skill/agent source modified; `check-skill-versions` green.
- **Generated CI in sync** ✓ `build-ci.sh --check` → "OK: .gitlab-ci.yml matches the CI
  capability reference"; `check-ci-parity.sh` → OK.
- **Closing keyword** ✓ `Closes #700` is correct here — #700 is the implementation ticket,
  not a spec-PR anchor.

---

### Test coverage

Suite at head: **34 passed, 0 failed** (`bash scripts/tests/test-spec-linter.sh`), matching
the PR's claim. Six new cases (28–33) map onto all five spec scenarios plus both environment
branches, and the mutation matrix above shows they are individually load-bearing rather than
collectively redundant. This is genuinely good coverage — Case 28 asserting both the named
offender *and* the un-named delta in one invocation is the right call, since an exit code
alone cannot distinguish "flagged the right file" from "flagged both".

---

### Findings

**1 · PR body overstates the corpus denominator — `class: tech`, non-blocking**

The body opens with "*false for **44 of 108** non-delta specs on `main`*". Measured:

```
git ls-tree -r --name-only crewrig/main -- specs/ | grep -E '^specs/[0-9]{4}-' | grep -v '\.delta-' | wc -l
→ 109
```

109, not 108, in every candidate tree. The numerator (44) is exact. The likely intent is
"109 minus spec 0109 itself", but the body does not say so. Cosmetic, but this PR's own
argument is that recorded values should be true.

**2 · New `realpathSync` crashes with an unhandled stack trace on a broken symlink under `specs/` — `class: tech`, non-blocking**

`scripts/lib/spec-linter.js:209` calls `fs.realpathSync(path.resolve(file))` on every linted
path. A dangling symlink under `specs/` throws an uncaught `ENOENT`:

```
Error: ENOENT: no such file or directory, stat '…/specs/zz-broken.md'
    at resolveBaseContext (…/scripts/lib/spec-linter.js:209:48)
    at run (…/scripts/lib/spec-linter.js:386:25)
```

Confirmed a regression, not pre-existing: the base linter (`b393e4c`) handles the identical
fixture cleanly — `[ERROR] markdownlint failed.`, exit 1, no trace. Impact is diagnostic
only; Node still exits non-zero, so fail-closed is preserved. A `try`/`catch` around the
`realpathSync` returning a clean error would restore parity.

**3 · The exit-2 rationale does not survive the CI invocation — `class: tech`, non-blocking**

The design note argues exit 2 "marks an environment fault rather than a lint finding,
matching `check-skill-versions.sh`". Measured through the path CI actually uses:

| Invocation | Environment fault | Lint finding |
|---|---|---|
| `node scripts/lib/spec-linter.js` | **2** | 1 |
| `task spec:lint` (what CI runs) | **201** | **201** |

Task collapses both to 201, so the two are indistinguishable by exit code downstream. The
cited sibling does not have this problem — `check-skill-versions.sh` is invoked directly
(`run: bash scripts/check-skill-versions.sh`, `build.yml:340`), so its exit 2 propagates.
Nothing is broken — the build fails either way and the `[ERROR]` line names the fault in the
log — but the stated property is only true of direct invocation. Either invoke the linter
directly in `lint-specs`, or soften the claim.

**4 · Shellcheck nits on new lines, plus a mode inconsistency — `class: tech`, non-blocking**

`scripts/tests/test-spec-linter.sh:512` and `:526` — SC2164 (`cd "$GITFIX"` without
`|| exit`) on lines this PR adds. Implausible to fail given the preceding `mkdir -p`, but
`cd` failing inside the subshell would commit fixtures to the wrong repository. The two
SC2086 `$targets` hits at `:544`/`:546` are deliberate word-splitting and match the existing
idiom at `:67` — not a finding. Separately: the file is mode `100644` while all 60 sibling
test scripts are `100755`. Pre-existing, but this PR is what promotes it to a CI entry point,
so it is the natural moment to fix it.

**5 · A consistent rename launders a status revert — `class: tech`, non-blocking, low**

Because the discriminator is *path* presence on base, renaming a merged spec's file **and**
its `slug` together lets it revert to `draft` undetected:

```
git mv specs/0009-spec-linter.md specs/0009-spec-linter-v2.md
# slug: spec-linter → spec-linter-v2, status: implemented → draft
→ exit 0
```

The naive variant (rename the file only) *is* caught, by the pre-existing slug↔filename check
— and the duplicate-id guard does not fire because the id remains unique. Contrived, and well
outside the forgot-to-flip-the-field failure this check exists to catch, so I would not hold
the PR for it. Worth a sentence in the spec's *Out of scope* if anyone wants it recorded.

**6 · Spec 0109 R2 leaves the degenerate case unstated — `class: spec`, non-blocking**

You asked whether I would add a seventh instance to #703's pattern rather than re-reporting
the five. **Yes — and this is a distinct instance from delta-01's.** R2 specifies the normal
case ("SHALL fail when a non-delta spec **present on the base branch** … carries
`status: draft`") and, to its credit, explicitly states the permitted case (a spec introduced
by the change SHALL NOT be flagged). What it never states is what the check produces when
R2's own subject — *the base branch of the change under test* — **does not exist or cannot be
resolved**. None of the five scenarios covers it, *Out of scope* does not mention it, and
*Open questions* says "None."

The consequence is visible in this PR: the implementer had to invent a three-outcome policy
(skip / enforce / exit 2) and defend it in the PR body under "*the one decision the spec left
open*". That is exactly #703's shape — a requirement stating what the normal case produces
without stating what the degenerate case produces — and the invented policy is load-bearing
security behaviour (fail-closed) that no requirement authorises.

Secondary, same family: **R1 is scoped to `main`** ("No spec file present on `main` …") while
**R2 generalises to "the base branch of the change under test"**. The spec never reconciles
them. For a PR targeting `release/**` — which `build.yml` explicitly triggers on — those are
different branches, and which one the invariant actually governs is unstated.

The implementation's resolution is, in my judgement, the *right* one; the finding is against
the spec, not the code. It should become a delta-02 or fold into #703's checklist item.

**7 · GitLab `lint-specs` receives no `BASE_REF` — `class: arch`, non-blocking, low**

`build.yml` sets `BASE_REF` conditionally so a PR compares against its real target branch.
`.gitlab-ci.yml` sets only `GIT_DEPTH: "0"`, so an MR targeting `release/**` — which that
job's own rule admits (`=~ /^main$|^release\//`) — falls back to the default derivation and
compares against `origin/main` instead of its actual target. Mostly yields false negatives
(a spec on `release/*` but not `main`, carrying `draft`, is missed).

The structural reason is worth recording: `ci/ci-capabilities.yml` has no vocabulary for
environment variables, so the GitHub side had to be hand-authored in `build.yml` while the
generated GitLab side cannot express it — and both `build-ci.sh --check` and
`check-ci-parity.sh` pass **blind** to the asymmetry, because neither compares env blocks. No
guard would catch this class of divergence today.

---

### Linter findings

`lint-markdown.sh` on all 45 changed markdown files — **exit 0, no output.**

`lint-shell.sh` on `scripts/tests/test-spec-linter.sh` — verbatim, new-code hits only
(pre-existing `:67` SC2086 omitted):

```
scripts/tests/test-spec-linter.sh: not executable in git (mode 100644; expected 100755)

In scripts/tests/test-spec-linter.sh line 512:
  cd "$GITFIX"
  ^----------^ SC2164 (warning): Use 'cd ... || exit' or 'cd ... || return' in case cd fails.

In scripts/tests/test-spec-linter.sh line 526:
  cd "$GITFIX"
  ^----------^ SC2164 (warning): Use 'cd ... || exit' or 'cd ... || return' in case cd fails.

In scripts/tests/test-spec-linter.sh line 544:
    output=$( ( cd "$workdir" && env -u BASE_REF node "$LINTER_JS" $targets 2>&1 ) ) || actual_exit=$?
                                                                   ^------^ SC2086 (info): Double quote to prevent globbing and word splitting.

In scripts/tests/test-spec-linter.sh line 546:
    output=$( ( cd "$workdir" && BASE_REF="$base_ref" node "$LINTER_JS" $targets 2>&1 ) ) || actual_exit=$?
                                                                        ^------^ SC2086 (info): Double quote to prevent globbing and word splitting.
```

`lint-skill.sh`, `lint-python.sh`, `lint-json.sh` — not applicable, no matching files in the
diff.

---

### Summary

Approve. The invariant is enforced, the enforcement is tested against its own removal, the
44-spec correction is exact and provably metadata-only, and the two scope judgements the
author flagged (CI test-wiring, the three-outcome environment policy) are both defensible —
the first mandated by the exemption file's own hygiene rule, the second a reasonable
resolution of a gap the spec genuinely left open. The claims I could most easily have taken
on trust — the removal-detector, the 44-file offender set, `fetch-depth: 0`'s necessity, the
15s cost — all survived independent re-derivation. The one that did not is the `108`
denominator, and the one that is only half-true is exit 2's downstream distinguishability.

Not merging: merge authorisation is the user's gate.

*Posted as a plain comment rather than a formal review event: the `gh` CLI authenticates as
`hcross`, this PR's author, so GitHub refuses review events on it. Issue #692 tracks that
limitation. This comment is the canonical verdict artifact.*

hcross:
## Review findings closed — verified by the orchestrator, not by the DEV pass

The three actionable `tech` findings are closed in `38fbf49`. Stated plainly: **the DEV pass went idle without producing a report**, so every claim below is the orchestrator's own measurement rather than a relayed one. The work itself was on disk and sound; only the account of it was missing.

### Finding 2 — the regression, and the one that mattered

`fs.realpathSync` threw an uncaught `ENOENT` on a dangling symlink under `specs/`, where the base linter handled the same fixture cleanly. Reproduced and re-verified after the fix:

```
$ ln -sf /nonexistent specs/zz-broken-link.md && node scripts/lib/spec-linter.js
[ERROR] Cannot resolve linted path: specs/zz-broken-link.md
$ echo $?
1
```

No stack trace, the path named, exit 1 — the linter's own code rather than the environment-fault exit 2, because a path that cannot be canonicalized cannot be read either: it is unlintable rather than merely unresolvable against the base.

**Failing-first, mutation-proven.** My first attempt at the mutation was invalid — it broke the syntax and failed Cases 1, 9 and 10 broadly, which is noise rather than evidence. The valid mutation makes the `catch` re-throw instead of diagnosing:

```
node --check  -> OK
mutated       -> FAIL  Case 34 — an uncanonicalizable linted path is named,
                       not a stack trace (expected exit 1, got 1)
                 Results: 34 passed, 1 failed
restored      -> Results: 35 passed, 0 failed
```

Note what Case 34 asserts: that the path is **named**. Not the exit code — both routes exit 1, since an uncaught throw also exits 1. Asserting the exit code there would have produced a green test attesting nothing, which is the defect this spec exists to remove.

### Finding 3 — the rationale asserted a property CI cannot observe

Measured before accepting the finding:

```
node scripts/lib/spec-linter.js                            -> 1    (lint finding)
BASE_REF=refs/nonexistent node scripts/lib/spec-linter.js  -> 2    (environment fault)
task spec:lint                                             -> 201
BASE_REF=refs/nonexistent task spec:lint                   -> 201
```

Both engines run it through `task`, which collapses either outcome into Task's own 201. So the distinction the code takes trouble to make is unobservable exactly where it was meant to matter. Fail-closed still holds — 201 is not 0 — but only the `[ERROR]` line survives the wrapper, and the comment now says so, with the measurement, at the call site.

**The wrapper is deliberately kept**, and the reasoning is better than the framing I gave the DEV pass. I called bypassing `task` a wider blast radius; the concrete cost is cross-engine symmetry — `ci/ci-capabilities.yml` declares this command for both engines, `check-ci-parity.sh` enforces the agreement, and `.gitlab-ci.yml` is generated from it — bought for no observable distinction, since nothing downstream reads the status beyond pass/fail.

The cited sibling `check-skill-versions.sh` propagates its exit 2 only because `build.yml` invokes it directly. It is the model for the *code*, not evidence of a CI-observable property here. That correction is now in the comment rather than left implicit.

### Finding 4 — hygiene

`SC2164` gone from both new `cd` lines; the file mode is `100755`, matching all 60 sibling test scripts. Pre-existing, but this PR makes the file a CI entry point.

### Routed rather than fixed here

- **Finding 1** — the count. `44 of 108` was wrong; measured **109**. Corrected in the PR body. The finding's own point is the sharp one: this PR's thesis is that recorded values should be true.
- **Findings 5 and 7** — the consistent-rename hole and the CI-parity blind spot, recorded on **#709**. The second is the more valuable: `build-ci.sh --check` and `check-ci-parity.sh` both report the engines in agreement while their env blocks differ, because neither models env vars and the reference format cannot express one. Two defects stacked, fixable only in order.
- **Finding 6** — the seventh instance of the forbid-without-permit pattern, recorded on **#703** with the observation that it is the sharpest evidence in that issue: it was authored *after* the pattern had been named, by the author who named it.

### Verification after the fix

```
bash scripts/tests/test-spec-linter.sh  -> Results: 35 passed, 0 failed
task spec:lint                          -> Linting passed!
shellcheck scripts/tests/test-spec-linter.sh  -> 0 × SC2164
```

**No second cold review requested for this pass, and that is a judgement rather than an oversight.** The `small` tier's team includes `pr-reviewer`, and it ran and approved; the tier does not require a fresh reviewer per iteration. The delta since that APPROVE is one `catch` block, two comment blocks, a file mode and two `cd` lines — all verified by mutation above. The prior reviewer now carries this ticket's context and could not provide a cold pass anyway.

PR #715, opened 2026-08-05
🐛 Persist transcripts while a peer holds the palace write lock

### LINKED ISSUE #713: 🐛 Systematic lock acquisition error in mempalace-transcript.sh when the HTTP server wrapper is active
### Problem Description

When running multiple CLI agents or background tasks with the MemPalace HTTP daemon (`mempalace-http-wrapper.py`) active, the HTTP wrapper holds the exclusive write lock on the SQLite database (`~/.mempalace/palace`).

When a transcript event triggers `hooks/mempalace-transcript.sh`, the hook spawns a separate python subprocess which imports `mempalace.mcp_server` and calls `tool_add_drawer`. 
Even though the hook monkey-patches `chromadb.PersistentClient` to route database writes over HTTP via `HttpClient` (as designed by Spec 0073/0088), the `mempalace` library internally attempts to acquire an OS-level file lock on disk via `mine_palace_lock(palace_path)` inside `ChromaCollection` write methods (specifically `mempalace/palace.py` around line 1090).

Because the HTTP daemon process is already holding that exact file lock, the transcript hook's subprocess fails to acquire it, throwing a systematic `ADD_FAILED: palace is held by PID` error on every turn and completely halting the persistence of transcripts.

### Proposed Solution

An elegant way to resolve this systematic lock acquisition error is to mock/monkey-patch `mempalace.palace.mine_palace_lock` to be a no-op context manager inside the Python script of `hooks/mempalace-transcript.sh` before calling `tool_add_drawer`.

This is completely safe and correct since:
1. `chromadb.PersistentClient` is already patched to route operations over HTTP (port 8001).
2. The HTTP daemon itself already maintains the exclusive file lock on disk and can process incoming HTTP requests safely.

#### Example code:

```python
import contextlib
try:
    import mempalace.palace
    @contextlib.contextmanager
    def _mock_mine_palace_lock(palace_path):
        yield
    mempalace.palace.mine_palace_lock = _mock_mine_palace_lock
except ImportError:
    pass

try:
    from mempalace.mcp_server import tool_add_drawer
except ImportError as e:
    ...
```

hcross:
The transcript hook's per-entry subprocess took the on-disk palace write lock even
though every write it performs already leaves the process over HTTP to the shared
ChromaDB daemon, so any concurrent writer made every entry fail with
`ADD_FAILED: palace ... is held by PID ...` and silently ended transcript
persistence for the rest of the session. This relieves that lock inside the hook's
own subprocess only, strictly after the daemon-reachability probe, and refuses to
persist with a distinct exit status when it cannot prove the relief is in force on
the write path the entry actually takes.

Implements `specs/0110-transcript-hook-lock-bypass.md` (spec-PR #714, merged as
`e834378`). Closes #713.

## How to read this PR?

Two files, `131` insertions / `1` deletion in the hook and `666` insertions in the
test suite (`git diff --numstat crewrig/main...HEAD`). Read in this order:

1. **`specs/0110-transcript-hook-lock-bypass.md`** (already on `main`) — the seven
   requirements. Every code comment in the diff cites the requirement it serves, so
   the spec is the index to the change.
2. **`hooks/mempalace-transcript.sh`** — the load-bearing file. The change is one
   contiguous block inside the existing Python heredoc, opening at the
   `--- spec 0110: relieve the palace write lock, for THIS subprocess only ---`
   banner. The only other edits are the exit-code table added to the script header
   comment and the single deleted line (`import os, sys` → `import contextlib, os, sys`).
3. **`scripts/tests/test-mempalace-transcript-hook.sh`** — the new section opens at
   the `spec 0110 / issue #713` banner comment; everything above it is untouched
   (`666` insertions, `0` deletions). The banner documents the fixture topology and
   the three lock-resolution modes before the first test that uses them.

### Where the lock is actually taken

The anchor issue's own text put the acquisition at `mempalace/palace.py:1090`. That
line is only the `def`; the corrected diagnosis is in logbook comment
[`5190274518`](https://github.com/crewrig/crewrig/issues/713#issuecomment-5190274518).
Verified against the installed library (`mempalace/backends/chroma.py`):

```
tool_add_drawer                                             # mcp_server.py
  -> ChromaCollection(raw, palace_path=_config.palace_path) # mcp_server.py:1177, :1188
    -> _write_lock()                                        # backends/chroma.py:1339
      -> from ..palace import mine_palace_lock              # backends/chroma.py:1348 — LATE import
        -> with mine_palace_lock(self._palace_path)         # backends/chroma.py:1350
```

That the import at `:1348` is *late* is the whole reason a monkey-patch works at
all. Three sibling modules in the same library bind the symbol at module load
instead — `sync.py:24`, `miner.py:36`, `convo_miner.py:33` — which is why R5
forbids assuming a single resolution site.

### Non-obvious decisions, and one trade-off the reviewer should weigh

- **The ordering is normative, not stylistic.** The relief sits *after* the
  `heartbeat()` probe's `sys.exit(4)` so a persistence attempt whose HTTP routing
  is not established keeps the lock's protection (R2). A reorder still passes every
  fixture that has a reachable daemon, so it is pinned twice: behaviourally
  (`spec-0110-r2: relief not installed before the daemon is established reachable`,
  which turns on the *absence* of an import marker) and statically on source line
  order (`spec-0110-r2: ... (source order)`).
- **Two patch targets, `hasattr`-guarded, and no attribute is ever created.**
  Rebinding a name the library does not read would install a decoy and mask a
  genuine relocation instead of surfacing it through the R3 guard.
- **The third guard stage is behavioural on purpose.** Coverage and
  attribute-identity checks only prove something about a module attribute. Stage 3
  enters `ChromaCollection._write_lock()` itself and requires that the stand-in was
  reached, which is what makes the guard a statement about the real write path. Its
  probe palace path is synthetic (`/nonexistent/crewrig-transcript-lock-relief-probe`)
  and the lock key is a hash of the resolved path, so it cannot contend with any
  real palace lock.
- **⚠️ The guard fails closed — this is a decision, not an oversight.** If a future
  MemPalace renames `ChromaCollection._write_lock` or changes
  `ChromaCollection.__init__`'s signature, stage 3 raises, the hook exits 5, and
  transcripts stop *even though the relief itself might still have been working*.
  That is what R3 mandates: the hook "SHALL decline to persist and report a failure"
  when it cannot establish the relief is in force. Reversing it to fail open would
  need a delta-spec, not a code change. Worth a reviewer's explicit agreement rather
  than a silent pass.
- **`mine_global_lock` is deliberately left unpatched.**
  `mempalace/palace.py:1213` defines `mine_global_lock = mine_palace_lock`. It is
  dead in the shipped library — `grep -rn 'mine_global_lock' <mempalace-pkg> --include='*.py'`
  returns exactly one line, that definition, and zero call sites — and it is not on
  the write path. Patching it would widen the relief past R5's "every location from
  which *the write path* resolves the lock".
- **No CLI asymmetry is introduced.** `hooks/mempalace-transcript.sh` is the single
  shared script all four hook manifests invoke, and the relief lives inside its
  Python heredoc, so all four CLIs get identical behaviour. `docs/cli-matrix.md`
  rows 8 and 9 describe the per-CLI manifests and the project-dir env var; this diff
  touches neither, and the co-maintenance trigger in `AGENTS.md` names
  `hooks/*-transcript-hooks.json`, not this script.

## How to test this PR?

**Prerequisites:** `bash`, `python3`, `jq`. No MemPalace install and no running
ChromaDB daemon are needed — the spec-0110 tests run the real shipped heredoc
against a mocked `mempalace` package that reproduces the shipped lock topology
with a genuine `fcntl.flock`.

**The palace is not touched.** Every fixture takes its lock directory from
`$CREWRIG_TEST_LOCK_DIR` and its palace path from `$CREWRIG_TEST_PALACE`, both
inside a per-test `mktemp -d` sandbox, and the lock key is a hash of the resolved
palace path. Verified on this branch: `~/.mempalace/locks` held `38` entries before
the run and `38` after, with an identical directory listing (`shasum` of the sorted
listing unchanged).

### Golden path — the full suite

```sh
bash scripts/tests/test-mempalace-transcript-hook.sh; echo "exit=$?"
```

Observed on this branch under `/bin/bash` 3.2.57(1)-release (arm64-apple-darwin25),
7s wall clock:

```
Summary: 25 passed, 0 failed, 0 skipped
exit=0
```

`25` = `13` pre-existing assertions (issues #90–#94, spec 0073, spec 0074, spec
0088) + `12` new spec-0110 assertions. The suite is already wired into all three CI
surfaces, so no wiring change was needed: `.github/workflows/build.yml:128`,
`.gitlab-ci.yml:66`, `ci/ci-capabilities.yml:75`.

### Failure mode 1 — the defect itself, on demand

Test `spec-0110-r1: fixture reproduces the defect without the relief (control)` is
the two-sided half that keeps the R1 test from passing vacuously: it replays the
same mocked write path *without* the hook's relief while a peer holds the lock, and
must be refused. If that control ever stops being refused, the fixture has stopped
reproducing the bug and the neighbouring R1 assertion means nothing.

### Failure mode 2 — prove the new assertions can actually fail

Mutate a scratch copy of the hook (the test resolves `HOOK` relative to its own
location, so no repo file is touched):

```sh
MUT=$(mktemp -d); mkdir -p "$MUT/hooks" "$MUT/scripts/tests"
cp scripts/tests/test-mempalace-transcript-hook.sh "$MUT/scripts/tests/"
cp hooks/mempalace-transcript.sh "$MUT/hooks/"
# Drop the second patch target — a single-location relief.
python3 - "$MUT/hooks/mempalace-transcript.sh" <<'PY'
import sys; p=sys.argv[1]; s=open(p).read()
open(p,"w").write(s.replace("for _mp_mod in (_mp_palace, _mp_chroma):\n    if hasattr",
                            "for _mp_mod in (_mp_palace,):\n    if hasattr", 1))
PY
bash "$MUT/scripts/tests/test-mempalace-transcript-hook.sh"
```

Expected: `Summary: 24 passed, 1 failed`, the failure being
`spec-0110-r5: relief covers a module-load resolution site (not one location)`. Two
further mutants and their observed kills are tabulated in the detailed section
below.

### `hooks/mempalace-transcript.sh` — +131 / -1

**Modified — header comment.** Added an *Inner Python payload exit codes* table
documenting `2 IMPORT_ERROR:`, `3 ADD_FAILED:`, `4 DAEMON_UNREACHABLE:`, and the
new `5 LOCK_BYPASS_INEFFECTIVE:`, plus the standing note that the hook itself
always exits `0`. The three pre-existing codes were previously only discoverable by
reading the heredoc.

**Modified — the single deleted line.** `import os, sys` → `import contextlib, os, sys`.
`contextlib` is needed for the stand-in context manager.

**Added — the relief block**, inserted between the heartbeat probe's `sys.exit(4)`
and the `from mempalace.mcp_server import tool_add_drawer` that follows. In order:

1. `_lock_relief_hits` list + `_relieved_palace_lock` — a `@contextlib.contextmanager`
   stand-in for `mine_palace_lock` that records the palace path it was called with
   and yields without locking. The recording list is what stage 3 of the guard reads.
2. `import mempalace.palace` / `import mempalace.backends.chroma`, wrapped in
   `try/except ImportError` reporting the pre-existing `IMPORT_ERROR:`/`2`. This is
   the first point at which the hook loads the locking library at all — which is
   precisely what the R2 ordering test detects by absence.
3. The R5 rebind loop over `(_mp_palace, _mp_chroma)`, guarded by
   `hasattr(_mp_mod, "mine_palace_lock")`, accumulating `_relieved_modules`.
4. The three-stage R3/R4 guard, cheapest first: **coverage** (`_relieved_modules`
   non-empty), **identity** (every location that exposed the symbol now *is* the
   stand-in), **behavioural probe** (construct a `ChromaCollection` on the synthetic
   probe path, enter `_write_lock()`, require `len(_lock_relief_hits)` to have
   grown). Any failure sets `_relief_error`, prints `LOCK_BYPASS_INEFFECTIVE: <why>`
   to stderr, and exits `5` before `tool_add_drawer` is imported — so an
   unprovable relief cannot write.
5. The probe's `except Exception` carries an `acknowledged-exception:` comment
   explaining why narrowing it would forfeit the distinguishability R4 mandates:
   every failure mode (renamed `_write_lock`, changed `__init__` signature, a real
   lock acquired and raising `MineAlreadyRunning`) answers the same question — "is
   the relief in force?" — with "no", and must take the identical decline-and-report
   path rather than escape as a bare traceback.

R6 needs no code: the relief is an in-memory attribute rebind in a subprocess the
hook launches for one entry. It writes no file, exports no variable, and neither
releases nor removes any lock another process holds. Test `spec-0110-r6` proves
that empirically rather than by assertion.

### `scripts/tests/test-mempalace-transcript-hook.sh` — +666 / -0

`12` new assertions, appended as one section after the existing body; no existing
line was modified except the file-header comment block, which gained a spec-0110
paragraph. They run the **real shipped heredoc**, extracted from the hook with the
same `sed -n "/<<.PYEOF./,/^PYEOF$/p"` technique the pre-existing spec-0073 test
already uses — so the tests cannot drift from a paraphrase of the source.

**Fixture builders added:** `mp_sandbox <dir> <mode>` materialises a mock `chromadb`
(reachable daemon; `PersistentClient` remains a poison pill so the relief cannot
become an excuse to touch the on-disk palace) and a mock `mempalace` whose
`mine_palace_lock` is a real non-blocking `fcntl.flock` keyed by
`sha256(realpath(palace_path))`. `mp_holder_script` / `mp_wait_ready` run and
synchronise the peer that holds the lock.

**Three resolution modes**, because R5 forbids assuming one: `late` (today's shipped
shape), `modload` (the symbol bound at module load, the shape `sync.py` /
`miner.py` / `convo_miner.py` already use), `elsewhere` (a third module the hook
does not know about, bound before the patch lands — where the relief is genuinely
ineffective and the hook must say so).

**The five spec scenarios, each as an automated check:**

| Spec scenario | Assertion(s) |
|---|---|
| an entry persists while a peer holds the lock | `spec-0110-r1: fixture reproduces the defect without the relief (control)` + `spec-0110-r1: entry persists while a peer holds the palace write lock` |
| nothing changes when no peer holds the lock | `spec-0110: uncontended path unchanged — entry stored, success reported` |
| an ineffective relief is reported, not silently absorbed | `spec-0110-r3: ineffective relief declines to persist and exits 5`, `spec-0110-r4: status 5 + LOCK_BYPASS_INEFFECTIVE: collide with no existing failure`, `spec-0110-r7: hook still exits 0 and reports rc=5 like every other failure` |
| an unreachable service keeps its existing behaviour | `spec-0110-r2: relief not installed before the daemon is established reachable` + `spec-0110-r7: hook still exits 0 on the unreachable-service path` |
| a peer's lock protection is unaffected | `spec-0110-r6: a peer's lock protection is unaffected by the relief` |

The remaining three cover requirements no scenario states directly:
`spec-0110-r5: relief covers a module-load resolution site (not one location)` (the
regression a single-location patch cannot pass), plus two static locks on the
shipped source — `spec-0110-r2: ... (source order)` and
`spec-0110-r5: both known resolution sites named in the hook source`.

Two assertions are deliberately two-sided, because a one-sided version would pass
vacuously: the R1 control (above), and `spec-0110-r6`, which requires a separate
process to still **ACQUIRE** the lock after the hook persisted under the relief
*and* a third acquisition to still be **REFUSED** while that process holds it.

### Mutation evidence — the new assertions do fail when the fix is broken

Each mutant is a single edit to a scratch copy of the hook; the repo was not
modified. Observed summaries:

| Mutant | Observed result |
|---|---|
| Patch only `mempalace.palace` (drop the `backends.chroma` target) | `24 passed, 1 failed` — kills `spec-0110-r5: relief covers a module-load resolution site`. Note it dies *loudly*: `rc=5 stderr=LOCK_BYPASS_INEFFECTIVE: mempalace.backends.chroma.mine_palace_lock was not replaced` |
| Remove the behavioural write-path probe, keeping coverage + identity (`if _relief_error is None:` → `if False:`) | `23 passed, 2 failed` — kills `spec-0110-r3` and `spec-0110-r7`. The decisive observation: the ineffective-relief fixture **stored the entry** (`stored=transcripts\|room-16\|[USER] must not be stored\|transcript-hook`, `rc=0`, want `5`). Stage 3 is what stands between an unprovable relief and a silent write |
| Hoist the relief above the heartbeat probe (R2 ordering drift) | `21 passed, 4 failed` — kills both R2 assertions: behavioural (`mempalace_loaded=yes`, want `no`) and source-order (`relief installed at line 246, before the DAEMON_UNREACHABLE exit at line 252`). The other two failures are collateral from the crude hoist, which imports `mempalace` inside a fixture that provides only `chromadb` |

### For the next agent

- `specs/0110-transcript-hook-lock-bypass.md` still reads `status: approved` on
  purpose. `docs/spec-format.md` assigns the flip to `implemented` to *after* this
  PR merges; it lands as a separate metadata-only transition PR.
- The out-of-scope list in spec 0110 names the architectural alternative — extending
  the memory service so the hook records an entry without loading the library
  in-process. It is tracked separately and remains the better long-term answer; this
  PR adapts the hook to the library as shipped.
- If a future MemPalace release moves the lock resolution, the symptom will be exit
  `5` with `LOCK_BYPASS_INEFFECTIVE:` naming the location, not a silent stall. Start
  from the message, then re-run the `modload` and `elsewhere` fixtures.

### DIFF
--- a/hooks/mempalace-transcript.sh
+++ b/hooks/mempalace-transcript.sh
@@ -33,6 +33,16 @@
 #   (GitHub Copilot CLI does NOT export a $COPILOT_PROJECT_DIR — the project
 #    path is read from the hook stdin JSON payload, with $PWD as fallback.)
 #
+# Inner Python payload exit codes (surfaced by the hook as `rc=<n>` on the
+# failure log line, alongside the matching stderr prefix). The hook itself
+# always exits 0 — a failed persistence never fails the agent's turn.
+#   2  IMPORT_ERROR:            chromadb or mempalace unavailable
+#   3  ADD_FAILED:              the drawer write was refused
+#   4  DAEMON_UNREACHABLE:      the shared ChromaDB daemon did not answer
+#   5  LOCK_BYPASS_INEFFECTIVE: the spec-0110 palace-write-lock relief could
+#                               not be proven in force on the write path, so
+#                               the entry was deliberately not persisted
+#
 # Requires: jq, mempalace (Python package)
 
 set -euo pipefail
@@ -176,7 +186,7 @@ if [ -n "$CONTENT" ]; then
     TRANSCRIPT_ROOM="$TRANSCRIPT_ROOM" \
     TRANSCRIPT_AGENT="$TRANSCRIPT_AGENT" \
     ${_HOOK_TIMEOUT:+$_HOOK_TIMEOUT 5} "$MEMPALACE_PYTHON" - 2>"$_HOOK_ERR" <<'PYEOF'
-import os, sys
+import contextlib, os, sys
 
 # ADR-0006 routing: patch chromadb.PersistentClient -> HttpClient BEFORE
 # `mempalace.mcp_server` resolves the symbol (same technique as
@@ -227,6 +237,121 @@ try:
 except Exception as e:  # acknowledged-exception: broad except intentional — any HttpClient.heartbeat() failure (connection refused, DNS, protocol) means the daemon is unreachable and must soft-skip this persistence attempt (spec 0073 R3); it does not block startup like the wrapper's probe, so it must not be silently swallowed or misclassified either
     print(f"DAEMON_UNREACHABLE: {_host}:{_port} — {e}", file=sys.stderr)
     sys.exit(4)
+
+# --- spec 0110: relieve the palace write lock, for THIS subprocess only ---
+#
+# Every write this subprocess performs already leaves the process over HTTP
+# to the shared ChromaDB daemon (the `chromadb.PersistentClient`
+# substitution above, spec 0073/0088), and `tool_add_drawer` performs no
+# direct disk write of its own. The MemPalace library nevertheless has
+# `ChromaCollection._write_lock()` take the on-disk per-palace lock, so any
+# concurrent writer — a sibling agent, the memory server, a mine — made
+# every transcript entry fail with `ADD_FAILED: palace ... is held by PID
+# ...` and silently ended transcript persistence for the whole session
+# (spec 0110 R1).
+#
+# R2 — THE ORDERING BELOW IS NORMATIVE. This block sits AFTER the
+# heartbeat probe above, deliberately. A persistence attempt whose remote
+# routing is not established must keep the lock's protection, so the relief
+# must never be installed on a path that can still reach the on-disk
+# palace. Do not move it earlier.
+#
+# R6 — the relief is confined to this interpreter: it rebinds an in-memory
+# module attribute in the process this hook launched for one entry. It
+# writes no file, sets no environment variable for anyone else, and neither
+# releases nor removes a lock any other process holds. A concurrently
+# running memory server or maintenance writer keeps the exact lock it takes
+# today.
+_lock_relief_hits = []
+
+
+@contextlib.contextmanager
+def _relieved_palace_lock(palace_path, *args, **kwargs):
+    """Stand-in for `mempalace.palace.mine_palace_lock` — records, never locks."""
+    _lock_relief_hits.append(palace_path)
+    yield
+
+
+try:
+    import mempalace.backends.chroma as _mp_chroma
+    import mempalace.palace as _mp_palace
+except ImportError as e:
+    print(f"IMPORT_ERROR: {e}", file=sys.stderr)
+    sys.exit(2)
+
+# R5 — patch every location the write path can resolve the primitive from,
+# and do not assume a single one. Today `ChromaCollection._write_lock()`
+# resolves it by a LATE import from `mempalace.palace`, so the canonical
+# definition is the site that matters; but three sibling modules in the same
+# library (`sync`, `miner`, `convo_miner`) bind the symbol at module load
+# instead, and if `backends.chroma` ever joins them the canonical patch
+# alone would quietly stop taking effect. Rebind the attribute wherever it
+# already exists; never create one that does not, since that installs a name
+# the library never reads and would mask a genuine relocation instead of
+# surfacing it through the R3 guard below.
+_relieved_modules = []
+for _mp_mod in (_mp_palace, _mp_chroma):
+    if hasattr(_mp_mod, "mine_palace_lock"):
+        _mp_mod.mine_palace_lock = _relieved_palace_lock
+        _relieved_modules.append(_mp_mod.__name__)
+
+# R3/R4 — prove the relief is in force on the path the entry actually
+# takes, and decline to persist when it is not. Three stages, cheapest
+# first:
+#
+#   1. Coverage — at least one known location must have exposed the symbol.
+#      An empty list means the library relocated it entirely.
+#   2. Identity — every location that exposed it must now BE the stand-in.
+#      Side-effect free.
+#   3. Behavioural probe — enter the very method the write goes through,
+#      `ChromaCollection._write_lock()`, and require that it reached the
+#      stand-in. This is what makes the guard a statement about the real
+#      write path rather than about a module attribute: if `_write_lock` is
+#      renamed, or resolves the primitive from a location stage 1 does not
+#      cover, the stand-in is never reached and we refuse to persist.
+#
+# The probe's palace path is synthetic and is NEVER the configured palace.
+# The lock key is sha256(realpath(path)), so a synthetic path cannot contend
+# with any real palace lock; it is passed only so `_write_lock` takes its
+# locking branch instead of its documented palace_path-is-None no-op.
+#
+# The guard fails closed on purpose: an unprovable relief stops persistence
+# rather than silently reverting to the lock-contention failure this spec
+# exists to end. Distinct exit status (5) and stderr prefix
+# (LOCK_BYPASS_INEFFECTIVE:) per R4 — no other failure this hook reports
+# uses either (IMPORT_ERROR:/2, ADD_FAILED:/3, DAEMON_UNREACHABLE:/4).
+_LOCK_PROBE_PALACE = "/nonexistent/crewrig-transcript-lock-relief-probe"
+
+_relief_error = None
+if not _relieved_modules:
+    _relief_error = (
+        "no known location exposes mine_palace_lock "
+        f"(looked in {_mp_palace.__name__}, {_mp_chroma.__name__})"
+    )
+else:
+    for _mp_mod in (_mp_palace, _mp_chroma):
+        _resolved = getattr(_mp_mod, "mine_palace_lock", _relieved_palace_lock)
+        if _resolved is not _relieved_palace_lock:
+            _relief_error = f"{_mp_mod.__name__}.mine_palace_lock was not replaced"
+
+if _relief_error is None:
+    _hits_before = len(_lock_relief_hits)
+    try:
+        _probe = _mp_chroma.ChromaCollection(None, palace_path=_LOCK_PROBE_PALACE)
+        with _probe._write_lock():
+            pass
+    except Exception as e:  # acknowledged-exception: broad except intentional — the probe's only job is to answer "is the relief in force on the write path"; every failure mode (renamed/removed _write_lock, changed ChromaCollection signature, a real lock acquired and raising MineAlreadyRunning) answers "no" and takes the identical R3 decline-and-report path, so narrowing the catch would let an unanticipated type escape as a bare traceback and forfeit exactly the distinguishability R4 mandates
+        _relief_error = f"write-path probe raised {type(e).__name__}: {e}"
+    else:
+        if len(_lock_relief_hits) == _hits_before:
+            _relief_error = (
+                "ChromaCollection._write_lock() did not resolve the relieved "
+                f"lock (patched: {', '.join(_relieved_modules)})"
+            )
+
+if _relief_error is not None:
+    print(f"LOCK_BYPASS_INEFFECTIVE: {_relief_error}", file=sys.stderr)
+    sys.exit(5)
 
 try:
     from mempalace.mcp_server import tool_add_drawer
--- a/hooks/mempalace-transcript.sh
+++ b/hooks/mempalace-transcript.sh
@@ -240,10 +240,15 @@ except Exception as e:  # acknowledged-exception: broad except intentional — a
 
 # --- spec 0110: relieve the palace write lock, for THIS subprocess only ---
 #
-# Every write this subprocess performs already leaves the process over HTTP
-# to the shared ChromaDB daemon (the `chromadb.PersistentClient`
-# substitution above, spec 0073/0088), and `tool_add_drawer` performs no
-# direct disk write of its own. The MemPalace library nevertheless has
+# Every palace write this subprocess performs already leaves the process
+# over HTTP to the shared ChromaDB daemon (the `chromadb.PersistentClient`
+# substitution above, spec 0073/0088). `tool_add_drawer` does perform one
+# local disk write of its own — `_wal_log` appends to the write-ahead log
+# (`mcp_server._wal_log` -> `wal.py`) — but that append sits OUTSIDE
+# `ChromaCollection._write_lock()`, so the palace lock never guarded it and
+# relieving the lock leaves its exposure exactly as it already was. The
+# relief therefore opens no race that did not already exist. The MemPalace
+# library nevertheless has
 # `ChromaCollection._write_lock()` take the on-disk per-palace lock, so any
 # concurrent writer — a sibling agent, the memory server, a mine — made
 # every transcript entry fail with `ADD_FAILED: palace ... is held by PID
--- a/scripts/tests/test-mempalace-transcript-hook.sh
+++ b/scripts/tests/test-mempalace-transcript-hook.sh
@@ -39,6 +39,16 @@
 #         `DAEMON_UNREACHABLE:` on stderr) when the daemon is unreachable —
 #         never falling back to `PersistentClient`.
 #
+#   spec 0110 / issue #713 — Survive a concurrently held palace write lock.
+#         The heredoc MUST neutralise `mine_palace_lock` in its own
+#         subprocess so an entry persists while a peer holds the palace
+#         write lock, at EVERY location the write path resolves that
+#         primitive from, only AFTER the daemon is established reachable,
+#         and MUST refuse to persist with a distinct status (5) and prefix
+#         (`LOCK_BYPASS_INEFFECTIVE:`) when it cannot prove the relief is in
+#         force. See the section header before test 12 for the full topology
+#         and the three resolution modes exercised.
+#
 # Usage:
 #   bash scripts/tests/test-mempalace-transcript-hook.sh
 #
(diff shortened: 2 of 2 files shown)


hcross:
## Verdict: APPROVE

Cold review — no authoring context. Every factual claim below was re-derived from
the PR head (`bc02a7a`), the merged spec, and the `mempalace 3.6.0` package
actually installed on this machine
(`~/.local/pipx/venvs/mempalace/lib/python3.13/site-packages/mempalace`).

**No formal `APPROVE` review event was postable.** `gh api user` resolves to
`hcross`, which is also this PR's author, and the GitHub API refuses a
self-approval (`Can not approve your own pull request`) — the single-identity
constraint tracked as #692. This plain comment is therefore the canonical verdict
artifact, per `pr-reviewer` step 6 rung 2. It is a genuine APPROVE, not a
COMMENT-event review relabelled as one; the distinction is that GitHub's review
tab will show zero approvals, so a merge cannot cite a green review checkmark.

---

### CI status

All **12/12** required checks pass on `bc02a7a` (`gh pr checks 715`): `build`,
`check-agents-size`, `check-ci-parity`, `check-components`,
`check-extension-version-bump`, `check-feedback-routing`, `check-skill-versions`,
`gitlab-ci-check`, `grep-anti-patterns`, `lint-markdown`, `lint-specs`,
`test-harness-curate`. No failing job, no pending job. Preflight clear.

### Correctness

**The library claims in the PR body are accurate.** I checked each one against the
installed package rather than accepting it:

| Claim | Verified |
|---|---|
| Issue #713's `palace.py:1090` is only the `def` | ✅ `palace.py:1090` = `def mine_palace_lock(palace_path: str):` — the acquisition is in the body, so the PR's correction of its own anchor issue is right |
| `backends/chroma.py:1339` `_write_lock` | ✅ `def _write_lock(self):` |
| `backends/chroma.py:1348` **late** import | ✅ `from ..palace import mine_palace_lock`, inside the function body |
| `backends/chroma.py:1350` acquisition | ✅ `with mine_palace_lock(self._palace_path):` |
| `sync.py:24`, `miner.py:36`, `convo_miner.py:33` bind at module load | ✅ all three |
| `palace.py:1213` `mine_global_lock = mine_palace_lock`, zero call sites | ✅ package-wide grep returns exactly that one line |

**The write-path diagram is complete.** I specifically hunted for an acquisition
the diagram omits, and found `mcp_server.py:443–445`
(`lock_cm = mine_palace_lock(_config.palace_path)`). It is **not** a gap: its only
caller is `_mcp_peer_writer_refusal()` (`mcp_server.py:474`), reached from the
JSON-RPC dispatch guard. The hook calls `tool_add_drawer` directly as a Python
function and never enters that dispatch, so it is off the hook's path — and were
it ever on it, its import at `:443` is late from `.palace`, which the relief
already covers. Correctly omitted.

**The probe matches the real constructor.**
`ChromaCollection.__init__(self, collection, palace_path: Optional[str] = None)`
(`chroma.py:1334`), so `ChromaCollection(None, palace_path=...)` is well-formed,
and `_write_lock`'s documented `palace_path is None` no-op is correctly avoided by
passing a non-`None` synthetic path.

**R5's `hasattr` guard behaves correctly against the shipped library.**
`mempalace.backends.chroma` exposes no module-level `mine_palace_lock` (the import
at `:1348` is function-local), so in production `_relieved_modules` is
`['mempalace.palace']` alone and the chroma target is skipped without creating a
decoy attribute — exactly as the PR describes. The late import then resolves the
patched symbol at call time. Ordering also means any module-load binding in a
module imported *after* the patch (`mcp_server` included) picks up the stand-in
automatically — a nice consequence that makes the relief more robust than R5
strictly requires.

**R2 ordering re-derived independently** from the head-ref source:
`sys.exit(4)` at line 239 → relief block opens at 241 → first `mempalace` import
at 277 → `sys.exit(5)` at 354 → `tool_add_drawer` imported at 357. The relief
cannot be installed on a path whose HTTP routing is unestablished, and an
unprovable relief cannot reach the writer. Correct.

#### Finding 1 — the safety rationale rests on a false premise (non-blocking)

`class: tech`

`hooks/mempalace-transcript.sh`, in the spec-0110 banner comment:

> Every write this subprocess performs already leaves the process over HTTP to the
> shared ChromaDB daemon […] and `tool_add_drawer` performs no direct disk write
> of its own.

Both halves are false. `tool_add_drawer` calls `_wal_log(...)` at
`mcp_server.py:2562`, which appends to `~/.mempalace/wal/write_log.jsonl` via
`os.open(..., O_WRONLY|O_APPEND|O_CREAT, 0o600)` (`wal.py:74`, `wal.py:96–98`).
That is a local disk write, and it is not hypothetical — on this machine the file
is 7.9 MB and its three most recent entries are `add_drawer` records with
`added_by=transcript-hook`, i.e. produced by this exact code path.

**This is not a correctness defect and does not block merge.** The relief remains
safe, for a reason the comment does not give:

- `_wal_log` runs at `:2562`, *before* `col.upsert()` (`:2608`) and entirely
  outside `ChromaCollection._write_lock()`. The palace write lock never protected
  the WAL append, so relieving that lock changes nothing about it.
- The append is `O_APPEND` — the designed concurrent-appender mode — and
  `_wal_log` swallows every exception (`wal.py:99–100`), so it is non-fatal by
  construction.
- Everything else in `tool_add_drawer` is `col.upsert` / `col.get`, i.e. through
  `ChromaCollection`, i.e. over HTTP. The comment's *conclusion* is right.

Why fix it anyway: this comment is the documented justification a future
maintainer will consult when deciding whether the bypass is still safe. A
categorical "no direct disk write" invites the wrong answer if upstream ever moves
`_wal_log` inside `_write_lock`. Suggested narrowing:

> Every write the palace lock protects already leaves the process over HTTP to the
> shared ChromaDB daemon (the `chromadb.PersistentClient` substitution above, spec
> 0073/0088). The one direct disk write `tool_add_drawer` performs — `_wal_log`'s
> append to `~/.mempalace/wal/write_log.jsonl` (`mcp_server.py:2562`) — is taken
> outside `ChromaCollection._write_lock()` and is unaffected by the relief.

Comment-only; no behavioural change, no test change.

### The fail-closed decision — explicit agreement, as requested

**I agree with fail-closed.** Not as a silent pass; here is the reasoning, and the
strongest objection I could build against it.

Four arguments for:

1. **R3 mandates it.** The requirement is that the hook "SHALL establish that the
   relief […] is in force" and "SHALL decline to persist" otherwise. An
   inconclusive probe is not an establishment. Reading R3 as satisfied by "I could
   not disprove it" empties the requirement. The spec is merged; reversing this is
   a delta-spec, not a code change — the PR is right about that too.
2. **The asymmetry of cost is decisive.** Transcripts are a best-effort
   observability side-channel, and R7 keeps the agent's turn green (hook exits 0)
   regardless. So fail-closed costs "no transcripts, plus a named diagnostic" —
   not a broken session.
3. **Fail-open is the worse failure, not the safer one.** If the relief is
   genuinely broken and the hook proceeds, the write hits the real lock and dies
   with `ADD_FAILED: palace ... is held by PID ...` — precisely the generic,
   cause-free message spec 0110's Intent section exists to eliminate. Fail-open
   would reinstate the original silent-loss bug with an extra layer of machinery
   in between, making the *next* diagnosis harder than the one that produced this
   PR.
4. **It converts an unbounded diagnosis into one line.** Exit 5 plus
   `LOCK_BYPASS_INEFFECTIVE: <location>` names the failing site.

The objection, stated fairly: **stage 3 tests a broader proposition than it
claims.** It is titled "is the relief in force", but it also silently asserts that
`ChromaCollection` still constructs as `(collection, palace_path=...)`. A future
upstream adding a required third `__init__` parameter would raise `TypeError`,
trip exit 5, and stop transcripts while the relief on `mine_palace_lock` is
perfectly intact — a false negative caused by coupling the check to a constructor
signature it does not actually depend on.

I still land on agree, for two reasons. First, under R3 the two cases collapse:
"the probe could not run" and "the relief is not in force" both fail the predicate
"is established in force", and preferring persistence on an unproven relief is
exactly the judgement call R3 removes from the implementation. Second — and this
defuses most of the objection's sting — the diagnostic **already distinguishes**
the two sub-cases at the operator's eye level: a constructor break reports
`write-path probe raised TypeError: ...`, whereas a real relocation reports
`ChromaCollection._write_lock() did not resolve the relieved lock`. The operator
therefore knows immediately whether to look at the constructor or at the lock
primitive. That is the part that actually matters, and it is present.

No change requested here. Recording the objection so a future delta-spec, if one
is ever raised, starts from a stated trade-off rather than rediscovering it.

### Convention compliance

- **Gitmoji title** — `🐛 Persist transcripts while a peer holds the palace write lock`. ✅
- **PR body template** (`docs/pr-format.md`) — purpose, *How to read this PR?*,
  *How to test this PR?*, *Detailed description (for agents)* all present. ✅
- **Branch naming** — `feat/0713-transcript-hook-lock-bypass` matches
  `<prefix>/<NNNN>-<slug>`. ✅
- **Spec-PR ordering** — `specs/0110-...md` is on `main` and its commit is an
  ancestor of the PR head (`git merge-base --is-ancestor`). Implementation lands
  after the spec, per `docs/spec-pr-workflow.md`. ✅
- **Spec status left at `approved`** — correct. The flip to `implemented` is a
  separate metadata-only PR, matching the pattern just used for spec 0109
  (#711, `e919939`). `lint-specs` passes, so the 0109 truth-enforcement guard is
  satisfied. ✅
- **Close keyword** — `gh pr view --json closingIssuesReferences` returns exactly
  `[713]`. Correct for an implementation PR closing its anchor issue (Rule C); no
  stray spec-PR close. ✅
- **Logbook** — #713 is its own logbook per Rule A (3 incremental comments,
  including the diagnosis correction `5190274518`, which I confirmed exists and
  matches the PR's account). ✅
- **CLI-matrix co-maintenance** — the `AGENTS.md:242` trigger glob is
  `hooks/*-transcript-hooks.json`. This PR touches `hooks/mempalace-transcript.sh`
  and `scripts/tests/test-...sh`; neither matches that glob, nor
  `scripts/{build,install,setup,import,manage}-*.sh`. Rows 8 and 9 of
  `docs/cli-matrix.md` cover the per-CLI manifests and the project-dir env var —
  untouched here. No update owed, and no CLI asymmetry: the change is inside the
  one shared script all four manifests invoke, so parity holds by construction. ✅
- **Version-bump convention** — no skill/agent source in the diff; rule N/A, and
  `check-skill-versions` passes. ✅

### Test coverage

I ran the suite at the PR head in a clean worktree rather than trusting the body:

```
Summary: 25 passed, 0 failed, 0 skipped     (~7s)
```

Matching the claimed 13 pre-existing + 12 new spec-0110 assertions. Every spec 0110
requirement maps to at least one automated check, and R1/R6 are two-sided so they
cannot pass vacuously.

**Mutation evidence reproduces.** I re-ran mutant 1 (drop the `_mp_chroma` patch
target) against a scratch copy:

```
FAIL  spec-0110-r5: relief covers a module-load resolution site (not one location)
      — rc=5 stderr=LOCK_BYPASS_INEFFECTIVE: mempalace.backends.chroma.mine_palace_lock was not replaced
Summary: 24 passed, 1 failed, 0 skipped
```

Exactly the summary and the stderr the PR body predicts. The new assertions
genuinely fail when the fix is broken — this is the evidence that separates real
coverage from coverage theater, and it holds.

**The palace really is untouched.** `~/.mempalace/locks` holds 38 entries before
and after the run, matching the body's count. I also derived the probe path's lock
key independently — `sha256(realpath("/nonexistent/crewrig-transcript-lock-relief-probe"))[:16]`
→ `mine_palace_1c9e270e2a7c8667.lock` — and confirmed no such file exists in the
real lock directory. The synthetic probe path provably cannot contend with a real
palace lock.

### Linter findings

`lint-shell.sh` (shellcheck) on both changed files — exit 1, all findings
`info`/`warning` level, none introduced as a new *class* of problem:

- `hooks/mempalace-transcript.sh:113,134` — `SC2034` (`TOOL_INPUT`, `STOP_REASON`
  appear unused). **Pre-existing**; both lines are outside this diff.
- `scripts/tests/test-mempalace-transcript-hook.sh` — `SC2030`/`SC2031` on
  `export`s inside `( … )` groups, including new lines 945–950 and 1021–1026.
  These are **deliberate test isolation**: each scenario runs in a subshell so its
  environment cannot leak into the next, which is precisely what SC2030/SC2031
  flag. The new blocks follow the identical pattern already used at lines 131–132,
  212–213 and 363–365. False positives here.
- `SC2016` on single-quoted grep patterns — intentional; the `$` is regex, not
  expansion.

No new suppressible defect. `lint-markdown.sh`, `lint-python.sh`, `lint-json.sh`,
`lint-skill.sh` — not applicable, no matching files in the diff.

---

### Summary

One non-blocking `class: tech` finding: a rationale comment whose stated premise
("no direct disk write") is factually false, though the conclusion it supports is
correct. Everything else — the library diagnosis, the R2 ordering, the R5 coverage
model, the fail-closed guard, the mutation-verified tests, and every project
convention — holds up under independent verification. The fail-closed design has
my explicit agreement, with the constructor-coupling objection recorded above.

Merging with finding 1 unaddressed is defensible; addressing it is a
comment-only edit and would leave the next maintainer with an accurate premise.

hcross:
## Verdict: APPROVE

REVIEW iteration 2 — cold pass, no authoring context. Head reviewed:
**`a214005ecd5b107de7d84a14586e7c578592f49f`**.

One **non-blocking** `class: tech` finding (a stale number in the PR body, not in
the code). Nothing blocks merge. Because the finding is non-zero, the ADR-0010
termination criterion (*APPROVE + zero findings of any class + green CI*) is not
yet met; the remedy is a two-number PR-body edit that touches no file and needs no
CI re-run.

*Posting note (issue #692):* this repository runs under a single GitHub identity
(`gh api user` → `hcross`, PR author → `hcross`), so the API refuses to record a
formal `APPROVE` review event on it — `Can not approve your own pull request`. This
is a plain comment, and it is the canonical verdict artifact for the shared-identity
case. It is **not** a review-event approval dressed up as one, and no other agent
was asked to post it.

---

### CI status

All **12** required checks **pass** at `a214005`
(`gh pr checks 715 --repo crewrig/crewrig`, exit 0): `build`,
`check-agents-size`, `check-ci-parity`, `check-components`,
`check-extension-version-bump`, `check-feedback-routing`, `check-skill-versions`,
`gitlab-ci-check`, `grep-anti-patterns`, `lint-markdown`, `lint-specs`,
`test-harness-curate`. Zero failing, zero pending. `mergeStateStatus: CLEAN`.

---

### Iteration-1 follow-up — verified, not assumed

**1. Is the corrected rationale accurate?** Yes, on every clause. Checked against
the installed package the hook actually runs
(`MEMPALACE_PYTHON=~/.local/pipx/venvs/mempalace/bin/python`, package at
`…/lib/python3.13/site-packages/mempalace`):

| Clause in the new comment | Verification |
|---|---|
| `tool_add_drawer` does perform one local disk write (`_wal_log` → `wal.py`) | `mcp_server.py:2562` calls `_wal_log` inside `tool_add_drawer` (`def` at `:2530`); `wal.py:93-95` opens the WAL `O_WRONLY\|O_APPEND\|O_CREAT` and writes a line. Real disk write — confirmed. |
| That append sits **OUTSIDE** `ChromaCollection._write_lock()` | `_wal_log` is called at `mcp_server.py:2562` with no enclosing `with`. `_write_lock()` is entered only *inside* the collection's write methods (`backends/chroma.py:1426`, `:1439`, `:1459`, `:1604`), i.e. in later frames. The append precedes them. Confirmed. |
| So the palace lock never guarded it | `wal.py` imports no locking primitive at all — no `flock`, no `mine_palace_lock` (`grep -n "import\|flock\|lock" wal.py`). It never was guarded. Confirmed. |
| Relieving the lock leaves its exposure exactly as it was → opens no race that did not already exist | Follows from the two rows above. Independently: the append is a single `write()` of one line under `O_APPEND`, so its concurrency posture is unchanged by the relief. Confirmed. |

The load-bearing claim — *the WAL append sits outside `_write_lock()`* — holds. The
fix did not replace one false premise with another.

**2. Is `a214005` comment-only?** Yes, proven mechanically rather than by reading.
Stripping comments and blank lines from `hooks/mempalace-transcript.sh` at both
`bc02a7a` and `a214005` yields **178 code lines each, byte-identical**
(`diff -q` clean). `git diff --numstat bc02a7a a214005` touches one file, `9/4`,
every changed line `#`-prefixed. The test file is untouched by this commit. No
behavioural change smuggled in.

**3. Does the suite pass at head?** Yes. Run from a detached worktree at exactly
`a214005`, under `/bin/bash` 3.2.57(1)-release (arm64-apple-darwin25):
`Summary: 25 passed, 0 failed, 0 skipped`. The palace was not touched —
`~/.mempalace/locks` held `38` entries before and `38` after.

---

### Correctness — full pass against spec 0110

**The relief works against the real shipped library, not only the mock.** This is
the one thing the test suite structurally cannot prove (it runs against a mocked
`mempalace`), so I reproduced the hook's exact probe under the real installed
package. Result: `patched: ['mempalace.palace']`, probe constructed, `_write_lock()`
entered, `_lock_relief_hits` grew to
`['/nonexistent/crewrig-transcript-lock-relief-probe']`.

Two things fall out of that run, both reassuring:

- `ChromaCollection.__init__(self, collection, palace_path=None)`
  (`backends/chroma.py:1334`) is a two-line field assignment with no validation, so
  the probe's `ChromaCollection(None, palace_path=…)` is safe against the real class
  — it does **not** fail closed in production while passing in the mock. This was my
  main correctness worry and it is cleared empirically.
- Only `mempalace.palace` is patched in reality, because `backends.chroma` resolves
  the primitive by a *late* import (`:1348`) and exposes no module attribute — so
  `hasattr(_mp_chroma, "mine_palace_lock")` is `False` today. This is exactly why the
  `hasattr` guard and the `getattr(..., default=_relieved_palace_lock)` in the
  identity stage are correct rather than sloppy: a module that never exposed the
  symbol is legitimately skipped and legitimately passes stage 2.

**Requirement-by-requirement.** All 7 satisfied; hook line numbers are from the file
at head ref.

| R | Verdict | Evidence |
|---|---|---|
| R1 persists while a peer holds the lock | ✅ | `_relieved_palace_lock` stand-in records and yields without locking; rebind loop replaces the primitive. Confirmed live against the real library (above). |
| R2 relief only after reachability established | ✅ | The block opens after the `heartbeat()` probe's `sys.exit(4)` — the relief's `import mempalace.palace` is the first load of the locking library at all, so an unreachable daemon exits at 4 before it. Pinned twice in tests: behaviourally (absence of an import marker) and on source line order. |
| R3 establish relief in force, else decline | ✅ | Three-stage guard: coverage → identity → behavioural probe entering the real `ChromaCollection._write_lock()` and requiring `len(_lock_relief_hits)` to grow. Sets `_relief_error`, exits 5 **before** `tool_add_drawer` is imported, so an unprovable relief cannot write. |
| R4 distinguishable status + prefix | ✅ | Exit `5` / `LOCK_BYPASS_INEFFECTIVE:`; the pre-existing set is `IMPORT_ERROR:/2`, `ADD_FAILED:/3`, `DAEMON_UNREACHABLE:/4`. No collision. Now documented in the header exit-code table (hook `:36-44`), which matches the heredoc's actual `sys.exit` calls. |
| R5 every write-path resolution site, no single-site assumption | ✅ | See the completeness audit below. |
| R6 confined to the one subprocess | ✅ | In-memory attribute rebind in the per-entry subprocess; writes no file, exports no variable, releases no foreign lock. Test `spec-0110-r6` proves a peer can still **acquire** afterwards *and* a third acquisition is still **refused** — two-sided, so it cannot pass vacuously. |
| R7 hook always exits 0 | ✅ | `set +e` brackets the Python call (hook `:181`), `STATUS_RC=$?` (`:379`), `set -e` restored, script ends `exit 0` (`:393`). rc=5 surfaces on the same generic failure line as every other code. |

**R5 completeness audit — I went looking for a missed lock site and found none.**
Grepping the whole installed package for lock resolution turned up three candidates
beyond `_write_lock`, and all three are correctly out of the relief's scope:

- **`mcp_server.py:443-445`, `_acquire_mcp_writer_lock`** — a second genuine
  `mine_palace_lock(...)` acquisition. Not a gap: it is the MCP server's long-lived
  *writer lease*, which spec 0110 → *Out of scope* excludes by name, and it is not on
  `tool_add_drawer`'s path. Note it is also a **late** import from `.palace`, so it
  would resolve the relieved symbol anyway if ever reached.
- **`palace.py:722 mine_lock(source_file)`** — a *different* primitive (per-source-file,
  different signature), not an alias. Imported by `palace_graph.py:35`. Correctly
  untouched.
- **`palace.py:1213 mine_global_lock = mine_palace_lock`** — **independently
  confirmed dead**, as the PR body claims: `grep -rn 'mine_global_lock' <pkg>
  --include='*.py'` returns exactly that one definition line and **zero** call sites.
  Off the write path, so patching it would widen the relief past R5's *"every location
  from which the write path resolves the lock"*. Leaving it unpatched is right. Worth
  recording *why* this is safe rather than merely unused: because the alias was bound
  at module load, the rebind of `mempalace.palace.mine_palace_lock` does **not**
  follow it — so if a future release ever routed the write path through
  `mine_global_lock`, the stand-in would not be reached, stage 3 would see
  `_lock_relief_hits` flat, and the hook would exit 5. The gap fails closed and loud,
  which is the correct posture.

**The fail-closed efficacy guard — my own view, concurring.** I read R3 as literally
mandating this: *"SHALL establish that the relief … is in force … and SHALL decline to
persist and report a failure when it is not."* A probe that cannot be constructed has
not established anything, so declining is compliance, not over-reach. Iteration 1's
recorded counter-argument — that coupling to `ChromaCollection.__init__` means a future
required third parameter would stop transcripts while the relief is intact — is a real
residual risk, but I do not think it warrants a change, for three reasons:

1. The failure is **loud and diagnosable**, not silent: exit 5 with
   `write-path probe raised TypeError: …` naming the cause. Spec 0110 → *Intent* asks
   precisely that the operator "learns that specifically."
2. The signature it couples to is minimal and public-ish
   (`(collection, palace_path=None)`), and I verified it is a bare field assignment.
3. Every alternative probe that genuinely *enters* the write path couples to some
   private internal instead. The obvious one — calling `_write_lock` unbound with a
   duck-typed `self` — trades a dependency on the `palace_path` keyword for a
   dependency on the private `_palace_path` attribute name, which is not obviously more
   stable. Static inspection would be weaker still: it proves something about a module
   attribute, not about reach.

So: no finding here. Reversing to fail-open would require a delta-spec against R3, as
the PR body says, and I would not support one on this evidence. Recording explicit
reviewer agreement, which is what the author asked for rather than a silent pass.

---

### Convention compliance

- **Gitmoji** — both commit subjects and the PR title lead with an emoji
  (`🐛 Persist transcripts…`, `📝 Correct the lock-relief safety rationale…`). No
  conventional-commit prefix. ✅
- **Branch naming** — `feat/0713-transcript-hook-lock-bypass` matches
  `<prefix>/<NNNN>-<slug>`. ✅
- **Spec status left at `approved`** — correct, and I checked this specifically because
  the enforcement landed days ago (#708/#711). `docs/spec-format.md:279-287` is explicit:
  every transition other than `draft` → `approved` — including `approved` →
  `implemented` — is recorded **after** merge by a metadata-only edit, authority
  *"Implementation team's `pr-logbook`"* (`:196`). Deferring the flip to a separate
  transition PR is the documented behaviour, not an omission. `lint-specs` passes. ✅
- **CLI-matrix co-maintenance** — not triggered, verified against the `AGENTS.md`
  trigger list: it names `hooks/*-transcript-hooks.json` (the per-CLI manifests) and
  `scripts/{build,install,setup,import,manage}-*.sh`. This PR changes
  `hooks/mempalace-transcript.sh` and `scripts/tests/test-…`, neither of which matches.
  The relief lives inside the single shared script all four manifests invoke, so all
  four CLIs get identical behaviour and no parity gap is introduced. ✅
- **Version bump** — no skill/agent/extension provenance carrier touched;
  `check-skill-versions` and `check-extension-version-bump` both pass. ✅
- **Built components** — `artifacts/` untouched, so no `build-components.sh` obligation. ✅
- **Logbook (Rule A)** — #713 is a pre-existing feature issue serving as its own
  logbook, with 4 comments including the corrected lock-path diagnosis. No duplicate
  logbook issue was opened. ✅
- **Close keyword** — `gh pr view --json closingIssuesReferences` resolves to exactly
  `#713`, the intended target. ✅
- **Language** — all repository and GitHub content in English. ✅

---

### Test coverage

**25 assertions at head, 12 of them new** (`grep -oE '"spec-0110[^"]*"' | sort -u`
→ 12), on top of 13 pre-existing. Every one of the **7 requirements** and all
**5 spec scenarios** has at least one assertion; mapping verified individually
against the run output, not taken from the PR body.

What raises this above coverage theater:

- **Tests run the real shipped heredoc**, extracted from the hook with the same
  `sed -n "/<<.PYEOF./,/^PYEOF$/p"` technique the pre-existing spec-0073 test uses, so
  they cannot drift from a paraphrase of the source.
- **All three resolution modes are genuinely exercised** — `late` (×4, lines 751, 817,
  978, 1052), `modload` (845), `elsewhere` (887) — which is what makes R5 more than an
  assertion about today's shape.
- **Two assertions are deliberately two-sided**, so neither can pass vacuously: the R1
  control replays the same path *without* the relief and must be refused, and R6
  requires a later acquisition to succeed *and* a third to be refused.
- **The mutation evidence in the PR body is the right kind of evidence** for a guard
  like this, and the second mutant is the decisive one: dropping stage 3 while keeping
  coverage + identity causes the ineffective-relief fixture to **store the entry**
  (`rc=0`, want `5`). That is a direct demonstration that stage 3 — the part coupling
  to `__init__` — is what stands between an unprovable relief and a silent write.

Suite wiring needed no change: it is already invoked by `.github/workflows/build.yml`,
`.gitlab-ci.yml`, and `ci/ci-capabilities.yml`.

---

### Linter findings

Changed files are `.sh` only, so `lint-shell.sh` is the sole applicable script
(`lint-markdown`/`lint-json`/`lint-python`/`lint-skill` have no targets — the Python
lives in a heredoc, not a `.py` file).

`lint-shell.sh hooks/mempalace-transcript.sh scripts/tests/test-mempalace-transcript-hook.sh`
reports `SC2034` on the hook (`TOOL_INPUT` line 113, `STOP_REASON` line 134) and
`SC2016`/`SC2030`/`SC2031` info items on the test file. **All pre-existing, none
introduced by this PR** — I baselined it rather than reporting it: shellcheck on
`crewrig/main`'s hook and on the head hook both yield the same `3 × SC2034` and
nothing else. Net-new shellcheck findings: **zero**.

---

### Findings

**1. `class: tech` — non-blocking. The PR body's hook diff-stat is stale after the
iteration-1 fix.**

The body states `126` insertions / `1` deletion for the hook in two places — *How to
read this PR?* (“`126` insertions / `1` deletion in the hook”) and the *Detailed
description* heading (“`hooks/mempalace-transcript.sh` — +126 / -1”) — and explicitly
cites the command `git diff --numstat crewrig/main...HEAD`. That command at head
`a214005` returns:

```
131	1	hooks/mempalace-transcript.sh
666	0	scripts/tests/test-mempalace-transcript-hook.sh
```

`126` was correct at `bc02a7a`; `a214005` added net +5 comment lines (`9/4`) and the
body was not refreshed. The `666`/`0` test figure is still accurate, as is “the single
deleted line.”

*Why report a number:* the two figures are the reader's entry point into a
126-vs-131-line review, and the body names the very command that contradicts them.
Small, but a verifiable inaccuracy in a required PR-body section rather than a matter
of taste — so I am not going to file it under nits.

*Remedy:* update both occurrences to `131`. PR-body edit only — no file changes, no
code impact, no CI re-run. Owner is `pr-logbook` rather than DEV, despite the `tech`
class; `tech` is the closest fit in the three-class taxonomy for an implementation-
artifact accuracy defect (it is neither an architectural nor a specification concern).

---

### Summary

The implementation is sound, the iteration-1 finding is genuinely fixed with an
accurate premise, and the fix is provably comment-only. I confirmed the relief works
against the **real** installed MemPalace rather than only the mock — the check the test
suite cannot make — and went looking for a missed lock resolution site without finding
one. `mine_global_lock` is independently confirmed dead and correctly left alone. I
concur with the fail-closed guard as R3-mandated and would oppose reversing it without
a delta-spec.

The single finding is a stale pair of numbers in the PR body. Merge is not blocked by
it; only the zero-findings half of the termination criterion is.

hcross:
## Verdict: APPROVE

REVIEW iteration 3 — cold pass, no authoring context. Head reviewed:
**`a214005ecd5b107de7d84a14586e7c578592f49f`**.

**Zero findings of any class.** Nothing below is tagged `tech`, `arch`, or `spec`,
because nothing rose to a finding. One observation I deliberately declined to
class as a finding is recorded at the end, with my reasoning, so a later reader
can disagree on the record.

A note on the form of this verdict: this repository runs under a single GitHub
identity (`gh api user` → `hcross`, PR author → `hcross`), so the API refuses to
record a formal `APPROVE` review event on it (`Can not approve your own pull
request`, issue #692). This is a plain issue comment, and it is the canonical
verdict artifact for the shared-identity case. It is **not** a `COMMENT`-event
review dressed up as an approval.

### Head-unchanged check

The head is `a214005`, unchanged since iteration 2 reviewed and approved it. The
branch carries exactly two commits (`bc02a7a` implementation, `a214005`
comment-only correction); `git diff a214005 --stat` against the local worktree at
that SHA is empty and the tree is clean. So the code surface here is
byte-identical to what iteration 2 saw; only the PR description moved.

Iteration 2's finding is **resolved**. `git diff --numstat crewrig/main...a214005`
reports `131 1 hooks/mempalace-transcript.sh` and
`666 0 scripts/tests/test-mempalace-transcript-hook.sh` — the PR body now states
`131` insertions / `1` deletion and `666` insertions / `0` deletions, matching
exactly. I re-derived every other number in the body rather than inheriting it.

### CI status

All **12** required checks **pass** on `a214005`; none failing, none pending.

`build`, `check-agents-size`, `check-ci-parity`, `check-components`,
`check-extension-version-bump`, `check-feedback-routing`, `check-skill-versions`,
`gitlab-ci-check`, `grep-anti-patterns`, `lint-markdown`, `lint-specs`,
`test-harness-curate`.

### Correctness

I re-derived the lock topology against the **installed** library
(`mempalace 3.6.0`, `~/.local/pipx/venvs/mempalace/.../mempalace/`) rather than
trusting the diagnosis in the PR body or the anchor issue. Every claim holds at
the cited line:

- `ChromaCollection._write_lock` — `backends/chroma.py:1339`; late import
  `from ..palace import mine_palace_lock` at `:1348`; `with mine_palace_lock(...)`
  at `:1350`. The lateness is what makes the rebind reach the write path.
- The four write methods that take the lock — `chroma.py:1426` (`add`), `:1439`
  (`upsert`), `:1459` (`update`), `:1604` (`delete`) — all route through that one
  late import. `col.get` is unlocked, so the idempotency probe never contends.
- `palace_path=None` is a documented no-op (`chroma.py:1343-1345`), which is why
  the guard's probe must pass a non-None synthetic path.
- Module-load binders exist exactly where R5's rationale says: `sync.py:24`,
  `miner.py:36`, `convo_miner.py:33`. This is what makes the second patch target
  a real precaution rather than decoration.
- `mine_global_lock = mine_palace_lock` at `palace.py:1213` is a backward-compat
  alias, one definition and zero call sites, off the write path. Leaving it
  unpatched is correct — patching it would widen the relief past R5's "every
  location from which *the write path* resolves the lock".

**Iteration 1's corrected WAL rationale is accurate.** `_wal_log` is called from
`tool_add_drawer` at `mcp_server.py:2562`, directly in the function body — not
inside any `_write_lock()`, which only wraps the `ChromaCollection` write methods
listed above. So the palace lock never guarded that append, and relieving the
lock leaves its exposure exactly as it already was, as the comment now says.

**One design detail worth recording, because it is not obvious and it is load-bearing.**
Against the shipped library, `hasattr(mempalace.backends.chroma, "mine_palace_lock")`
is **False** — the late import at `:1348` is function-local, so no module attribute
exists. Only `mempalace.palace` is actually rebound today (`_relieved_modules ==
['mempalace.palace']`). The identity stage survives this because it reads
`getattr(mod, "mine_palace_lock", _relieved_palace_lock)` — the *default* makes an
absent attribute pass rather than fail. That is precisely the "never create an
attribute the library does not read" design the diff documents, and it is coherent:
I replayed the hook's exact three-stage guard against the real library and it
reaches `VERDICT: relief in force`, with `_lock_relief_hits` growing via
`ChromaCollection._write_lock()` and **no** filesystem access at all on the
effective path.

**End-to-end proof against the real library**, beyond the mocked suite — a peer
process holding the genuine lock on a throwaway palace, the real write path
replayed against it:

- control, no relief → `REFUSED: MineAlreadyRunning` (the defect, reproduced)
- with the hook's relief → `ACQUIRED under relief, hits=1` (R1)
- after the peer exits → `ACQUIRED - protection intact` (R6)

The architectural crux holds too. The lock exists to serialize writers against
the HNSW corruption of #974/#965; the hook's writes all leave the process over
HTTP because `chromadb.PersistentClient` is replaced by `HttpClient`, so this
subprocess touches no local HNSW graph, and a per-machine client-side file lock
could not protect the daemon's internals in any case. R2's ordering is what keeps
that argument sound: the relief is installed strictly after the heartbeat, so a
run whose HTTP routing is *not* established retains the lock's protection. The
better long-term answer remains the service-side contract the spec lists as out
of scope.

**On the fail-closed guard (R3), which the PR asks a reviewer to agree with
explicitly rather than pass over: I agree.** R3 says the hook "SHALL decline to
persist and report a failure" when it cannot establish the relief is in force.
Fail-open would convert an unprovable relief back into the silent, whole-session
transcript loss this spec exists to end, and it would do so invisibly. Exit `5`
naming the location is the strictly better failure. Reversing it would need a
delta-spec, not a code change — correctly stated.

### Requirement coverage

All 7 requirements of `specs/0110-transcript-hook-lock-bypass.md` are implemented,
and all 5 scenarios have named automated assertions:

| Req | Where | Verified |
|---|---|---|
| R1 persist while peer holds lock | relief block + two-sided control | ✅ mocked + real library |
| R2 only after reachability established | block sits after `sys.exit(4)` | ✅ behavioural (import-marker absence) + source-order lock |
| R3 prove in force, else decline | 3-stage guard, `sys.exit(5)` | ✅ mutant 2 proves stage 3 load-bearing |
| R4 distinguishable status + prefix | `5` / `LOCK_BYPASS_INEFFECTIVE:` | ✅ unique vs `2`/`3`/`4` and their prefixes |
| R5 every resolution site, not one | `(_mp_palace, _mp_chroma)`, `hasattr`-guarded | ✅ mutant 1 kills the single-site variant |
| R6 confined to this subprocess | in-memory rebind only | ✅ two-sided ACQUIRE-then-REFUSED |
| R7 session turn always succeeds | hook `exit 0` unconditionally | ✅ on both the rc=5 and rc=4 paths |

### Test coverage

`bash scripts/tests/test-mempalace-transcript-hook.sh` at head:
**`Summary: 25 passed, 0 failed, 0 skipped`**, ~7s — `13` pre-existing + `12` new
spec-0110 assertions, matching the body's arithmetic.

The tests run the **real shipped heredoc**, extracted with the same
`sed -n "/<<.PYEOF./,/^PYEOF$/p"` technique the pre-existing spec-0073 test uses,
so they cannot drift from a paraphrase of the source. The three resolution modes
(`late` / `modload` / `elsewhere`) are faithful to the shipped shapes, and the
mock's `tool_add_drawer` reaches the lock the way `mcp_server.py:1177` does.

**The palace-untouched claim is independently confirmed:** `~/.mempalace/locks`
held `38` entries before the run and `38` after. The suite needs no MemPalace
install and no daemon.

**Both mutation-evidence claims reproduce verbatim** — I did not take these on
trust:

- Drop the `backends.chroma` target → `24 passed, 1 failed`, killing
  `spec-0110-r5`, and dying loudly:
  `rc=5 stderr=LOCK_BYPASS_INEFFECTIVE: mempalace.backends.chroma.mine_palace_lock was not replaced`.
- Disable the behavioural probe (`if _relief_error is None:` → `if False:`) →
  `23 passed, 2 failed`, killing `spec-0110-r3` and `spec-0110-r7`, with the
  decisive observation that the ineffective fixture **stored the entry**
  (`rc=0`, `stored=transcripts|room-16|[USER] must not be stored|transcript-hook`).
  Stage 3 is indeed the only thing standing between an unprovable relief and a
  silent write.

The two-sided assertions (the R1 control, and R6's ACQUIRE-then-REFUSED) are the
right call — one-sided versions of both would pass vacuously.

### Convention compliance

- **Gitmoji** — PR title `🐛 …`; commits `🐛 …` / `📝 …`. No conventional-commit
  prefixes. ✅
- **Branch naming** — `feat/0713-transcript-hook-lock-bypass` matches
  `<prefix>/<NNNN>-<slug>`. ✅
- **Spec-PR ordering** — spec 0110 is on `main` (merged as `e834378`, PR #714)
  before this implementation PR. ✅
- **PR format** — purpose sentences, *How to read this PR?*, *How to test this
  PR?*, *Detailed description (for agents)* all present. ✅
- **Close keywords** — `closingIssuesReferences` resolves to exactly `#713`, the
  anchor/logbook issue. No stray auto-close. ✅
- **CLI matrix** — not triggered. The co-maintenance list names
  `hooks/*-transcript-hooks.json`; this diff touches `hooks/mempalace-transcript.sh`
  (the single shared script all four manifests invoke) and
  `scripts/tests/test-*.sh`, which matches neither that glob nor
  `scripts/{build,install,setup,import,manage}-*.sh`. No parity asymmetry: the
  relief lives inside the shared heredoc, so all four CLIs get identical
  behaviour. ✅
- **Version bump** — no `artifacts/**` source touched, so the rule does not fire;
  `check-skill-versions` and `check-components` both pass. ✅
- **Spec status** — `specs/0110-*.md` remaining at `status: approved` is
  legitimate. I checked this against the freshly-merged spec 0109
  (`0109-spec-status-invariant-on-main.md`), which forbids only `status: draft`
  on `main` for non-delta specs; `approved` is untouched by it. The
  `approved → implemented` flip as a separate metadata-only PR is also the
  repo's established practice (spec 0109's own flip landed separately as #711).
  `lint-specs` passes. ✅
- **No spec conflict** — `specs/0103-mempalace-write-lock-fallback.md` sounds
  adjacent but governs friction-tag fallback on MCP error `-32001` (the server's
  writer lease), a different mechanism from the per-write palace lock this spec
  turns on. Spec 0110 explicitly places that lease out of scope. No
  contradiction. ✅

### Linter findings

`lint-shell.sh` (shellcheck) exits 1, but **the diff introduces no new finding in
the load-bearing file**, and none of any consequence:

- `hooks/mempalace-transcript.sh` — **head 2, main 2**. The `131` added lines add
  zero findings. Both are pre-existing `SC2034` notices (`TOOL_INPUT`,
  `STOP_REASON` appear unused) at `:113` and `:134`; they shifted down by 10 lines
  purely because of the new exit-code table in the header.
- `scripts/tests/test-mempalace-transcript-hook.sh` — head 36, main 20. All 16
  new ones are `info`/`note` severity `SC2030`/`SC2031` at lines `945-950` and
  `1021-1026`, e.g.:

  ```
  note: Modification of PYTHONPATH is local (to subshell caused by (..) group). [SC2030]
  note: CREWRIG_TEST_PALACE was modified in a subshell. That change might be lost. [SC2031]
  ```

  These are false positives on a deliberate idiom: `export` inside a `( … )` group
  is exactly how these tests scope environment to one hook invocation, and the
  pre-existing tests in the same file already trip the identical notes at lines
  `131`, `132`, `212`, `213`, `363-365`, `379`, `381`. The new code follows the
  established house pattern rather than departing from it, and CI does not gate on
  shellcheck info notes (`grep-anti-patterns` passes).

`lint-markdown.sh`, `lint-python.sh`, `lint-json.sh`, `lint-skill.sh` — not
applicable; the diff changes two `.sh` files and nothing else.

### One observation, deliberately NOT raised as a finding

In the R4 assertion (`scripts/tests/test-mempalace-transcript-hook.sh:909-923`),
two branches inside the `for pair in "IMPORT_ERROR:2" "ADD_FAILED:3"
"DAEMON_UNREACHABLE:4"` loop are structurally unreachable, because both operands
are literals from that same hardcoded list:

- `echo "$prefix" | grep -q "LOCK_BYPASS_INEFFECTIVE"` — `$prefix` is only ever
  one of the three literals above, none of which contains that substring.
- `[ "$code" = "5" ]` — `$code` is only ever `2`, `3`, or `4`.

So the uniqueness of status `5` and of the `LOCK_BYPASS_INEFFECTIVE:` prefix is
established by *construction* of the list rather than derived from the hook, and a
hypothetical future fifth failure path reusing exit `5` would not be caught by
this loop.

**I am not raising this as a finding, and I do not think it warrants a fourth
iteration.** R4 is genuinely satisfied — I verified the four exit codes and four
prefixes are mutually distinct by reading the hook directly — and the same
assertion does perform real work: it greps the hook source for all four
`sys.exit(N)` and requires that stderr carry the new prefix and none of the other
three. The dead branches cause no false pass of any property that matters; the
requirement speaks to "every failure the hook already reports", and the hardcoded
list *is* that set. Editing two unreachable conditionals would change no
behaviour and no coverage. Routing the loop back to DEV for it would be
manufacturing a marginal finding to avoid the appearance of a rubber stamp, which
is a worse failure than saying plainly that the PR is clean.

I record it only so the next reader of that assertion is not misled about how
much it derives, and so my judgment call is auditable rather than silent.

### Verdict

**APPROVE — zero findings of any class**, CI green (12/12) on the reviewed head
`a214005`. Two prior passes each surfaced one non-blocking issue and both are
resolved: iteration 1's WAL rationale is now accurate against the installed
library, and iteration 2's diff-stat now matches `git diff --numstat` exactly.

This pass was not a formality — I re-derived the lock topology line-by-line
against `mempalace 3.6.0`, reproduced both mutants, and proved R1 and R6 against
the real library rather than only the mocks. I found nothing further. The
termination criterion for the retroactive loop is met.

PR #734, opened 2026-08-08
✨ Reserve a spec id atomically at ticket pickup

hcross:
Spec ids are allocated with an unsynchronised `max(existing) + 1` over the local tree, so two sessions picking up two tickets in the same second walk away with the same number and the collision is discovered by an unrelated third pull request *after* both specs have merged. This change secures the id on the reference repository **before** the spec file, the branch name, or the frontmatter exist, and adds a pull-request backstop that fails when an id was never secured or was secured for a different ticket.

Implements `specs/0112-spec-id-reservation.md` and `specs/0112-spec-id-reservation.delta-01.md`. Governing plan: **PLAN v11** on the logbook. Logbook: #726.

## How to read this PR?

24 files, +2994/−76, 10 commits. Only three files carry design; the rest are consumers, docs, and generated output.

> **Read this before conformance-checking against the plan.** Five behaviours in this branch **post-date the plan the user validated (v9)**: the `CREWRIG_SPEC_ID_ORIGIN` override, its refusal inside a CI context, the platform-pair-beats-override precedence, the refusal of a carrier file that exists but names no namespace, and the `--id`-on-occupied-ref semantics (same ticket exits `0` idempotently, a different ticket fails naming both). The last two are documented in the exit-code tables below rather than under *Post-plan surface*, so they are easy to conform against v9 by mistake. They arrived during DEV and were **ratified retroactively in PLAN v11** — they were not in the approved plan when DEV started. A reviewer conforming the diff against v9 would have no entry for them and would read them as deviation. They are itemised under *Post-plan surface* below; judge them, do not skip them.

**1. `.crewrig/spec-id-carrier`** (new, 36 lines, almost all comment). Read this first — it states the mechanism, the closed pair of admissible values, why the setting is a tracked repository-scoped file rather than an environment variable, and why the org namespace is a sibling. Everything else implements what this file says.

**2. the `--force-with-lease` push in `scripts/reserve-spec-id.sh`** — the whole guarantee, in six lines:

```sh
push_out="$(git -C "$REPO_DIR" push --force-with-lease="${REF}:" \
                "$REMOTE" "${OBJ}:${REF}" 2>&1)" || push_rc=$?
```

Two properties are load-bearing, for **different** reasons, and neither substitutes for the other:

- `OBJ` is **parentless** (`reservation_object()` at `:475` — `git commit-tree` with no `-p`). That guarantees *our own* object can never descend from an incumbent. It is a property of our behaviour, not a constraint on anyone else's.
- `--force-with-lease="${REF}:"` — an **empty** expected value, meaning "this ref must still not exist". That turns the property above into a precondition the **server** enforces against any pusher, rather than one a reader has to infer from the object's shape.

Why the pair is not redundant, reproduced on a throwaway bare repository:

```
# empty lease at an occupied ref → refused
 ! [rejected]  f16194d -> refs/spec-ids/0001 (stale info)

# leaseless push of a DESCENDANT at the same occupied ref → ACCEPTED
   f69c653..16fd684  16fd684 -> refs/spec-ids/0001
```

The second line is the hazard. A leaseless push of a descendant is an ordinary fast-forward and the server's create-only lock does not stop it. Case 10 in `scripts/tests/test-reserve-spec-id.sh` (Case 10) asserts that acceptance **deliberately**, so the case fails loudly if git's behaviour ever changes and the premise evaporates. Note the refusal text is `(stale info)` — the `cannot lock ref … already exists` quoted by earlier plan revisions is the *leaseless* message and was never what ships.

**3. `determine_origin()` in `scripts/check-spec-id-reserved.sh`** (`determine_origin`) — the only function with non-obvious precedence, and the densest concentration of post-plan surface. Read the *ordering*, not just the branches.

Then, in any order: `docs/spec-pr-workflow.md` → *Reserving the spec id*, `artifacts/core/skills/spec-author/SKILL.md` → *ID allocation*, `ci/ci-capabilities.yml` + `.github/workflows/build.yml`, then the two suites.

### Post-plan surface — not in v9, recorded in v11

PLAN step 10 required a loud failure when pull-request origin is undeterminable, on the grounds that this is a wiring defect. During DEV that rule turned out to make `check-spec-id-reserved.sh` **unrunnable by a maintainer outside CI**, where there is no wiring to be defective and no way to satisfy the demand. Three things followed, all in this branch and all now in v11:

1. **An explicit override, `CREWRIG_SPEC_ID_ORIGIN=same|fork`.** Explicit in every context, with the variable named in the failure message. The alternative considered and **withdrawn** was inferring leniency from the *absence* of a CI marker — inferring permission-to-be-lenient from an absent signal is the shape of reasoning this plan spent three revisions removing.
2. **The override is refused outright inside a CI context** (`CI` / `GITHUB_ACTIONS` / `GITLAB_CI`), pair or no pair (`:174-194`). Not merely demoted below a *present* pair: a CI job whose platform pair **failed to map** is exactly where a stray `fork` does its damage, and demoting-only would have left that case open. Before `17344e0` the override was read first, so `CREWRIG_SPEC_ID_ORIGIN=fork` silently downgraded the blocking branch to report-only on every pull request. A guard an environment variable can switch off is not a guard.
3. **A complete platform pair outranks the override anywhere**, and is deliberately **not** gated on the CI marker — a complete pair is the platform's own statement of provenance and is authoritative in any context. A malformed override value fails wherever it appears rather than degrading to "unset".

Three cases pin exactly these, each mutation-verified to bite: Case 11b (inside CI the override does not rescue an unresolvable pair), Case 11c (outside CI it resolves origin and the run proceeds), Case 11d (a present platform pair outranks it).

### Other non-obvious decisions a reviewer will otherwise misread

**The org namespace is a sibling, not a child — a correctness property, not a style choice.** Org reservations live at `refs/spec-ids-org/<ID>`, never `refs/spec-ids/org/<ID>` (the namespace constants in `scripts/reserve-spec-id.sh`). An `ls-remote` refspec pattern's `*` **crosses `/`**, so a read of `refs/spec-ids/*` returns `refs/spec-ids/org/<ID>` too — measured, and recorded in the v9 preamble carried into v11. Nesting would have placed org reservations inside the upstream allocated union, breaching the delta's replaced R3 and R15. Filtering `org/` out of the read would satisfy an assertion while leaving the violation in place; the separation is structural instead. Pinned by Cases 13-14.

**`05a428e` is a correctness fix wearing a refactor's clothes.** Restructuring remote resolution onto the repository's documented `BASE_REF` idiom (the reference-ref resolution in `scripts/reserve-spec-id.sh`, aligned with `resolveBaseRef()` in `scripts/lib/spec-linter.js` and the `BASE_REF` idiom in `scripts/check-skill-versions.sh`) exposed that `merged_spec_paths()` read the merged-id set from a hardcoded `"${REMOTE}/main"`. A `BASE_REF` pointing at `release/**` — a branch pattern both CI engines already trigger on — would have been honoured for the remote *name* and **silently ignored** for the *branch*, computing the allocated set against `main` while the change targeted `release/**`. The diff reads as one variable being renamed; it is a fix, and v11's step 1 was amended to stop specifying the defect.

**Two maintenance hazards of the same class — a test that stays green against the defect it exists to catch.**

- **Do not rewrite the namespace assertions with `for-each-ref`.** `git for-each-ref` does **not** cross `/`; `git ls-remote` does. The rewrite would leave the assertions **green against the nested-namespace defect the sibling namespace exists to prevent**. Both suites carry the warning at the top (the header note in `scripts/tests/test-reserve-spec-id.sh`, the header note in `scripts/tests/test-check-spec-id-reserved.sh`).
- **Do not let the check suite inherit ambient environment.** Its subject *reads* `CREWRIG_SPEC_ID_ORIGIN` and the three CI markers, so a hermetic suite that inherits them is not hermetic — a maintainer who exported the override in order to run the check by hand, the very workflow it exists for, would turn Case 11 green on their laptop and red on a runner. `run_check()` clears exactly those variables and each case sets the ones it is about (`run_check()` in `scripts/tests/test-check-spec-id-reserved.sh`).

### Deliberate non-changes

- **`AGENTS.md` is untouched.** 21879 B against the `-ge 22000` threshold in the `THRESHOLD` guard in `scripts/check-agents-size.sh` — 121 B of headroom. The workflow prose lives in `docs/spec-pr-workflow.md`, which `AGENTS.md` already points at.
- **`specs/org/README.md` is untouched.** Org-owned (`excluded` in `.crewrig/core-paths.txt`). Upstream gains the capability to secure an org identifier and documents it in its own tree; it does not write into the overlay.
- **`scripts/build-ci.sh` and `scripts/check-ci-parity.sh` are untouched.** The new capability declares **no `tools:` key** and installs git conditionally as `command[0]`, following the `lint-markdown` precedent. `requires: tools: [git]` would fail `check-ci-parity.sh` Arm 1 unless three coordinated edits landed in the generator and the parity harness; this shape needs none. The inline install carries **no nested `sudo`** — `normalize_cmd` strips only a *leading* one.
- **No `--remote` flag.** The remote is the reference ref's first path component, so the documented `BASE_REF` idiom already carries it. A third resolution idiom in a repository that explicitly has one would be a review finding.

## How to test this PR?

Both suites are hermetic — a local bare repository over `file://`, no network, no forge, no credential. **122 assertions** across the two (78 + 44). Prerequisite: `git` and `bash`. Every run below is green on `f3060c7`.

```sh
bash scripts/tests/test-reserve-spec-id.sh         # → Results: 78 passed, 0 failed
bash scripts/tests/test-check-spec-id-reserved.sh  # → Results: 44 passed, 0 failed
```

The `check-components` job runs both with `CI` and `GITHUB_ACTIONS` set, and the check's subject reads those markers, so verify the ambient-CI arm — and the deliberately **hostile** one, which is the real assertion that the suites do not inherit what their subject reads:

```sh
CI=1 GITHUB_ACTIONS=true bash scripts/tests/test-reserve-spec-id.sh         # → 78 passed, 0 failed
CI=1 GITHUB_ACTIONS=true bash scripts/tests/test-check-spec-id-reserved.sh  # → 44 passed, 0 failed

CREWRIG_SPEC_ID_ORIGIN=fork CI=true GITHUB_ACTIONS=true \
  bash scripts/tests/test-reserve-spec-id.sh                                # → 78 passed, 0 failed
CREWRIG_SPEC_ID_ORIGIN=fork CI=true GITHUB_ACTIONS=true \
  bash scripts/tests/test-check-spec-id-reserved.sh                         # → 44 passed, 0 failed
```

Per the DEV record, **20 mutations of a throwaway reference implementation were verified to bite** — each caught by the case written for it and no others. "The suite passes" and "the suite would catch the regression it exists for" are different claims; only the second is worth the maintenance cost.

**Golden path against this repository.** Secures a real ref on `crewrig/crewrig`; pick a number well past the corpus so you are not taking an id someone is about to use, and delete it afterwards:

```sh
bash scripts/reserve-spec-id.sh --id 0999 --issue 726   # → prints 0999, exit 0
git ls-remote crewrig 'refs/spec-ids/*' | grep 0999     # → the ref is there
git push crewrig --delete refs/spec-ids/0999            # → clean up
```

**Failure mode 1 — the id is held by another ticket.** A re-run with a *different* issue number is refused; a re-run with the *same* one is idempotent, so a maintainer can safely repeat the pre-merge command:

```sh
bash scripts/reserve-spec-id.sh --id 0999 --issue 726 ; echo "rc=$?"   # rc=0
bash scripts/reserve-spec-id.sh --id 0999 --issue 726 ; echo "rc=$?"   # rc=0, "nothing to do"
bash scripts/reserve-spec-id.sh --id 0999 --issue 727 ; echo "rc=$?"   # rc=1, names both tickets
git push crewrig --delete refs/spec-ids/0999
```

**Failure mode 2 — no remote reachable.** Exit `3`, an id on stdout, and the frontmatter line to paste:

```sh
bash scripts/reserve-spec-id.sh --issue 726 --offline ; echo "rc=$?"
# → <id>
#   unsecured-id: true
#   rc=3
```

**Failure mode 3 — a non-representable org identifier.** Refused before any push:

```sh
bash scripts/reserve-spec-id.sh --corpus org --id 'RFC 2026' --issue 726 ; echo "rc=$?"   # rc=1
```

**The backstop, locally** — the post-plan surface, exercised by hand. It refuses to guess origin; outside CI the override is the admissible signal and the error message names it:

```sh
bash scripts/check-spec-id-reserved.sh ; echo "rc=$?"
# → [ERROR] cannot determine whether this pull request comes from the reference
#           repository or from a fork … rc=2

CREWRIG_SPEC_ID_ORIGIN=same bash scripts/check-spec-id-reserved.sh ; echo "rc=$?"
# → Base ref: crewrig/main | remote: crewrig | origin: same (blocking: true)
#   Specs added or renamed by this change: 0 (delta-specs exempt: 0; org-owned exempt: 0)
#   OK: this change adds or renames no non-delta upstream spec, so no id needs securing.
#   rc=0

CI=1 CREWRIG_SPEC_ID_ORIGIN=same bash scripts/check-spec-id-reserved.sh ; echo "rc=$?"
# → the override is IGNORED inside a CI context and the run fails: rc=2
```

**Generated-CI agreement and the repository's own guards** — `.gitlab-ci.yml` in this diff is generated, not hand-written:

```sh
bash scripts/build-ci.sh --check      # → OK
bash scripts/check-ci-parity.sh       # → OK
bash scripts/check-core-paths.sh      # → OK
bash scripts/check-test-wiring.sh     # → OK
task spec:lint                        # → OK
```

### Added

**`scripts/reserve-spec-id.sh`** (609 lines) — the allocation tool. Bash 3.2 per spec 0111, invocable by a human with no agent involved (R12).

- *Allocated set (upstream corpus only)*: the union of the non-delta spec ids merged on the reference branch **outside `specs/org/`**, `refs/spec-ids/*`, and `refs/tags/spec-id/*`. Both upstream carriers are read regardless of which is configured, so a repository mid-transition keeps a complete set. Org namespaces are never read into it (delta R3, R15).
- *Reservation object*: `git commit-tree` with **no `-p`** against the empty tree, message `reserve <ID> for issue #<N>` — that message is the record the backstop parses.
- *Write*: one push to the configured carrier only, with `--force-with-lease="${REF}:"`. On refusal the tool re-reads the id — present → lost race, advance and retry under a bounded budget; absent → **exit `3`**. It **never** switches carrier: under `transfer.hideRefs` the re-read reports absent, and a fallback would let one id be held twice.
- *Exit contract* (authoritative): `0` secured, id on stdout — also returned idempotently when `--id` names a ref already secured **for the same issue**. `3` allocated locally but not secured (offline, unreachable remote, no write access — the fork case — or a carrier the remote hides): id on stdout plus `unsecured-id: true` on line 2, in the **exact frontmatter syntax** so `spec-author` pastes without reformatting. Not `key=value`: the exit code already carries the boolean, so no shell caller needs the line, and its only consumer writes YAML. `1` genuine failure — malformed carrier setting, non-representable identifier, `--corpus org` without `--id`, an id held by a *different* ticket, or a retry budget exhausted by lost races — reason on stderr, nothing on stdout.
- *Flags*: `--issue <N>` (default mode), `--id <ID>`, `--corpus upstream|org` (default `upstream`), `--offline`. `--corpus org` **requires** `--id`: computing the next free org identifier is out of scope by delta R16, so the tool exits `1` saying so rather than silently falling back to an upstream computation. The corpus is **told, never inferred** from the identifier's shape — inferring would require upstream to recognise an org convention, which `specs/0071-org-specs-lint-exclusion.md` deliberately prevented.
- *Representability gate*: `git check-ref-format` before any push. Delta R14 calls the identifier opaque, and it stays opaque to the tool's *logic* — nothing parses or interprets it — but the carrier is a git ref and cannot hold a space, `~`, `..`, or a trailing `.lock`. Surfacing that as a named refusal costs one call; discovering it through a raw git diagnostic costs an adopter an afternoon.
- *Remote resolution*: `BASE_REF` when set, else the first remote matching `crewrig|origin`, falling back to the first remote at all, with `/main` appended (`:344`). **No repository path and no URL is named anywhere in either script**, which is what lets the suites aim the tool at a `file://` bare repository.

**`scripts/check-spec-id-reserved.sh`** (528 lines) — the pull-request backstop (R7 as replaced by the delta).

- For each non-delta spec **added or renamed** by the change **outside `specs/org/`**: resolve the id, read **both** upstream namespaces, compare the recorded issue against `related-issue`.
- *Origin discrimination* (R10, and the post-plan surface above): platform pair first — `CREWRIG_PR_HEAD_REPO`/`CREWRIG_PR_BASE_REPO`, then `CI_MERGE_REQUEST_SOURCE_PROJECT_PATH`/`CI_PROJECT_PATH` — then, **outside a CI context only**, `CREWRIG_SPEC_ID_ORIGIN`. Nothing infers origin from a remote URL, a branch name, or a CI variable of the wrong engine.
- *Two-records rule*, scoped to the two **upstream** carriers: one upstream id recorded in both `refs/spec-ids/*` and `refs/tags/spec-id/*` naming two different issues **fails, naming both tickets**. It does not pick a winner — that state is the exact duplicate-holder outcome this ticket exists to prevent. An upstream `0042` and an org `0042` are **not** a collision (delta scenario 5; Case 5).
- *Two deliberately asymmetric degradations*: reserved set **unreadable** → report without failing, because a credential gap must never block an adopter's pipeline and an unreadable namespace is measurably indistinguishable from an empty one. Origin **undeterminable** → fail loudly.
- *Exit codes*: `0` every checked id secured for its own ticket, or a fork's conditions reported, or an unreadable reserved set reported. `1` at least one blocking finding on a same-repository pull request. `2` wiring fault — origin undeterminable, base ref unresolvable, or not a git repository. Fail closed: a run that could not look must not read as a run that found nothing.
- *Delta-specs exempt* (R11). *`specs/org/` exempt unconditionally* — see `c618b3e` below.

**`.crewrig/spec-id-carrier`** — one tracked line, `carrier=refs/spec-ids/`, wrapped in its rationale. Closed pair: `refs/spec-ids/` or `refs/tags/spec-id/`. Any other value exits `1` naming it, which closes the typo path where `refs/spec-id/` would write into a third namespace neither the union nor the check reads. An **absent** file means the built-in default (the entry is `excluded`, so the synchroniser will never restore one a fork deleted); a **present but unparsable** file exits `1` naming the file. Intent is the discriminator: absent expresses none, so the default is right; a present file is a statement of intent, and silently ignoring a malformed value would leave an adopter whose remote refuses the default writing to it forever, seeing exit `3` on every allocation with nothing pointing at the cause. `CREWRIG_SPEC_ID_CARRIER` survives only as a documented one-off override for a single invocation.

**`scripts/tests/test-reserve-spec-id.sh`** (678 lines, 59 assertions) and **`scripts/tests/test-check-spec-id-reserved.sh`** (530 lines, 27 assertions) — hermetic, local bare repository, no network, no forge, and neutralised against the ambient variables their subject reads. Wired into both engines via `ci/ci-capabilities.yml` → `check-components`; `scripts/check-test-wiring.sh` passes.

### Modified

- **`.crewrig/core-paths.txt`** — adds `.crewrig/spec-id-carrier	excluded`, carved out of the `strict` `.crewrig` parent exactly as `.crewrig/.synced-markers` already is. `adopt-on-edit` does **not** work here: the only nesting mechanism the manifest defines and `sync-from-upstream.sh` implements is an `excluded` carve-out, nothing performs most-specific resolution, so a nested `adopt-on-edit` member is governed by its `strict` parent and the adopter's own edit would **halt** their sync. `check-core-paths.sh` only verifies that listed paths resolve, so that defect would have reached a real adopter with CI green. The trade-off is symmetric and permanent: upstream can never update the value afterwards either. Distribution is unaffected — `excluded` governs synchronisation, not the tree a fork clones.
- **`docs/layers.md`** — classifies the new setting in the Overlay layer, in the same diff as the manifest entry, per the co-maintenance rule in `AGENTS.md` → *CLI Matrix Maintenance*.
- **`docs/spec-format.md`** — adds the optional `unsecured-id` boolean to the frontmatter schema. Absent by default, **never back-filled**, meaningful only when present and `true`. Discharges spec 0112's single open question, which assigned the amendment to the implementation pull request.
- **the `unsecured-id` validation in `scripts/lib/spec-linter.js`** — validates `unsecured-id` when present. A null value is tolerated on the same terms as `max-iterations`; `false` is accepted rather than rejected as redundant, because the check validates the field and does not police whether writing it was necessary. The mistake worth catching is the quoted string `"true"`, which reads as set to a human and is not a boolean to any consumer. The spec-0098 cross-file duplicate-id guard is untouched and stays as the backstop it was (R8), and `task spec:lint` acquires no network dependency.
- **`artifacts/core/skills/spec-author/SKILL.md`** (`1.3.2` → **`1.4.0`**, MINOR) — *ID allocation* replaced. Run the tool first, before the filename, branch name, or frontmatter exist (R2). Three obligations close tempting shortcuts: on exit `1` the skill SHALL NOT fall back to `max(existing) + 1` and SHALL NOT re-run hoping for a different answer; on exit `3` the mark is **not optional**; the skill never passes `--corpus`, never passes `--id`, and never sets `CREWRIG_SPEC_ID_CARRIER`. Delta-spec mode does not call the tool at all.
- **`artifacts/core/agents/spec-author/AGENT.md`** — mirrored, version bumped in the same diff (`scripts/check-skill-versions.sh`).
- **`docs/spec-pr-workflow.md`** — new *Reserving the spec id* section with three sub-sections: the offline and fork paths, how an adopter changes the carrier namespace, and how an organization secures an identifier for a spec under `specs/org/`. That last names the **representability constraint itself**, not merely the command — an organization needs to know that a space, `..`, `~`, or a trailing `.lock` cannot be secured *before* it designs a convention, not at first use.
- **`docs/cli-matrix.md`** — row 28. Parity here is **structural, not asserted**: both scripts are plain `git`-only Bash invoked from the shell, with no per-CLI surface, no MCP tool, and no forge API (R4, and `AGENTS.md` → *Forge Access*). The contract each CLI's `spec-author` consumes is the exit code alone, and enforcement is CI-side rather than CLI-side.
- **`ci/ci-capabilities.yml`** — the `check-spec-id-reserved` capability: `portable`, `requires: {history-depth: full}`, `branches: [main, "release/**"]`, no `tools:` key. Both new test scripts appended to `check-components`.
- **`.github/workflows/build.yml`** — the job, gated `if: github.event_name == 'pull_request'` (origin discrimination has no meaning on a push), `fetch-depth: 0`, and an `env:` block mapping `github.event.pull_request.head.repo.full_name` → `CREWRIG_PR_HEAD_REPO` and `github.repository` → `CREWRIG_PR_BASE_REPO`. Those are workflow **contexts**, not ambient environment; they reach the process only through this mapping. Named `CREWRIG_*` rather than `GITHUB_*` because GitHub reserves that prefix and there is no native head-repo variable to mirror anyway — `GITHUB_HEAD_REF` is the branch, not the repository, which is precisely why the mapping must exist. On GitLab CI the equivalent pair is already ambient in a merge-request pipeline, so the generated job needs no counterpart.
- **`.gitlab-ci.yml`** and the four CLI bundles (`.claude/`, `.gemini/`, `.github/`, `.agents/`) — **generated**. Reproduce with `bash scripts/build-components.sh` and `bash scripts/build-ci.sh`; do not hand-edit.

### Two commits worth reading on their own

- **`17344e0`** — the origin-precedence hardening described under *Post-plan surface*, plus the unparsable-carrier refusal. Neither widens the tool's surface; both remove a way for it to be silently wrong.
- **`c618b3e`** — the `specs/org/` exemption had been derived *solely* from the `excluded` entries in `.crewrig/core-paths.txt`. That is the right *extension* mechanism (a future org-owned path is covered without editing the script), but it made the exemption contingent on a file the script happens to be able to read: with the manifest absent or unreadable the derived list is empty and the guard becomes upstream enforcement over org-owned content — the precise layer breach `specs/0071-org-specs-lint-exclusion.md` exists to prevent, degrading in the direction that *blocks an adopter*. `specs/org/` is now the unconditional floor, which is what the delta's replaced R7 states literally, with the manifest read layered on top. Floor, not ceiling — the same posture as `ci/bash32-forbidden.txt`. Caught by Case 6b.

### Known limits, stated rather than hidden

- **GitLab remains unmeasured.** No instance was reachable during PLAN or DEV. An adopter whose remote refuses `refs/spec-ids/` changes one tracked setting, so a refusal is a documented configuration step rather than a design failure.
- **Retry-budget exhaustion under contention is untestable hermetically.** A `pre-receive` hook cannot create refs (`ref updates forbidden inside quarantine environment`, measured), and every other mechanism collapses to a different branch. Exit `1` is covered by four other paths; what is untested is the loop's *termination*, not the exit code.
- **The representability boundary has a tested refusal and no tested workaround, because there is none.** An organization on an unrepresentable convention cannot use the mechanism until it changes the convention, and upstream cannot do that for it. Flagged because *accepted risk* and *untested* look identical in a coverage report and are not the same thing.
- **`git push --force` defeats the guard.** Custom namespaces cannot be branch-protected. Accepted: a deliberate act, and the backstop still catches the mismatch before merge.

Refs #726.


hcross:
## Verdict: REQUEST CHANGES

Cold review — no authoring context. Audited against `specs/0112-spec-id-reservation.md`, its `delta-01`, `AGENTS.md`, and `docs/`, at head `e5053f5`. This is careful, unusually well-reasoned work: the sibling-namespace decision, the parentless-object premise, and the origin-precedence fix are all correct, and I verified them by measurement rather than by reading the body. Two findings, both in the *check* script's periphery rather than in the mechanism itself.

### CI status

All 13 required checks **pass** on `e5053f5` (run `31255003608`): `build`, `check-agents-size`, `check-ci-parity`, `check-components`, `check-extension-version-bump`, `check-feedback-routing`, `check-skill-versions`, `check-spec-id-reserved`, `gitlab-ci-check`, `grep-anti-patterns`, `lint-markdown`, `lint-specs`, `test-harness-curate`. No failing or pending job. Preflight clear.

Note that `check-spec-id-reserved` passing here is not evidence of the guard working: this PR adds no spec file, so the run took the `candidate_count -eq 0` early exit at `scripts/check-spec-id-reserved.sh:343-346`. I exercised the guard against constructed repositories instead.

### Correctness

**The PR body's load-bearing claims are accurate.** I re-measured each one rather than accepting it:

- `git ls-remote 'refs/spec-ids/*'` **does** cross `/` — it returned `refs/spec-ids/org/0042` alongside `refs/spec-ids/0042`. `git for-each-ref 'refs/spec-ids/*'` returned only the flat ref. The sibling-namespace decision at `scripts/reserve-spec-id.sh:289-290` and the maintenance warnings at `scripts/tests/test-reserve-spec-id.sh:70-81` are therefore both justified as written.
- A plain leaseless push of a **descendant** to an occupied reservation ref **succeeds** as an ordinary fast-forward (measured, git 2.55.0). The Case 10 premise is real, not defensive decoration.
- `--force-with-lease="<ref>:"` refuses when the ref exists (`! [rejected] … (stale info)`) and succeeds when absent. The create-only compare-and-swap at `scripts/reserve-spec-id.sh:566-567` does what the comment says. It also refuses the descendant, so the lease alone carries the guarantee and parentlessness is defence-in-depth — which is what the body claims, hedged correctly.
- `05a428e` is indeed a fix, not a rename: `merged_spec_paths()` now reads `"$REFERENCE_REF"` at `scripts/reserve-spec-id.sh:388`, not a hardcoded `"${REMOTE}/main"`.
- Origin precedence at `scripts/check-spec-id-reserved.sh:141-197` is ordered as described — platform pairs first, override refused inside CI, override admissible only outside it.
- `AGENTS.md` is 21879 B against the `-ge 22000` threshold at `scripts/check-agents-size.sh:5-10` — 121 B of headroom, as stated.
- `normalize_cmd` at `scripts/check-ci-parity.sh:86-88` strips only a leading `sudo` (`sed -e 's/^sudo //'`), so the no-nested-`sudo` reasoning holds.
- `specs/org/README.md` is absent from the diff; `.crewrig/core-paths.txt:124` carries the `excluded` entry; a nested `adopt-on-edit` child would not be carved out, since `excluded_children_of` at `scripts/sync-from-upstream.sh:140-143` selects `excluded` only. The manifest reasoning is sound.

**Finding 1 — `class: tech` — the GitLab arm resolves the wrong base ref on a `release/**` merge request, and blocks on another ticket's spec.**

`.github/workflows/build.yml:394` maps `BASE_REF: origin/${{ github.event.pull_request.base.ref }}`. The generated GitLab job at `.gitlab-ci.yml:168-176` sets no `BASE_REF` and passes no positional argument, so `scripts/check-spec-id-reserved.sh:225` falls through to `origin/main` — while the job's own rule at `.gitlab-ci.yml:176` fires on merge requests targeting `release/**` too.

Reproduced on a constructed repository: a release branch carrying spec `0600` (issue #900, added by someone else), and an MR into `release/2.0` adding only `0601`, properly secured for #726.

- GitHub arm, `BASE_REF=origin/release/2.0` → `Specs added or renamed by this change: 1`, exit `0`.
- GitLab arm, exactly what `.gitlab-ci.yml` runs → `Base ref: origin/main`, `Specs added or renamed by this change: 2`, and a **blocking** finding: `specs/0600-s0600.md — id '0600' was never secured … --issue 900`. Exit `1`.

A legitimate merge request is failed, and the message attributes a third party's spec to it — the misattribution failure mode spec 0112's Intent paragraph exists to remove, arriving through the guard itself. This is also the defect class the body claims credit for fixing in `05a428e`, using `release/**` as the motivating branch pattern precisely because "both CI engines already trigger on" it; the sibling engine ships with it unaddressed. `AGENTS.md` → *Multi-CLI parity* prohibits silent asymmetry.

The script already special-cases GitLab for origin discrimination at `:154-165`, so reading `CI_MERGE_REQUEST_TARGET_BRANCH_NAME` in the `BASE_REF` fallback is consistent with the shape already there and needs no generator change.

**Finding 2 — `class: tech` — nothing reconciles the `unsecured-id` mark with the reservation state, so a merged spec can carry a stale `true`.**

`docs/spec-format.md:34`, added by this PR, states the invariant flatly: "The mark SHALL be removed in the same act that secures the id (requirement 10) — **a merged spec never carries it**." Nothing enforces it.

- `scripts/lib/spec-linter.js:357-361` only type-checks the boolean.
- `scripts/check-spec-id-reserved.sh:486-492` compares holder against ticket and, on a match, logs `OK` without ever reading the frontmatter mark.

Reproduced: a spec carrying `unsecured-id: true` whose id **is** secured for its own ticket yields `OK … id '0500' secured for issue #726`, exit `0`, with the stale mark still in the file. It merges green.

I want to be fair about the class of this one: it is not an R7 violation — R7 governs *never secured* and *secured for a different ticket*, and R10 assigns mark removal to the maintainer. It is an unenforced invariant that this PR's own documentation asserts, on a field whose entire purpose (R9) is to be machine-readable. A false `true` makes every consumer of the field wrong in the permissive-looking direction, and the check already holds both facts at `:486`, so closing it is a few lines. Given how fastidious the rest of this change is about silent degradation — the malformed-carrier refusal at `:268-278` exists for exactly this reason — the asymmetry stands out.

### Convention compliance

Clean. No findings.

- **Gitmoji** — title `✨ Reserve a spec id atomically at ticket pickup`; all 10 commits carry a gitmoji, no conventional-commit prefixes.
- **Version bumps** — `artifacts/core/skills/spec-author/SKILL.md` `1.3.2` → `1.4.0` and `artifacts/core/agents/spec-author/AGENT.md` `1.0.1` → `1.1.0`, both MINOR for an additive contract change, both in this diff. `scripts/check-skill-versions.sh` passes.
- **Built components** — `bash scripts/build-components.sh` produces no drift; the four CLI bundles agree with `artifacts/core/`.
- **CLI matrix** — row 28 at `docs/cli-matrix.md:64` names both scripts, the carrier file, and the capability, with parity argued structurally. Satisfies R13 and `AGENTS.md` → *CLI Matrix Maintenance*.
- **Core-paths co-maintenance** — `.crewrig/core-paths.txt:124` and `docs/layers.md` updated in the same diff, as `AGENTS.md` requires.
- **PR body** — all four mandated sections present.
- **Repository guards** — `check-core-paths`, `check-test-wiring`, `check-ci-parity`, `check-skill-versions`, `check-agents-size`, `build-ci.sh --check` all exit `0` locally.
- **Docs** — `docs/spec-pr-workflow.md` → *Reserving the spec id* (`:30`) carries the three claimed sub-sections at `:58`, `:91`, `:117`.

### Test coverage

Strong, and honest about its own edges. Both suites reproduce the body's figures exactly:

- `scripts/tests/test-reserve-spec-id.sh` → `59 passed, 0 failed`
- `scripts/tests/test-check-spec-id-reserved.sh` → `27 passed, 0 failed`
- Both again under `CI=1 GITHUB_ACTIONS=true` → `59 / 0` and `27 / 0`.

Hermetic as claimed — local bare repository over `file://`, no network. Case 10 (`:489-514`) pinning git's *acceptance* of the descendant push is the right instinct: it fails loudly if the premise ever evaporates. The `Known limits` section correctly separates accepted risk from untested behaviour, which most PRs conflate.

Two gaps, corresponding to the findings above:

- No case drives `check-spec-id-reserved.sh` with a base ref other than `<remote>/main`, which is why Finding 1 is invisible to the suite. A case asserting the resolved base ref on a `release/**` target would have caught it.
- No case pairs a present `unsecured-id: true` with a secured id (Finding 2). Cases 1-12 cover the reservation states but never read the frontmatter mark.

### Linter findings

`lint-shell.sh` on `scripts/reserve-spec-id.sh`, `scripts/check-spec-id-reserved.sh` and both test scripts — exit `0`, no output. All four carry `set -euo pipefail`. I also confirmed every `git` refspec in both scripts uses the braced `"${VAR}:..."` form (`reserve-spec-id.sh:490,566-567`, `check-spec-id-reserved.sh:371`), which is the correct defence against a shell eating the `:`-suffixed path.

`lint-skill.sh` on `artifacts/core/skills/spec-author/SKILL.md`:

```
artifacts/core/skills/spec-author/SKILL.md: missing required frontmatter key: compatibility
```

**Not a finding against this PR.** The key is absent from the same file on `crewrig/main`, so this is a pre-existing mismatch between the bundled linter's expectations and this project's skill schema. Recording it for transparency, not routing it.

`lint-markdown.sh`, `lint-python.sh`, `lint-json.sh` — not applicable or covered by the green `lint-markdown` CI job.

---

**Summary:** 2 findings, both `class: tech` → both route to DEV. Finding 1 is the blocking one: it fails a legitimate merge request on the GitLab engine and misattributes another ticket's spec. Finding 2 is an unenforced invariant this PR's own documentation asserts. The mechanism at the heart of this change — the create-only lease, the parentless object, the sibling namespace, the origin precedence — is correct, and I could not break it.

No merge action taken; merging requires explicit human authorization per `AGENTS.md` → *Branching Strategy*.

hcross:
## Verdict: APPROVE

Second cold pass, head `db4fcaa` (previously reviewed at `e5053f5`). **Both `class: tech` findings from the first pass are genuinely closed, not moved.** I re-ran my original two reproductions verbatim against the new script, then mutation-tested each fix to confirm the suite actually catches its own regression. One finding remains, and it is in the PR body rather than the code.

### CI status

All 13 required checks **pass** on `db4fcaa` (run `31256128427`), including `check-components` — which was still pending when I began this pass and has since completed at 3m17s. No failing, no pending. Preflight clear.

### Correctness — do the fixes close what I found?

**Finding 1 (GitLab base ref) — CLOSED, and the fix is better than the one I proposed.**

I re-ran my original reproduction unchanged: `main` with no specs, `release/2.0` carrying a third party's spec `0600` (issue #900), an MR into `release/2.0` adding only `0601` secured for #726, invoked exactly as `.gitlab-ci.yml` does.

- Before (`e5053f5`): `Base ref: origin/main`, `Specs … 2`, **blocked** while naming `0600` and issue #900.
- Now (`db4fcaa`): `Base ref: origin/release/2.0`, `Specs … 1`, exit `0`.

The resolution ladder at `scripts/check-spec-id-reserved.sh:270-303` does mirror `determine_origin()`: explicit base → ambient target branch inside CI → wiring fault inside CI with neither → `<remote>/main` outside CI. I confirmed the new refusal fires with the right exit code (`2`, via `wiring_fault` at `:102-106`, correctly on stderr) rather than degrading.

Hoisting `IN_CI` to `:117-121` and consuming it in both decisions is the right call — the comment's claim that the two "must not drift apart" is the actual lesson of the original defect, and the two ladders now have one shared premise. The knock-on the developer found is real: the old `[ "$BASE_REF" = "/main" ]` guard could not survive a variable branch, and moving the emptiness test onto `base_remote_name` at `:275-278` is the correct place for it.

**Your correction is the accurate reading, and I verified it independently.** The `check-spec-id-reserved` job in `.gitlab-ci.yml` carries exactly one rule, `merge_request_event`; the two rules you first read do belong to a neighbouring job. My first-pass review quoted that single MR rule too, so nothing downstream was built on the wrong premise. And the hardening does stand on its own footing: the ladder is now correct for *any* caller that omits the argument, not only for the generated job whose rule happens to guarantee the variable.

**I agree with the developer over your suggestion on the push-pipeline exit 0.** Their two reasons are both sound, and the second is the stronger: with the job's rule restricted to `merge_request_event`, a push-trigger branch is reachable by no configuration this repository ships, so it would be code pinned by nothing — and `check-test-wiring.sh` would not notice. The first reason is the one I would have led with, though: a bare `exit 0` on an undefined trigger produces a green indistinguishable from "checked and clean", which is precisely the failure mode this PR spent a commit documenting. Adding it would have contradicted `db4fcaa` in the same breath as writing it.

**Finding 2 (`unsecured-id` mark) — CLOSED, and the fork path is intact.**

Reading the mark alongside the reservation state rather than as a trigger is the right shape. I drove all seven states directly:

| State | Same-repo | Fork |
|---|---|---|
| secured + stale mark `true` | **fails** (`rc=1`), names the file, says delete the line | **reports**, `rc=0` |
| secured + no mark | `OK … (no stale mark)`, `rc=0` | — |
| secured + mark `false` | `OK`, `rc=0` — treated as absent, matching `docs/spec-format.md` | — |
| never secured + mark `true` | fails, "correctly declares" wording | **reports**, `rc=0` |
| never secured + no mark | fails, original wording | — |

Routing through the existing `BLOCKING` flag rather than around it is the load-bearing decision, and it holds: **a fork is never blocked by this guard**, so R10's asymmetry survives. `unsecured_mark_of()` at `:471-488` correctly reduces the vocabulary to `{true, false, absent}` and `tolower`s the value, so YAML's `True`/`TRUE` do not slip through.

The message correction is right and worth having. A contributor who did exactly what R9 asks should not read an accusation; "the spec correctly declares" fixes that without softening the verdict, which stays a finding.

**Mutation testing — the fixes are pinned, not merely accompanied by tests.**

I reverted each fix in isolation and re-ran the suite:

- Restoring `BASE_REF="${base_remote_name}/main"` → `Results: 37 passed, 2 failed`, both in Case 16 (`a merge request onto release/1.x resolves its base from the target branch`, and `no third party's spec is drawn into the change under test`). That second assertion is the one that matters: it pins the *misattribution*, not just the exit code.
- Removing the stale-mark enforcement → `Results: 38 passed, 1 failed`, Case 13 and nothing else, exactly as the suite's own comment at `:555-556` predicts.

The disclosure block at `:553-568` is the best thing in this round. Stating that Case 14 "does NOT catch a mark merely routed to `add_finding` unconditionally" — and naming the property that *is* pinned — is coverage honesty I almost never see written down. It is also what let me audit this pass quickly, because I could test the claims instead of guessing at them.

### Convention compliance

Clean. Three files changed, all scripts; no `artifacts/`, `ci/`, or `docs/` touched, so no version bump, no `cli-matrix` row, and no rebuild are owed this round. Verified at `db4fcaa`: `check-core-paths`, `check-test-wiring`, `check-ci-parity`, `check-skill-versions`, `check-agents-size`, `build-ci.sh --check` all exit `0`, and `build-components.sh` produces no drift. Gitmoji on all five new commits.

**Finding — `class: tech` — the PR body's test plan is stale and now understates the change.**

The *How to test this PR?* section still reads "**86 assertions** across the two (59 + 27)" and "Every run below is green on `e5053f5`", and quotes `Results: 27 passed, 0 failed` in three places. The head is `db4fcaa` and the check suite is **39**, so the true figure is 98 across the two.

I am routing this rather than filing it as a nit for one reason: this body's entire rhetorical strategy is *verify me* — it volunteers measurements, names the commit they were taken on, and invites the reader to re-run them. That is why the first pass could confirm every claim in it. Three of those claims no longer verify, and the anchoring commit is superseded, which puts a reader who runs the documented command in the position of wondering whether they are looking at the right tree or whether twelve assertions arrived by accident. Body text only; no code change is implied.

### Test coverage

Reproduced from the pushed head, both arms:

```
test-reserve-spec-id          59 passed, 0 failed
test-check-spec-id-reserved   39 passed, 0 failed
  under CI=1 GITHUB_ACTIONS=true → 59 / 0 and 39 / 0
```

`27 → 39` covers both first-pass gaps and more: Case 16 (the release-branch defect itself, with a fixture whose `release/1.x` carries a third party's spec — the same construction I used), Case 17 (in CI with no base and no target branch → wiring fault), Case 19 (explicit base outranks the target-branch variable), and Cases 13/14/15 as the three-way mark discrimination. The reserve suite's `+19` lines are comment-only — I verified zero non-comment additions — so its `59` is unchanged behaviour, correctly.

**One observation, deliberately not routed.** No assertion pins the *wording* of the honest-mark message: I reverted it to the plain "Secure it with:" text and the suite stayed at `39 passed, 0 failed`. I am not filing this. The suite pins verdicts and file identity and states that it does; pinning prose buys a brittle test that fails on every legitimate rewording, and the reverted state changes nothing a consumer depends on. Recording it so the gap is known rather than assumed closed.

**On the CI-green note in `db4fcaa`** — recording it in the artifact rather than the review thread is the right place for it, and extending it to the reserve suite (which no CI job runs at all) is a generalisation I had not made. The framing that a badge "carries more authority than a comment" is the part worth keeping.

### Linter findings

`lint-shell.sh` on `scripts/check-spec-id-reserved.sh` at `db4fcaa` — exit `0`, no output. The pre-existing `lint-skill.sh` complaint about a missing `compatibility` key is unchanged and still not attributable to this PR; no skill source was touched this round.

---

**Summary:** 1 finding, `class: tech` → DEV. Both first-pass findings are closed, reproduced-closed, and mutation-pinned. The remaining item is three stale numbers and a superseded commit reference in the PR body.

Note for the loop: under `AGENTS.md` → *Retroactive review loop*, termination requires APPROVE **and** zero findings **and** green CI. The code is ready; a body-text edit and one confirming pass gets you there.

No merge, no approve event, no push — merging requires explicit human authorization per `AGENTS.md` → *Branching Strategy*.

hcross:
## Verdict: APPROVE

Delta pass, head `db0f6f3`, scoped as agreed: the new assertions, the body edit, and a CI re-check. The two mutation-pinned findings from pass 2 are not re-litigated — and I confirmed they need not be, because **both scripts are byte-identical to what I mutation-pinned at `db4fcaa`** (`check-spec-id-reserved.sh` → `13c06c0`, `reserve-spec-id.sh` → `c92926a`). This round changed test files only, `+28` lines, no production code.

### CI status

All 13 required checks **pass** on `db0f6f3` (run `31256638776`). `check-components` was pending when I began and has since completed. No failing, no pending.

### Correctness — do the new assertions pin, or merely accompany?

**Both pairs pin. Verified by mutation, not by reading.**

| Mutation | Result |
|---|---|
| Unreadable reserved set → bare silent `exit 0` | `39 passed, 2 failed` — both new Case 12 assertions |
| Unreachable remote → bare silent `exit 3` (no id, no marker) | `59 passed, 2 failed` — both new Case 3 assertions |

Before this commit the first mutation survived at `39/0` and the second at `59/0`, exactly as the two comments claim. The claims are measured, not asserted.

**Widening it to both report-and-continue paths was the right instinct, and the reserve-side case is the better catch of the two.** I had only identified the check-side gap. The reserve-side one is subtler for the reason its comment gives: Cases 2 and 21 *do* pin the payload, on the offline and hidden-namespace paths, which is precisely what made Case 3 look covered. I confirmed the three paths are genuinely distinct — dropping only the `unsecured-id: true` line fails Cases 2, 3 **and** 21 (`58 passed, 3 failed`), while the silent-`exit 3` mutation fails Case 3 alone. So Case 3 was the one unpinned path of the three, not a duplicate of them.

The reasoning in that comment is also the stronger version of my argument. I framed the value in terms of a reader misreading a badge; the tester framed it in terms of the *caller*: `spec-author` needs the id to name the file and the marker to write the frontmatter, so a silent `3` is indistinguishable from a crash to the one consumer that actually exists. That is a concrete broken contract rather than a legibility concern, and it is the better justification.

Asserting the log *text* rather than a bare `[REPORT]` marker is correct given what I measured last round — the four exit-`0` signatures are distinct, so matching `\[REPORT\].*could not be read` pins that branch specifically rather than any path that happens to exit `0`. The second assertion on `indistinguishable from an empty` pins the *rationale* too, which is what stops the announcement decaying into a bare marker.

### Convention compliance

**Finding — `class: tech` — three line-range citations in the PR body no longer point at what they name.**

The count correction landed cleanly; the *citations* drifted with the same commits and were not re-checked. I walked every `path:line` reference in the body against `db0f6f3`:

| Body citation | Claims to be | Actually at `db0f6f3` |
|---|---|---|
| `check-spec-id-reserved.sh:127-213` | `determine_origin` | `146-227`; line 127 is preamble comment, range ends 14 lines short of the close |
| `test-reserve-spec-id.sh:489-514` | Case 10 | Case 10 starts at `518`; line 489 is inside **Case 8b** |
| `test-check-spec-id-reserved.sh:236-247` | `run_check()` clearing env vars | `run_check()` is at `319-322`; `236-247` is the `reserve()` helper |

The remaining nine citations still resolve correctly, including `test-reserve-spec-id.sh:70-81` and `test-check-spec-id-reserved.sh:53-57`.

To be fair about cause: **all three were accurate when written.** They drifted because `IN_CI` was hoisted to `:117-121` and because content was inserted above them across `b8554a9`, `52754ec`, `2bb1310` and `db0f6f3` — that is, the fixes invalidated them. This is not carelessness. It matters anyway, because *How to read this PR?* is a navigation contract: a reviewer following citation 2 lands on the concurrency case and finds no descendant-push assertion, and one following citation 3 lands on `reserve()` and finds no variable clearing. The section exists to guide a cold reader, and two of its three signposts now point away from their subject.

**The durable fix is not three new numbers.** I would replace the brittle ranges with the symbols they name — `determine_origin()`, `Case 10`, `run_check()` — which do not drift and which `grep` resolves exactly. The nine surviving citations are stable only because `reserve-spec-id.sh` happened not to change; the next commit that touches it invalidates `:344`, `:289-290` and `:561-567` the same way. Patching numbers re-arms the same trap; naming symbols disarms it. That is a suggestion on *how*, not an additional requirement — the finding is satisfied either way.

Everything else clean at `db0f6f3`: `check-core-paths`, `check-test-wiring`, `check-ci-parity`, `check-skill-versions`, `check-agents-size`, `build-ci.sh --check` all exit `0`; `build-components.sh` produces no drift; gitmoji on the commit; no `artifacts/` change so no version bump owed. Closing references remain empty, so #726 survives the merge as the logbook anchor and is closed deliberately rather than by keyword — the right shape.

### Test coverage

Reproduced from the pushed head, both arms:

```
test-reserve-spec-id          61 passed, 0 failed
test-check-spec-id-reserved   41 passed, 0 failed
  under CI=1 GITHUB_ACTIONS=true → 61 / 0 and 41 / 0
```

102 assertions across the two, matching the corrected body exactly. The trajectory over three passes — `59+27` → `59+39` → `61+41` — closed the two behavioural gaps I found and then two coverage gaps I did not, both surfaced by the developers' own audit rather than by review. The suite that grew fastest is the one whose defects were hardest to see, which is the right direction for it to have grown.

### Linter findings

`lint-shell.sh` on both changed test scripts — exit `0`, no output. The pre-existing `lint-skill.sh` `compatibility` complaint is unchanged and still not attributable to this PR; no skill source was touched.

---

**Summary:** 1 finding, `class: tech` → DEV. Body text only; no code change implied. The mechanism, both fixes, and now both announcement halves are all pinned by tests I verified adversarially.

Under `AGENTS.md` → *Retroactive review loop*, termination needs APPROVE **and** zero findings **and** green CI. Two of three are already true.

No merge, no approve event, no push — merging requires explicit human authorization per `AGENTS.md` → *Branching Strategy*.

hcross:
## Verdict: COMMENT — supplement to the `db0f6f3` delta pass

Not a new pass. My APPROVE on `db0f6f3` stands for what it covered; this records three verifications I was asked to make rather than take on report, and **one new finding** that came out of the third. No code changed under me — still `db0f6f3`, CI 13/13.

### 1. The handed-over null result — CONFIRMED, exactly as predicted

Silencing the honest-mark `add_finding` arm at `scripts/check-spec-id-reserved.sh:547-548`:

```
FAIL  Case 15 — the mark buys no exemption from the reference repository
Results: 40 passed, 1 failed
```

One red, via Case 15, with the fork-side announcement covered transitively. The null was right: an assertion on Case 14 would have been redundant, not protective. Handing over the reproduction instead of the conclusion is what let me confirm that in one command — and a null result is exactly the kind of claim that normally evaporates unrecorded, so putting it in the message was the right instinct.

### 2. Both new mutations — CONFIRMED at 61/0, and the root cause is one capability, not two assertions

| Mutation | Result |
|---|---|
| `fail() { exit 1; }` — silent on every failure path | `61 passed, 0 failed` |
| `emit_unsecured` with all `note` lines stripped, payload kept | `61 passed, 0 failed` |

The structural diagnosis is correct and worth stating more sharply than "two gaps". The reservation suite's entire assertion vocabulary is `expect_rc`, `expect_stdout_line`, `expect_stdout_matches`, `expect_no_ref_matching` — **zero stderr assertions of any kind**. The check suite has `expect_log_matches` and `expect_log_not_matching`, which is precisely why its announcement gap was findable and this one was not. So this is **one missing harness capability**, not two missing assertions. That reframing has predictive value: any future path in this suite contracted to *say* something will be unpinned by default until the helper exists. Add the helper first; the two assertions then fall out of it.

One detail correction, since I am in the business of checking handed-over numbers: it is **14** `fail` call sites, not nine — lines 170, 176, 180, 183, 187, 190, 198, 268, 292, 316, 319, 543, 588, 593. That strengthens the point rather than changing it.

**I agree with declining to rank these alongside the Case 12 gap, and the reasoning is the important part.** Case 12's mutant exited **0** — a green pipeline over a guard that verified nothing. These two exit non-zero, so no pipeline goes green on a broken tool; what is lost is diagnostic quality, not safety. Those are different axes and collapsing them would make "surviving mutant" a severity-free label. Fixing A anyway on the grounds that R12 (*"SHALL print … the reason on failure"*) is an **unenforced requirement** rather than an unasserted nicety is the right basis, and the malformed-carrier refusal is the right example to name — that branch exists so an adopter is not left without a cause, which is the one place silence costs the most.

### 3. The `unpinned, reachability unknown` item — reachable, hermetically, and not exotic

**Answer: unpinned because missed, not because unreachable.**

`recorded_issue()` returns empty on **three** routes, and only the first needs a transport failure:

1. `fetch` refuses the ref — the route you tried to construct.
2. `git log -1 --format=%s` fails because the ref points at a **non-commit**.
3. The subject line simply does not contain `issue #`, so the `case` never matches and nothing is printed.

Route 3 needs no failure of any kind. Three lines of fixture:

```sh
OBJ=$(git commit-tree "$EMPTY_TREE" -m "hand-written reservation, no ticket recorded")
git push -q origin "${OBJ}:refs/spec-ids/0303"
# run the check → [REPORT], exit 0
```

I reproduced routes 2 and 3 against a constructed repository; both reach `add_report` and exit `0`. Route 3 is also not a contrivance — a ref written by hand, by an older tool, or under a different message convention all land there.

### 4. New finding — `class: tech` — the report message misdiagnoses the condition on both reachable routes

`scripts/check-spec-id-reserved.sh:582` says:

> id '0303' has a reservation ref on 'origin' but its recording ticket could not be read **(the reservation object could not be fetched)**. Not treated as a finding: **an unreadable record is a connectivity or credential gap**, not an unsecured id.

On route 3 the object *was* fetched — successfully — and simply records no ticket. On route 2 it *was* fetched and simply is not a commit. **The only route the message describes accurately is the one that could not be constructed.** A maintainer who hits this will be sent to check credentials and connectivity for a condition that is neither.

There is a second-order consequence worth stating, because it is the more interesting half. The decision not to fail rests on a stated premise — *"an unreadable namespace is measurably indistinguishable from an empty one"* — and on routes 2 and 3 that premise is simply false: the namespace was readable, the ref was returned, the object was fetched. There is no credential ambiguity to degrade around. The *verdict* is still defensible on other grounds (failing-open on a malformed third-party ref beats blocking a conformant change), so **I am not asking for the verdict to change — only for the reason to stop claiming a fetch failure that did not happen.** Something like "records no parseable ticket — the object was read but its message names no issue" covers all three routes honestly.

This is **pre-existing** from the original `e5053f5`, not a regression from this round. I missed it in pass 1 because I read the branch's rationale and found it coherent without constructing the states that reach it — which is the same mistake the developer made inferring the reservation suite was clean, and it is worth naming symmetrically.

---

**Outstanding for the next head: 2 findings, both `class: tech`, both text-only.**

1. The three stale line citations in the PR body (from the delta pass) — best fixed by naming symbols rather than renumbering ranges.
2. This message, at `scripts/check-spec-id-reserved.sh:582`.

Plus the two assertions already in flight, which I would now write against a new stderr-assertion helper rather than as one-offs. If the `add_report` branch gets a case too, route 3 is the cheapest fixture and pins the corrected message at the same time.

No merge, no approve event, no push — merging requires explicit human authorization per `AGENTS.md` → *Branching Strategy*.

hcross:
## Verdict: APPROVE

Final pass, head `813df7d`. The mechanism is sound, every new assertion pins, and the citation finding is closed structurally. Two items below, both `class: tech`, both text-or-test-only — **one of which is recorded as closed but is not in this head.**

### CI status

All 13 required checks **pass** on `813df7d`. No failing, no pending.

### Correctness

**Citation finding — CLOSED, structurally.** Zero `path:line` references remain in the body (regex count: `0`), all twelve now symbolic, anchored to `813df7d`, `112 assertions (68 + 44)`, closing references empty. Applying it to all twelve rather than the three that had drifted is the version that actually disarms the trap.

**Every new assertion pins. Verified by mutation, not by reading:**

| Mutation | Result |
|---|---|
| `fail() { exit 1; }` — silent on every failure path | `63 passed, **5 failed**` — Cases 4, 15, 16, 17, 22 |
| `emit_unsecured` with `note` lines stripped, payload kept | `66 passed, **2 failed**` — the Case 2 pair |
| `add_report` branch silenced | `42 passed, **2 failed**` — Case 12b |

All three previously survived. R12 is now enforced, and asserting **per branch** rather than once is the right shape for the reason given: the realistic regression is one message losing its identifying content in a refactor, not fourteen going silent together.

**The fourth route is confirmed by measurement.** A ref file written straight into the bare repository naming an object that was never created:

```
ls-remote → 0123…789  refs/spec-ids/0404          (listed — read without resolving)
fetch     → fatal: git upload-pack: not our ref   (refused)
```

That is a genuinely better construction than either of mine — routes 2 and 3 reach the branch through a *readable* object, so they never exercised the transport refusal the message actually describes. Case 12b now covers the one route that does.

**Finding 1 — `class: tech` — the `add_report` message finding is not closed in this head.**

`scripts/check-spec-id-reserved.sh` is byte-identical across `db4fcaa`, `db0f6f3` and `813df7d` — all three resolve to blob `13c06c0`. The message is unchanged:

> …its recording ticket could not be read **(the reservation object could not be fetched)**. Not treated as a finding: **an unreadable record is a connectivity or credential gap**…

With Case 12b landed, the situation is now sharper than when I raised it, not softer: the suite pins a message that is accurate for the newly-covered route and inaccurate for the two I reproduced, where the object *was* fetched and simply records no parseable ticket. Pinning the current wording makes the misdiagnosis a tested behaviour.

I read "the developer has it" as assigned rather than landed, which is fine — but it is inconsistent with "Both outstanding findings are closed", and **closed** and **deferred to a follow-up** are different states. Which one applies is a scope decision that belongs to the user's ruling, not to me; I am only reporting that the artifact does not currently match the status.

**Finding 2 — `class: tech` — the SC2287 fix is incomplete: five orphaned lines remain, still executing, still green.**

`813df7d` removed the two lines shellcheck flagged. Five more remain in `scripts/tests/test-reserve-spec-id.sh`, and a locale-neutral run of the **current head** shows every one of them executing:

```
line 420: \-\-offline: command not found
line 422: reserve-spec-id\.sh --id 0002: command not found
line 636: \-\-id: command not found
line 647: ORG 0001: command not found
line 658: ORG-0001\.lock: command not found
Results: 68 passed, 0 failed
```

The shape is identical to the two that were removed — a duplicate of the pattern in a different escaping style, left after an already-complete call:

```sh
expect_stderr_matches "Case 2 — stderr states why the id is unsecured" '[-][-]offline'
  '\-\-offline'          ← orphan, runs as a command
```

**Why the linter did not catch these, and why that matters here specifically.** SC2287 fires on *"a command name ending with `/`"*. The two it caught ended in `/`; these five do not, so shellcheck reports the file **clean** — I ran it, `0` findings at `813df7d` against `2` at `fe0558e`. The green was read as "class closed" when it meant "the subset matching one heuristic is closed."

That is the ticket's own theme landing on the harness a third time, and the sharpest instance yet: the first was a badge that meant *ran*, not *discriminated*; the second a suite that did not vet its own source; this one a **linter's green being read as coverage of a class it never claimed**. The suites already carry the warning that something can be green for a reason other than the one a reader assumes — this is that, one level up.

Scope: the assertions themselves are intact — all five sit *after* a complete call, so nothing is lost from coverage, and the mutation results above confirm they fire. The harm is five spurious errors per run plus the now-demonstrated fact that a stray command in this suite executes unnoticed. The **check** suite is clean; its visually-similar lines are line-continuation arguments, which I confirmed by the total absence of runtime errors there. The two shipped scripts are clean — `shellcheck` `0`, `bash -n` OK, no strays.

Whether this extends the ticket or becomes the follow-up is the user's call under the boundary you relayed. It is a five-line deletion, and `shellcheck` will not tell you when it is done — `LC_ALL=C bash scripts/tests/test-reserve-spec-id.sh 2>&1 | grep 'not found'` will.

### Convention compliance

Clean. Test-only diff, `+72` lines, no `artifacts/` change so no version bump owed; gitmoji on all three commits. `check-core-paths`, `check-test-wiring`, `check-ci-parity`, `check-skill-versions`, `check-agents-size`, `build-ci.sh --check` all exit `0`. Closing references empty, so #726 survives as the logbook anchor and is closed deliberately.

### Test coverage

```
test-reserve-spec-id          68 passed, 0 failed
test-check-spec-id-reserved   44 passed, 0 failed
  under CI=1 GITHUB_ACTIONS=true → 68 / 0 and 44 / 0
```

49 cases, 112 assertions, matching the body exactly. Across four passes: `59+27` → `59+39` → `61+41` → `68+44`. Two of those growths came from my findings and three from the team's own audits — the audits found more than the review did, which is the right ratio.

### Linter findings

`shellcheck` at `813df7d`: `0` findings on both shipped scripts and both suites. `lint-shell.sh`: exit `0`. The pre-existing `lint-skill.sh` `compatibility` complaint is unchanged and still not attributable to this PR.

Note for the record: `shellcheck` returning `0` on `test-reserve-spec-id.sh` is exactly the false comfort described in Finding 2 — five stray commands execute in a file the linter calls clean.

---

**Summary:** 2 findings, both `class: tech`. Neither touches the shipped mechanism: the guard, the tool, and all 112 assertions are correct and pinned. One is a five-line deletion; the other is a sentence, or an explicit deferral — either is fine, but the status should say which.

No merge, no approve event, no push — merging requires explicit human authorization per `AGENTS.md` → *Branching Strategy*.

hcross:
## Verdict: APPROVE — with one finding still open, not closed

Final pass, head `c5cb30d`. Everything claimed in this round verifies, and the message fix is better than what I proposed. **But my previous `class: tech` finding on the orphaned test lines is still live at this head**, and it is not among the two listed as closed — I believe it was conflated with the already-fixed `813df7d` item. Detail below, because this was about to go to the user as clean.

### CI status

All 13 required checks **pass** on `c5cb30d`. No failing, no pending.

### Correctness

**The message fix — closed, and your developer was right to diverge from my wording.** My suggestion (*"the object was read but its message names no issue"*) over-claims on route 1, where the fetch is refused and nothing is read at all. Their formulation names all three conditions and asserts none, which is the only thing true across the set:

> …but no ticket could be resolved from it. Three conditions produce this and the check cannot tell them apart: the object could not be fetched (transport or credentials), the ref does not point at a commit, or the commit subject names no issue.

Renaming `unreadable` → `unresolved_ticket` closes the same over-claim one layer down, which I had not thought to check. I proposed a fix that traded one inaccuracy for its mirror image; they found the formulation that has neither.

**The single-call-site correction is the more valuable finding, and it came from checking rather than inferring.** One `add_report` serves all three routes because `recorded_issue()` returns empty in all three and the caller cannot discriminate — which is exactly *why* the message must name all three rather than a shortcut around it. Case 12b going red on the corrected wording proved the inference wrong, and re-pinning on `\[REPORT\]` rather than the sentence is the right response. I verified the rewritten pair still discriminates: silencing `add_report` gives `42 passed, 2 failed`, and the file-identity assertion is what separates this branch from the global `[REPORT]` path, so the weaker text match does not weaken the case.

**Requirement 12 coverage went wider than reported.** `fail() { exit 1; }` now yields **11 reds**, not the five I measured at `813df7d` — Cases 4, 9, 12, 15, 16, 17, 22, 24, 25, 26, 27. Extending across the refusals a human actually hits is a materially stronger reading of R12 than the minimum that closed my finding.

**Finding — `class: tech` — five orphaned lines still execute at this head. Still open, not closed.**

A locale-neutral run of `c5cb30d`:

```
scripts/tests/test-reserve-spec-id.sh: line 420: \-\-offline: command not found
scripts/tests/test-reserve-spec-id.sh: line 422: reserve-spec-id\.sh --id 0002: command not found
scripts/tests/test-reserve-spec-id.sh: line 638: \-\-id: command not found
scripts/tests/test-reserve-spec-id.sh: line 649: ORG 0001: command not found
scripts/tests/test-reserve-spec-id.sh: line 660: ORG-0001\.lock: command not found
Results: 78 passed, 0 failed
```

Same five as at `813df7d`, shifted two lines by `5c42357`. `shellcheck` at this head reports **`0` findings** across all four files.

I flag the conflation explicitly because it is the mechanism, not the oversight, that matters: the closing note describes *"shellcheck caught two orphaned argument lines… a linter caught what no test could"* — that is the `813df7d` item, and it was real and is fixed. **My finding was that five more remain which SC2287 structurally cannot see**, because that check fires only on a command name ending in `/`. The two it caught ended in `/`; these five do not. Reading the linter's green as "class closed" is precisely how the remaining five survived a round that was specifically about them.

To be clear about severity, unchanged from last pass: the assertions are intact and fire, both shipped scripts are clean, the check suite is clean, and nothing here endangers the mechanism. `5c42357` added its 57 lines without introducing a sixth, so the authoring pattern is already corrected — these are residue. It is a five-line deletion, and `shellcheck` will not confirm it; `LC_ALL=C bash scripts/tests/test-reserve-spec-id.sh 2>&1 | grep 'not found'` will.

**Disposition is yours and the user's, not mine.** Under the closed-scope ruling this may legitimately become the follow-up ticket. What I am asking is narrower: it should not reach the user as *clean*, because it is not, and "deferred by decision" and "closed" are different states.

**On the deferral you asked me to judge — I endorse it.** Declining to write assertions against wording that was in flight was correct, and the single-call-site structure makes the two deferred items one correction rather than two. The residual gap is narrow and worth naming for the follow-up: all three routes share a call site, so Case 12b exercises the *path*; what stays unpinned is that routes 2 and 3 still *reach* it. If `recorded_issue()` were later changed to fail loudly on a non-commit instead of returning empty, nothing would catch it. Route 3 is a three-line fixture whenever the follow-up opens.

### Convention compliance

Clean. Body anchored to `c5cb30d`, reads `122 assertions (78 + 44)`, **zero** line citations remain, closing references empty so #726 survives as the logbook anchor. Gitmoji on all three commits; no `artifacts/` change so no version bump owed. `check-core-paths`, `check-test-wiring`, `check-ci-parity`, `check-skill-versions`, `check-agents-size`, `build-ci.sh --check` all exit `0`.

### Test coverage

```
test-reserve-spec-id          78 passed, 0 failed
test-check-spec-id-reserved   44 passed, 0 failed   (44 / 0 under CI=1 GITHUB_ACTIONS=true)
```

122 assertions, matching the body. Across five passes: `59+27` → `59+39` → `61+41` → `68+44` → `78+44`. Two growths came from my findings, three from the team's own audits, and the largest single jump — R12 from 5 reds to 11 — came from neither being satisfied with the minimum that closed a finding.

### Linter findings

`shellcheck` at `c5cb30d`: `0` across both shipped scripts and both suites. `lint-shell.sh`: exit `0`. The pre-existing `lint-skill.sh` `compatibility` complaint is unchanged and still not attributable to this PR.

That `0` is the finding above, restated: a linter's silence is not coverage of a class it never claimed.

---

**Summary:** 1 finding, `class: tech`, five lines of dead text in one test file. The mechanism, the tool, the guard, and all 122 assertions are correct and pinned — I have mutation-tested every fix across five passes and found no regression in any of them. This is genuinely good work.

No merge, no approve event, no push — merging requires explicit human authorization per `AGENTS.md` → *Branching Strategy*.

hcross:
## Verdict: APPROVE — zero findings

Final scoped pass, `c5cb30d..52e15b7`. **This one returns clean.** The orphaned-lines finding is closed, verified the way it had to be — by execution, not by the linter that could not see it.

### CI status

All 13 required checks **pass** on `52e15b7`. The job pending when you wrote has since completed. No failing, no pending.

### Correctness

**Finding closed, and closed at the class rather than the subset:**

```
LC_ALL=C bash scripts/tests/test-reserve-spec-id.sh        2>&1 | grep -c "not found"  → 0   (was 5)
LC_ALL=C bash scripts/tests/test-check-spec-id-reserved.sh 2>&1 | grep -c "not found"  → 0
```

The diff is exactly what it should be: five deletions and a header note. Every deletion removes a duplicate pattern sitting after an already-complete call; **no assertion was touched**, which I confirmed rather than assumed —

| Mutation | `c5cb30d` | `52e15b7` |
|---|---|---|
| `fail() { exit 1; }` | 11 red | **11 red** |
| `emit_unsecured` notes stripped | 2 red | **2 red** |

Identical. Counts hold at `78 / 0` and `44 / 0`. Nothing regressed and nothing was quietly dropped along with the residue.

**The header note is accurate, including the number.** Seven were orphaned; shellcheck named two; five survived. Recording the count rather than the remainder is what makes the note checkable later. And putting the *executable* check in the file — `grep -c "command not found"` MUST be 0 — is worth more than the prose around it: it is the only form of this lesson that a future maintainer can run.

**Verification hygiene, since I nearly filed a false alarm on it.** My first `shellcheck` sweep returned `203` notes, which looked alarming until I noticed my glob had pulled in every file under `scripts/tests/`. Scoped correctly, all four files in this PR are `0`, and the `203` reproduces identically on `crewrig/main` — pre-existing, in unrelated suites, untouched by this change. Recording it because a reviewer's own measurement deserves the same scrutiny I have been applying to everyone else's, and because the number will reappear for whoever runs that command next.

`bash scripts/check-bash32-portability.sh` — OK.

### Convention compliance

Clean. Body anchored to `52e15b7` with **zero** stale anchors, **zero** line citations, `122 assertions (78 + 44)`, closing references empty so #726 survives as the logbook anchor. Gitmoji on the commit. Test-file-only diff, so no version bump owed. `check-core-paths`, `check-test-wiring`, `check-ci-parity`, `check-skill-versions`, `check-agents-size`, `build-ci.sh --check` all exit `0`.

### Test coverage

```
test-reserve-spec-id          78 passed, 0 failed
test-check-spec-id-reserved   44 passed, 0 failed
```

Unchanged from `c5cb30d`, as a pure deletion should leave them.

**On the four deferrals — I endorse all three substantive ones, with one ranking note.** You asked me to judge rather than discover, so:

Items 1 and 3 are absences: a missing routing assertion, and a missing pin on the invariant the Case 12b discriminator rests on. Deferring an absence costs nothing beyond the gap staying open, and both are correctly characterized.

**Item 2 is different in kind and should lead the follow-up.** A missing assertion is *absence*; a comment that states a false structural fact is *misinformation*, and this one is false in the direction that tells a future maintainer a trap-arming change is impossible when it is not. Of the three it is the only one with negative rather than zero value, and it is a one-sentence edit. I am not reopening scope — deferring it is legitimate and the call is the user's — but it should be first in the follow-up, and it pairs naturally with item 3, which pins the very invariant item 2 misdescribes. Landing them together is one coherent change; landing item 3 alone would leave a passing assertion next to a comment contradicting it.

### Linter findings

`shellcheck` on all four files in this PR: `0`. `lint-shell.sh`: exit `0`. The pre-existing `lint-skill.sh` `compatibility` complaint is unchanged and still not attributable to this PR.

And the standing caveat, now recorded in the suite itself: that `0` is necessary and not sufficient for `test-reserve-spec-id.sh`. The execution check is the one that decides.

---

**Summary: zero findings.** Under `AGENTS.md` → *Retroactive review loop*, termination requires APPROVE **and** zero findings **and** green CI on the reviewed head. All three now hold at `52e15b7`.

Across six passes I raised five findings and mutation-tested every fix; none regressed, and the team's own audits found more than the review did. The mechanism — the create-only lease, the parentless object, the sibling namespace, the origin ladder, the base-ref ladder — has been pinned three times over and I could not break it at any point.

Merge authorization remains the user's under `AGENTS.md` → *Branching Strategy*. I have not merged, approved, or pushed.

hcross:
## Verdict: APPROVE — zero blocking findings

Final scoped pass. My `c5cb30d..52e15b7` pass already returned clean, so this audits `52e15b7..7236e19`: ten comment lines in one test file. **Clean, with one new non-blocking item routed to the follow-up** under the user's standing ruling rather than looped back here.

### CI status

All 13 required checks **pass** on `7236e19`. No failing, no pending.

### Correctness

**The orphaned-lines finding stays closed**, re-verified at this head: `grep -c "not found"` → **0** in both suites, `78 / 0` and `44 / 0`. Removing them **by predicate** rather than by the list a tool reported is the difference that matters — the first attempt closed what shellcheck named, the second closed what the predicate matches, and only the second is a class.

**The tester's "never committed" claim is correct, and I checked it rather than took it:**

```
git log --all -S'before candidate enumeration' -- scripts/ → (no commits)
```

The false ordering claim existed only in a message. A follow-up item saying "correct the recorded reason" would indeed have sent someone hunting a line that does not exist — and catching that is a better outcome than fixing what was reported.

**`7236e19` records the right thing.** The substantive claim is accurate: the namespace-unreadable branch runs after enumeration, the candidate list is in scope, and nothing structural stops it from naming the specs it could not verify. Writing the *fragility and the remedy* instead of an assurance is the correct inversion — and naming "name the specs we could not verify" as a plausible improvement is what makes it useful, because that is exactly the change someone will propose in good faith.

Ratifying it past the instruction given was right. A rationale consulted instead of the code is worse than none, and this comment is the case where that principle applies most directly.

**New item — `class: tech`, non-blocking, routed to the follow-up — the comment anchors its warning on line numbers.**

It cites `(:416 then :425)` into `scripts/check-spec-id-reserved.sh`. Two problems, in ascending order of importance:

1. `:416` is imprecise for what it names. Enumeration is the loop that *ends* at `:411`; `:416` is the `note` reporting `$candidate_count`. It does demonstrate the point — the count is in scope there — but the sentence reads as though `:416` is the enumeration.
2. Both numbers will drift. `check-spec-id-reserved.sh` changed in three of the last six commits; any insertion above `:411` invalidates them silently.

This is the citation-brittleness class I raised two passes ago, which you then closed across all twelve body references by switching to symbols — reintroduced one commit later, in a code comment. The irony is worth naming precisely because it is *not* carelessness: the comment exists to warn a future maintainer that a structural assumption may quietly stop holding, and it anchors that warning on the one reference form guaranteed to quietly stop holding. Symbols carry it with no drift — *"the namespace-unreadable branch runs after the candidate-enumeration loop"* needs no numbers at all.

Zero behavioural risk; a comment in a test file. Under the user's ruling this is a follow-up item, not a loop back here, and I am filing it as such rather than withholding it to preserve a clean sheet.

### Convention compliance

Clean. Comment-only diff, no version bump owed, gitmoji on the commit, body still anchored correctly with zero line citations and zero stale anchors, closing references empty.

### Test coverage

`78 / 0` and `44 / 0`, unchanged — as a comment-only commit should leave them.

**On the four deferrals, plus intelligence that makes one of them cheaper.**

I endorse all four. Two are now better-informed:

- **The sibling-suite sweep comes back clean — I ran it.** Every `scripts/tests/*.sh` and `scripts/*.sh` under a static orphan predicate surfaces exactly three candidates, all in `test-mempalace-version-range.sh`, and all three are `EXCLUDES=( … )` array members, not orphans. Runtime confirms: zero strays. So that deferred item is verification, not repair.
- **The CI guard's predicate is the part worth designing carefully**, and my scan just demonstrated why: a static *"bare quoted line in command position"* predicate false-positives on array literals — three hits, all benign. The execution check has no such ambiguity. If the guard is static it needs an array-context exclusion; if it is `LC_ALL=C … | grep -c "command not found"` it needs none. The runtime form is the sound one, and it is the one already recorded in the suite header.

The route-3 routing assertion and the never-names-a-spec invariant assertion both stand as correctly-characterized absences. The latter now pairs naturally with `7236e19`'s comment — the comment describes the fragility, the assertion would detect it.

### Linter findings

`shellcheck` on all four files in this PR: `0`. `lint-shell.sh`: exit `0`. The pre-existing `lint-skill.sh` `compatibility` complaint is unchanged and not attributable to this PR. The standing caveat holds and is now recorded in the file: that `0` is necessary and not sufficient here; the execution check decides.

---

**Summary: zero blocking findings.** Termination requires APPROVE, zero findings, and green CI — the first and third hold, and the one open item routes to the follow-up rather than here, which is precisely what the user's ruling exists to do. It is your call and the user's whether that satisfies termination or wants one comment edit first; it is two clauses either way.

**On the correction you made rather than left:** measuring the wrong revision of the record is the same class this ticket has been cataloguing, and you caught it in the summary of the thing being catalogued — which is where it is hardest to see. Making it unprompted is worth more than not having made it.

Seven passes, six findings, every fix mutation-tested, none regressed. The mechanism has been pinned three times over and I could not break it at any point. Merge authorization remains the user's; I have not merged, approved, or pushed.

PR #742, opened 2026-08-08
✨ Put the shared MemPalace MCP HTTP daemon into service

hcross:
Concurrent agent sessions all write to shared memory instead of every session but the first being refused for its whole life. Implements spec 0113 + delta-01, realising ADR 0016's derived specs 1 and 2.

## How to read this PR?

Three commits, in dependency order. Read them in sequence — the second assumes the first.

1. **`cd6a735` — the daemon.** Start at `scripts/lib/mcp-daemon-launcher.sh`: it carries the three guarantees no unit file can provide, and its header explains each. The most important is that it **refuses to start without a token** — see *The security defect this closes* below. Then `install_daemon_supervisor` in `common.sh` (the generalisation of `install_chroma_daemon`), then the two unit files.
2. **`3647007` — the switch.** `switch_assistants_to_http` in `common.sh` is the transaction. `register_mempalace_mcp` is why it had to be built rather than reused.
3. **`dc45bc6` — the tests.** `scripts/tests/test-mcp-daemon.sh`, starting with its isolation contract comment.

Non-obvious decisions, each with its reasoning at the call site: the launcher lives **outside the repository** (`~/.crewrig/`) so a revert cannot delete the program the units name; `stop` and `uninstall` are **separate scripts** because under a supervisor a stop is a restart request; ordering is **enforced in the launcher** rather than declared in the units, because launchd has no dependency ordering between user agents.

## How to test this PR?

```sh
bash scripts/tests/test-mcp-daemon.sh      # 21 assertions, isolated 3 ways
bash scripts/tests/test-chroma-server.sh   # 15 assertions, no regression
bash scripts/check-test-wiring.sh
bash scripts/check-ci-parity.sh
bash scripts/build-ci.sh --check
bash scripts/check-bash32-portability.sh
```

All green on the head commit. The chroma suite was run against this branch and the production daemon was verified untouched afterwards (same PID, heartbeat OK) — see *Isolation* below for why that verification matters.

To exercise it for real (this changes your machine):

```sh
task mempalace:switch-http    # all-or-nothing across every assistant present
task mempalace:status         # daemon, auth enforcement, drift, per-assistant state
```

Sessions already running keep their previous memory server until restarted; the switch says so explicitly.

### The security defect this closes

The launcher refuses to start when the bearer token file is absent or empty, exiting non-zero with a diagnostic. This is not defensive tidiness. MemPalace requires a token only on a **non-loopback** bind (`mcp_server.py:5210-5218`), and its auth gate reads `if require_auth and srv.auth_token:` (`:5291`) — an empty token short-circuits the entire bearer check. A daemon started without one serves happily, all four CLIs work, and the `Authorization` header they send is ignored. Authentication would be off with nothing reporting it.

`status-mcp-server.sh` verifies enforcement rather than assuming it: one **unauthenticated POST to `/mcp`** must return 401. It cannot use `/healthz`, which is served with `require_auth=False` (`:5301-5303`) and returns 200 in every state including the broken one.

### The transaction

`switch_assistants_to_http` establishes R12's floor first — every present assistant's config readable, writable, parseable — then captures, backs up, applies, and restores on failure. The floor exists because `claude mcp add` **cannot be pre-flighted**: no `--dry-run`, `--check` or `--validate` exists (verified against `--help`), so success cannot be determined without attempting. R13's restore covers what the floor cannot, and R14 reports which assistant is in which arrangement when a restore itself fails — `common.sh:343` already ships a live `could not auto-restore` branch, so that path is real rather than theoretical.

An assistant found in **neither** arrangement halts the run before anything is applied: repetition cannot resolve a configuration it cannot recognise, so it is reported (R14) rather than silently converged (R15). Claude's config write is `claude mcp add`'s own and carries no atomicity guarantee, unlike the framework's `jq > tmp && mv` (`common.sh:118/262/786`) — which is what makes that state reachable at all.

### Why the registration surface is new code

`apply_org_mcp_servers` (spec 0091) skips every name in `MCP_RESERVED_NAMES`, and `mempalace` is reserved — the org appliers deliberately refuse to touch the one server this ticket must register. And `mcp_register_user` (`setup-claude-interactive.sh:170`) takes `-- "$@"`, a stdio argv shape with nowhere to put a transport or a header. One helper now emits all four native shapes.

### Isolation

The new suite isolates on **label, port and `$HOME`**, and refuses to run if any resolves to production. The port is the axis a label guard does not watch and the one that matters here: this tier is reached over HTTP, so a correctly test-labelled unit that fails to bind because production holds the socket leaves the clients connecting to production — and the writes land in the operator's real palace with every assertion passing.

### Two defects found by these tests

- **The token key was a function of the filesystem, not the path.** `cd` fails on a missing parent, so the palace path fell back to its unresolved form; a later call, after another step created the directory, resolved differently and computed a different key — two token files for one palace, observed as a token that keeps changing.
- **The isolation guard compared `$HOME` after overriding it**, so the equality was always true and the guard refused every run — protecting by permanently skipping the tests it exists to protect.

### Also fixed

`config/systemd/mempalace-chroma-server.service:9` hardcoded `python3.13`; it is now substituted like its launchd sibling. `stop-chroma-server.sh` reported "not running (no PID file)" for a supervisor-managed daemon that was in fact running — it now says so and names the uninstall, without issuing any supervisor verb (a routine stop must not cancel an operator's autostart).

### Not done

Derived spec 3 (recovery beyond repetition), derived spec 4 (the transcript hook's write path, gated on a latency measurement that only becomes possible once this ships) and derived spec 5 (runbook, README) stay separate, per ADR 0016.

Refs #739


hcross:
## Verdict: REQUEST CHANGES

Cold review of PR #742 at head `88e751b`. No authoring context used.

> **Scope note.** This review audits commit `88e751b`, the PR head on GitHub. During the review the worktree accumulated uncommitted edits to `scripts/lib/common.sh`, `scripts/lib/mcp-daemon-launcher.sh`, `scripts/switch-mempalace-http.sh`, `scripts/uninstall-mcp-daemon.sh` and `scripts/tests/test-mcp-daemon.sh`. Those are **not** part of this PR and were excluded — every finding below was re-verified against a pristine `git show 88e751b:` extract. Some of that in-flight work appears to overlap finding **F6**.

The design is strong: the launcher's three stated guarantees are real, `status-mcp-server.sh` verifies auth enforcement by POSTing unauthenticated to `/mcp` and requiring 401 rather than trusting `/healthz` (`status-mcp-server.sh:57-70`), the drift check compares the recorded *source* hash rather than installed bytes, and the install/uninstall verb symmetry (`load -w` ↔ `unload -w`, `enable --now` ↔ `disable --now`) is correct and well-argued. The isolation contract in the test suite is the best part of the diff.

But the transaction has three reachable defects that lose an operator's memory configuration silently, the token reaches the process table on the registration path, and the requirement the ticket exists to satisfy has no behavioural coverage at all.

---

### 1. CI status — all green

13/13 required checks **pass** on `88e751b` (`build`, `check-components`, `check-ci-parity`, `check-skill-versions`, `lint-markdown`, `lint-specs`, `grep-anti-patterns`, `gitlab-ci-check`, and 5 others). No failing or pending check. CI does not gate this verdict; the findings below do.

Repository self-checks re-run locally, all clean: `check-test-wiring.sh` (64 wired, 1 honestly exempted), `check-ci-parity.sh`, `check-bash32-portability.sh` (147 files, no Bash 4+ construct).

---

### 2. Correctness

#### F1 — `mcp_assistant_arrangement claude` reports `http` for a Claude that is on stdio · **BLOCKING** · `class: tech`

`scripts/lib/common.sh:1075` classifies Claude by grepping `claude mcp get` output:

```sh
if claude mcp get mempalace 2>/dev/null | grep -qiE 'http|url|/mcp'; then
```

The stdio registration this ticket replaces is `claude mcp add --scope user mempalace -- bash …/tls-exec.sh <python> …/scripts/lib/mempalace-http-wrapper.py` (`setup-claude-interactive.sh:231`). The args contain **`mempalace-http-wrapper.py`**, so `grep -i http` matches and the stdio entry is reported as `http`.

Reproduced live on the authoring machine, unmodified:

```
$ claude mcp get mempalace
  Type: stdio
  Command: bash
  Args: …/scripts/lib/mempalace-http-wrapper.py
$ claude mcp get mempalace | grep -qiE 'http|url|/mcp' && echo MATCHES
MATCHES          # → reports "http" for a demonstrably stdio registration
```

Consequences, all on the primary CLI:

- **R16 is violated** — the operator asks which arrangement Claude is on and is told the wrong one. `status-mcp-server.sh:92-95` and `doctor-mempalace.sh` both consume this reader, so the comment at `common.sh:1062` ("so the two cannot drift apart") is satisfied only in the sense that both report the same wrong answer.
- **R15's partial-state NOTE is defeated** — `switch_assistants_to_http` sets `had_http=1` for a stdio Claude (`common.sh:944`), so a genuinely mixed machine never prints the notice, and a machine entirely on stdio prints it spuriously as soon as any other assistant is present.

The three JSON-based CLIs are fine: `has("url")`/`has("command")` on the parsed node is exact (`common.sh:1094-1096`). Only the Claude branch is textual, and it is the one branch the suite never exercises.

#### F2 — a failed `claude` registration is never rolled back and never reported · **BLOCKING** · `class: tech`

`register_mempalace_mcp` removes before it adds (`common.sh:883-885`):

```sh
claude mcp remove --scope user mempalace >/dev/null 2>&1 || true
claude mcp add --scope user --transport http mempalace "$url" \
  --header "Authorization: Bearer ${token}" >/dev/null 2>&1 || return 1
```

When `add` fails, the `remove` has already landed. The caller then excludes the *failing* CLI from the rollback list — `applied` only ever accumulates CLIs that **succeeded** (`common.sh:1004`), and the failure branch passes that same list to `_switch_rollback` (`common.sh:1007`). So the one assistant that was actually left mid-change is the one the rollback skips.

Reproduced against a pristine `88e751b` extract with a stubbed `claude` whose `remove` succeeds and `add` fails:

```
BEFORE: {"type":"stdio","command":"bash","args":["ORIGINAL"]}
  Backed up: .claude.json -> .claude.json.bak.20260808-190901
  ERROR: claude could not be switched.
  Restoring the assistants already changed in this run... (R13)
rc=1
AFTER:  "<<GONE>>"
```

Claude ends with **no mempalace registration at all** — neither arrangement. Three obligations missed at once:

- **R13** — "restore every assistant it has already changed". Claude was changed; it is not restored.
- **R14** — the `*** MANUAL REPAIR REQUIRED ***` block is gated on `failed` being non-empty (`common.sh:1035`), and `failed` only collects entries from the loop that skipped Claude. Nothing is reported.
- The operator is not even shown the arrangement table: `switch-mempalace-http.sh:36-38` does `exit 1` on a non-zero return, so the `status-mcp-server.sh` call at line 48 never runs on the failure path.

The PR body reasons carefully about `claude mcp add` having "no atomicity guarantee, unlike the framework's `jq > tmp && mv`" — that analysis is right, and it is precisely the case the rollback list does not cover. Suggested shape: add the CLI to a `changed` list **before** calling `register_mempalace_mcp`, and roll back from `changed`, not `applied`.

#### F3 — the bearer token is passed in argv on the registration path · **BLOCKING** · `class: tech`

`common.sh:885` puts the credential in `claude`'s command line:

```sh
--header "Authorization: Bearer ${token}"
```

R8 states the credential "SHALL NOT be observable in the process list". On Linux `/proc/<pid>/cmdline` is world-readable by default, so for the duration of the call the token is legible to **any user on the machine**, not merely the same user who can already read the `0600` file. On macOS the exposure is same-user.

This is the exact property the launcher is architected around — `mcp-daemon-launcher.sh:18-22` explains at length that "an exported variable is not argv", and `test-mcp-daemon.sh:129-132` asserts it. The assertion inspects the launcher only; the registration path violates the same requirement two functions away and nothing looks at it. The three JSON CLIs write the header into a file and are clean.

`claude mcp add` reads `--header` from argv with no documented stdin or env alternative, so this may need an upstream-shaped workaround (e.g. write the entry into `~/.claude.json` with the same `jq > tmp && mv` used for the others, which capture/restore already assume — `capture_mempalace_registration` reads `.mcpServers.mempalace` from that very file, `common.sh:848-856`).

#### F4 — the R12 floor does not cover an assistant whose config file is absent · **BLOCKING** · `class: tech`

The floor's checks are nested inside `if [ -f "$cfg" ]` (`common.sh:972`), so an assistant that is *present* (`command -v` succeeds) but has no config file passes the floor untested — and then fails during apply, because `register_mempalace_mcp` requires the file (`common.sh:890`, `:902`).

Reproduced against pristine `88e751b` (gemini configured, copilot CLI present with no `~/.copilot/mcp-config.json`):

```
  Backed up: settings.json -> settings.json.bak.…
  gemini: switched to the shared daemon      ← a change was applied
  ERROR: copilot could not be switched.
  Restoring the assistants already changed in this run... (R13)
    gemini: restored
rc=1
```

R12 requires: *"When setup establishes, before it has applied any change, that a required change cannot be applied, it SHALL apply none."* This condition is establishable before applying — the file is simply not there — and the code applies a change anyway. delta-01's first scenario states the expected behaviour verbatim ("Then no assistant SHALL be changed"). R13 rescues the end state, so this is a conformance defect rather than data loss, but it is a footgun on any machine with a freshly-installed CLI.

#### F5 — the switch is not durable: the setup scripts still register stdio · **BLOCKING** · `class: arch`

No `scripts/setup-*-interactive.sh` is modified by this PR (`git diff --name-only crewrig/main...88e751b | grep setup-` → empty), and the shipped templates still carry the stdio shape:

```
config/gemini/settings.json          → {"command":"python3","args":["…/mempalace-http-wrapper.py"]}
config/copilot/mcp-config.json.template → {"type":"stdio","command":"python3",…}
```

`setup-gemini-interactive.sh:159-163` copies that template over `~/.gemini/settings.json`. Because `mempalace` is in `MCP_RESERVED_NAMES` (`common.sh:45`), `merge_preexisting_mcp_servers` deliberately does **not** preserve the operator's entry — framework wins. So running any setup script after `task mempalace:switch-http` silently reverts that assistant to stdio, producing exactly the half-switched machine delta-01 calls "worse than either end state".

ADR 0016 — which this PR states it realises — makes this explicit in *Derived spec plan* step 2:

> **Four-CLI registration** — stdio → HTTP **in every setup script**, token provisioning, no committed secret.

The PR body claims to implement "ADR 0016's derived specs 1 and 2". Derived spec 2 is only half-implemented: a new standalone entry point exists, but the setup scripts named by the ADR are untouched. This bears on R3 ("**Setup** SHALL configure every supported assistant to reach shared memory the same way") and R11 ("no assistant remains able to reach shared memory the previous way after setup completes"). Classed `arch` because the fix is a placement decision — fold the switch into the four setup paths, or record why a separate entry point discharges R3 — not a local code repair.

---

### 3. Convention compliance

Clean on nearly everything checked against `AGENTS.md`:

- **Gitmoji** — PR title `✨ Put the shared MemPalace MCP HTTP daemon into service` and all five commits conform. No conventional-commit prefixes.
- **Branch naming** — `feat/0739-shared-mempalace-mcp-daemon` matches `<prefix>/<NNNN>-<slug>`.
- **Spec-PR ordering** — `specs/0113-…md` (#740) and `…delta-01.md` (#741) are both merged on `main` ahead of this implementation PR. Correct.
- **Version-bump convention** — no `artifacts/**` path touched, so the rule and `build-components.sh` are both N/A. `check-skill-versions` passes.
- **CLI Matrix Maintenance** — `docs/cli-matrix.md` rows 7c and 7d are substantively updated, as delta-01 explicitly requires (it removed the matrix from `## Out of scope`). Good.
- **Logbook** — `Refs #739`, Rule A satisfied (the feature issue is the logbook). No close-keyword on a spec issue.

Two nits:

#### F10 — `docs/cli-matrix.md` row 7d contradicts itself · non-blocking · `class: tech`

The mechanism cell now reads *"the four CLIs no longer each spawn a stdio memory server"*, but all four per-CLI cells in the same row still document exactly that: *"MCP entry → `bash <repo>/scripts/lib/tls-exec.sh <mempalace-python> …/mempalace-http-wrapper.py`"*. A reader consulting the Claude column is told the old arrangement. Given F5, the per-CLI cells are currently the *accurate* half — but the row cannot assert both.

#### F11 — PR body says "Three commits" · non-blocking · `class: tech`

*How to read this PR?* says "Three commits, in dependency order"; the PR carries five (`354ef0a` and `88e751b` are follow-up fixes). Minor, but the section is a navigation aid.

---

### 4. Test coverage

The suite is genuinely well-isolated — the three-axis contract (label, port, `$HOME`) and the reasoning about why the *port* is the axis that matters are correct and worth keeping. The count in the PR body is accurate: 21 assertions, 21 pass (verified by running it).

The problem is not count, it is what can fail.

#### F6 — the two headline security assertions pass against a launcher with the guard deleted · **non-blocking but high-value** · `class: tech`

`test-mcp-daemon.sh:104-115` asserts only that `bash "${launcher}"` exits non-zero. But the launcher has a *second* non-zero exit downstream: the ChromaDB wait deadline (`mcp-daemon-launcher.sh:115-123`). Mutation-tested — I materialised the launcher, deleted **both** token guards, and ran it:

```
guard lines remaining: 0
mutant exit=1        ← the test still reports "ok launcher refuses to start"
```

So the assertion carrying the PR's headline security claim cannot distinguish the guard from its absence. It should assert on the diagnostic — `die` emits `bearer token file not found` / `is empty` — not on the exit status. The same weakness means the suite cannot detect a divergence in the token-path derivation (F8): the launcher looking at an entirely different, never-created path also exits non-zero.

#### F7 — one assertion cannot fail at all · non-blocking · `class: tech`

`test-mcp-daemon.sh:133-135`, "no token value is baked into the launcher", greps the materialised launcher for `$tok`. `install_mcp_launcher`'s `sed` has seven substitution rules and none involves the token (`common.sh:~770-782`), so no implementation reachable from this code path could bake it in. It passes vacuously.

#### F9 — R1 has no behavioural coverage · **non-blocking, but this is the ticket's reason to exist** · `class: tech`

Stated plainly, since it was asked: **nothing in this diff proves R1.** No test in the repository starts the MCP daemon or performs two concurrent writes. The committed suite contains zero references to `install_mcp_daemon`, `switch_assistants_to_http`, `launchctl`, `systemctl`, or any HTTP request to the daemon (`grep -c` → 0). Every one of the 21 assertions is structural: file contents, `jq` shapes, a permission bit, a hash comparison. The `-32001` contention this ticket exists to eliminate is never reproduced, and the fix is never demonstrated to eliminate it.

Requirement-by-requirement, the untested set is systematic rather than incidental:

| Req | Covered by a test? |
|---|---|
| R1 concurrent writes succeed | **No** — nothing starts the daemon |
| R2 auto-restart | **No** — `KeepAlive` (plist:36-37) and `Restart=always` (service:21-22) are present but unasserted; deleting either ships green |
| R4 all-or-nothing | **No** — `switch_assistants_to_http` is never invoked by any test |
| R7 single data-directory owner (ADR 0006) | **No** on the daemon path |
| R12 / R14 / R15 floor, repair report, repeat | **No** |
| R13 restore | Primitives only, gemini only (`test-mcp-daemon.sh:186-200`); the rollback *orchestration* — where F2 lives — is untested |
| R16 arrangement reader | Four states, **gemini only**; the Claude branch (where F1 lives) is untested |
| R8 token | Yes, and this is the best-covered area — modulo F3, F6, F7 |

The pattern: the suite covers what was easy to isolate in-process and skips every part where the ticket's risk actually lives. The two defects that matter most (F1, F2) are both in code paths with zero assertions.

---

### 5. Linter findings

`lint-shell.sh` over the ten changed shell files — verbatim, only items attributable to this diff:

```
scripts/lib/common.sh: not executable in git (mode 100644; expected 100755)
scripts/lib/common.sh: missing 'set -e' (or 'set -euo pipefail') near the top

In scripts/lib/common.sh line 606:
          && echo "  Loaded launchd agent: $label" \
          ^-- SC2015 (info): Note that A && B || C is not if-then-else. C may run when A is true.

In scripts/lib/common.sh line 622:
        && echo "  Enabled and started: ${systemd_unit}.service" \
        ^-- SC2015 (info): Note that A && B || C is not if-then-else. C may run when A is true.
```

Assessment: the first two are correct-as-is (`common.sh` is a sourced library — it should be neither executable nor `set -e`). The two `SC2015` are carried over verbatim from the pre-existing `install_chroma_daemon` into the generalised `install_daemon_supervisor`; behaviour is unchanged, informational only.

The one `SC2088` shellcheck reports (`common.sh:1254`, tilde in quotes) is **pre-existing** — introduced by `57a1a6b`, not by any commit in this PR. Not attributed here.

`lint-markdown.sh docs/cli-matrix.md` → clean, no output. `lint-json.sh` / `lint-python.sh` → no matching files in the diff.

#### F8 — the launcher duplicates the token-path derivation without its fix · non-blocking · `class: tech`

`mcp_token_path` in `common.sh:1195` carries an explicit fix — `mkdir -p "$(dirname "$palace_path")"` before resolving — with a comment explaining that `cd` fails on a missing parent and yields a different key. `mcp-daemon-launcher.sh:60-69` reimplements the same derivation and **has no `mkdir`** (`grep -c mkdir` → 0 on the committed blob).

The launcher's fallback is also inoperative. `resolved="$(cd … && pwd)/$(basename …)" || resolved="${palace}"` (`mcp-daemon-launcher.sh:62`) — the assignment's exit status is that of the *last* command substitution (`basename`, always 0), so the `||` never fires. Verified on both Bash 5 and Bash 3.2:

```
$ p=/nonexistent/deep/palace
$ r="$(cd "$(dirname "$p")" 2>/dev/null && pwd)/$(basename "$p")" || r="$p"; echo "$r"
/palace          # not the intended fallback
```

The two derivations agree only while the palace parent exists — which `common.sh` now guarantees at install time, so this is latent rather than live. But it is a security-critical, fail-closed path with two copies and one fix, and per F6 the suite cannot detect the divergence. Worth extracting one helper, or having the launcher accept the path from the installer (`MEMPALACE_MCP_TOKEN_FILE` is already honoured at `:59`).

---

### Summary

| # | Finding | Class | Blocking |
|---|---|---|---|
| F1 | Claude on stdio is reported as `http` (`common.sh:1075`) — breaks R16, defeats R15 | `tech` | **yes** |
| F2 | Failed `claude` registration: no rollback, no report, registration lost (`common.sh:883-885`, `:1004-1007`) — breaks R13/R14 | `tech` | **yes** |
| F3 | Bearer token in argv on the registration path (`common.sh:885`) — breaks R8 | `tech` | **yes** |
| F4 | R12 floor skips an assistant with no config file (`common.sh:972`) — applies then rolls back | `tech` | **yes** |
| F5 | Setup scripts untouched, templates still stdio — reverts the switch; ADR 0016 derived spec 2 says "every setup script" | `arch` | **yes** |
| F6 | Token-refusal assertions pass with the guard deleted (`test-mcp-daemon.sh:104-115`) | `tech` | no |
| F7 | "no token value is baked" assertion cannot fail (`test-mcp-daemon.sh:133-135`) | `tech` | no |
| F8 | Launcher duplicates token-path derivation without the fix; dead `||` fallback (`mcp-daemon-launcher.sh:62`) | `tech` | no |
| F9 | R1 has no behavioural coverage; R2/R4/R7/R12/R14/R15 untested | `tech` | no |
| F10 | `cli-matrix.md` row 7d mechanism cell contradicts its four CLI cells | `tech` | no |
| F11 | PR body says "Three commits"; there are five | `tech` | no |

F1–F4 are local repairs. F5 is a placement decision that needs a call before it is coded. F9 is the one I would weigh hardest against merging on schedule: the diff ships a well-argued mechanism for a problem it never demonstrates it solves, and F1 and F2 both sit in the exact regions that have no assertions — which is how they survived to review.

hcross:
## Verdict addendum — REQUEST CHANGES (unchanged), findings F12–F18

A second, independent pass over the daemon/supervisor surface (`status-`/`stop-`/`uninstall-`/`switch-` scripts, unit files, launcher↔wrapper handoff) finished after my first comment and surfaced defects that pass missed. The verdict is unchanged, but the finding set above was **incomplete** — and one of the additions is the most severe security finding in the review. Correcting the record rather than leaving it to be discovered at merge.

Still against `88e751b`; the uncommitted worktree edits remain out of scope, with one exception called out in F13.

---

### F12 — the switch writes the bearer token into a **world-readable** file, and downgrades two configs that were `0600` · **BLOCKING** · `class: tech`

`register_mempalace_mcp` writes the registration with `jq … > "${cfg}.tmp.$$"` then `mv` (`common.sh:891-895` gemini/copilot, `:903-907` antigravity). The temp file is created at the caller's umask — `022` by default — so it is `0644`, and `mv` carries that mode onto the target. `restore_mempalace_registration` uses the identical pattern (`common.sh:862-868`), so the **rollback path downgrades `~/.claude.json` too**.

Verified end to end on this machine:

```
$ umask                                  022
$ stat -f '%Lp' cfg.json                 600      # before
$ jq … > "$cfg.tmp.$$"; stat -f '%Lp' …  644      # temp inherits umask
$ mv "$cfg.tmp.$$" cfg.json
$ stat -f '%Lp' cfg.json                 644      # target downgraded
$ grep -o 'Bearer SECRETTOKEN' cfg.json  Bearer SECRETTOKEN
```

Current modes on the authoring machine:

```
600  ~/.claude.json              ← would be downgraded to 644
600  ~/.gemini/settings.json     ← would be downgraded to 644
644  ~/.copilot/mcp-config.json
644  ~/.gemini/config/mcp_config.json
```

So the transaction takes a palace credential and puts it in a file any local user can read, and it makes two files *less* safe than it found them. This is strictly worse than F3: argv exposure lasts for the duration of one `exec`, this persists on disk.

It also falsifies the launcher's headline guarantee verbatim — `mcp-daemon-launcher.sh:18-22`, "**THE TOKEN NEVER TOUCHES A WORLD-READABLE FILE**". That is true of the unit file it is written about, and false of the four registrations the same transaction writes minutes earlier. Fix: `umask 077` around the writes, or `chmod 600 "$tmp"` before every `mv`. (`backup_file`'s `cp -P` does preserve the source mode, so the `.bak` copies are not an additional leak.)

### F13 — at `88e751b` a whitespace-only token file defeats the fail-closed guard · **BLOCKING** · `class: tech`

The guard at `88e751b:78-85` tests emptiness of the raw file contents:

```sh
MEMPALACE_MCP_HTTP_TOKEN="$(cat "${TOKEN_FILE}" 2>/dev/null || true)"
if [ -z "${MEMPALACE_MCP_HTTP_TOKEN}" ]; then
```

A file containing `"   "` is non-empty to the shell and passes. Upstream then strips it — `mcp_server.py:5208`, `os.environ.get("MEMPALACE_MCP_HTTP_TOKEN", "").strip()` — yielding `""`, and the auth gate is `if require_auth and srv.auth_token:` (`:5291`), which short-circuits on the empty string. The daemon serves the whole palace unauthenticated while the launcher reports success.

That is precisely the defect the launcher exists to prevent, reachable through the guard meant to prevent it. Worth noting the shape of the miss: the guard tests the value the *shell* sees, while the security property depends on the value the *server* sees, and `.strip()` sits between them.

**This one is already fixed in the worktree** — the uncommitted launcher adds `tr -d '[:space:]'` plus charset and ≥32-character shape validation. It is not in `88e751b`, so it is not in this PR. Commit and push it.

### F14 — the launcher never propagates the ChromaDB endpoint to the wrapper · **BLOCKING** · `class: tech`

`install_mcp_launcher` faithfully substitutes the operator's chroma host/port into the launcher (`common.sh:~775`, `MEMPALACE_CHROMA_PORT:-8001`), and the launcher uses them for its wait loop (`mcp-daemon-launcher.sh:42-43`, `:116`). But they are plain shell variables. The committed launcher exports exactly three things:

```
53:export LAUNCHER_SOURCE_SHA
85:export MEMPALACE_MCP_HTTP_TOKEN
90:export MEMPALACE_MCP_IDLE_HOURS
```

The wrapper reads the endpoint from the environment, defaulting to loopback:8001 (`mempalace-http-wrapper.py:222-223`):

```python
_host = os.environ.get("MEMPALACE_CHROMA_HOST", "127.0.0.1")
_port = int(os.environ.get("MEMPALACE_CHROMA_PORT", "8001"))
```

Neither launchd nor `systemctl --user` inherits the login shell, so on any machine with chroma on a non-default port: the launcher waits on the **right** port and succeeds, `exec`s the wrapper, which probes **8001**, fails loud, exits non-zero — and `KeepAlive` / `Restart=always` respawns it forever. That is the exact "restart us in a loop at boot" outcome the wait-loop comment at `:110-114` claims to have converted into "one long attempt". Fix: `export MEMPALACE_CHROMA_HOST MEMPALACE_CHROMA_PORT` before the `exec`.

### F15 — `MEMPALACE_PALACE_PATH` is never pinned into the launcher · `class: tech`

`mcp_token_path` derives the token key from the operator's shell `MEMPALACE_PALACE_PATH` (`common.sh:1196`); the launcher recomputes it inside an environment-less supervisor and gets the **default** key (`mcp-daemon-launcher.sh:61`). Host, port, chroma endpoint and interpreter are all substituted into the launcher; the palace path is not. An operator on a non-default palace therefore has the token written under `key(custom)` while the launcher looks under `key(default)`, dies "bearer token file not found", the health poll expires, and `switch-mempalace-http.sh:22-27` reports "no assistant has been switched". Fails closed and visibly — but the feature is simply unreachable for that operator. Same root cause as F8: two derivations of one path.

### F16 — a re-install rewrites the launcher without restarting the daemon, and that **silences** the drift alarm · `class: tech`

`install_mcp_daemon` calls `install_mcp_launcher` (re-materialising `~/.crewrig/mcp-daemon-launcher.sh` with a fresh `LAUNCHER_SOURCE_SHA`) and then `install_daemon_supervisor`, which sees the label already loaded and prints "launchd agent already loaded — skipping load" (`common.sh:602-604`). Nothing restarts the unit. The running daemon continues executing whatever the *previous* launcher did, while `status-mcp-server.sh:73-90` compares the just-rewritten recorded hash against the source, finds them equal, and prints "in sync with the repository".

So the drift detector reports green in exactly the situation it exists to catch: installed launcher ≠ what the running daemon was launched from. The check certifies a file, not the process. Either restart on re-install, or compare against something the running daemon actually carries.

### F17 — `launchctl list | grep -q "$label"` is a substring match · `class: tech`

`common.sh:602`, `common.sh:1143`, `stop-mcp-server.sh:23`. `com.mempalace.mcp-server-test-<pid>` — the label the suite exports at `test-mcp-daemon.sh:48` — matches a `grep` for `com.mempalace.mcp-server`. Dots are unescaped regex wildcards too. Latent rather than live, because the suite never actually loads a unit; but a leftover test agent from any future test that *does* load one would make a production install announce "already loaded — skipping load", never load, and then fail the 15-second health poll with a misleading message. Prefer `launchctl list "$label"` and test the exit status.

### F18 — `uninstall-mcp-daemon.sh` is not the inverse of what its header claims · `class: tech`

It removes the unit and the launcher (`:20-28`) but leaves all four CLI registrations pointing at `http://127.0.0.1:8021/mcp`. Its header (`:6-9`) describes it as "what a rollback runs, BEFORE reverting". Run it, revert the ticket, and every assistant now points at a daemon that no longer exists and whose launcher has been deleted — with no counterpart to `switch_assistants_to_http` to put them back on stdio, and no message naming `setup-*-interactive.sh` as the way back. This compounds F5: the forward path is not integrated into setup, and the reverse path does not exist at all.

### Minor

- `status-mcp-server.sh:58-61` — `curl -w '%{http_code}'` already emits `000` on connection failure, so the `|| echo "000"` yields `000\n000` inside the operator-facing message.
- `stop-mcp-server.sh:33` gates on `is-active`, so a unit that is **enabled-but-failed** reports "no supervisor unit active" and is left failed — the state an operator is most likely running a stop for.
- `stop-mcp-server.sh:26,36` say the supervisor "brings it straight back"; `plist:45-46` sets `ThrottleInterval 10` and the systemd unit `RestartSec=3`, so there is a 3–10 s window in which every CLI's memory calls fail. Worth stating in the message.

---

### Verified sound (answers to questions worth recording)

- **The token refusal is unbypassable.** The only path that starts the HTTP daemon is `install_mcp_daemon` → `install_daemon_supervisor` → the unit, whose `ExecStart` is the launcher. Every other invocation of `mempalace-http-wrapper.py` in the repo — setup scripts, config templates, `hooks/mempalace-transcript.sh`, tests — is stdio and needs no token. The one soft spot is `MEMPALACE_MCP_LAUNCHER_PATH` (`common.sh:724`), which lets an environment variable choose the program the unit names. Modulo F13, the guard holds.
- **No token is ever printed.** `status-mcp-server.sh` echoes neither token nor header; `doctor-mempalace.sh:263-266` deliberately reports "bearer header present (value not shown)"; the unit files carry no credential; MemPalace's HTTP access log records the request line only (`mcp_server.py:5255`).
- **The 401 probe is correct**, verified against installed MemPalace 3.6.0 rather than assumed: `_request_rejected` (`:5271-5296`) runs Host-pin → Origin → bearer before dispatch, `do_POST` passes `require_auth=True` (`:5312`) while `/healthz` uses `require_auth=False` (`:5302-5305`). It is a plain `BaseHTTPRequestHandler`, so the absent `Accept: text/event-stream` header does not perturb the probe, and curl's default `Host` is in `_http_allowed_host_values`.
- **Portability is clean on both platforms.** `check-bash32-portability.sh` passes; `bash -n` clean on all touched scripts; no `declare -A`, `${var^^}`, `mapfile`, `&>>`, `grep -P`, `timeout`, or GNU-only `date`/`sed -i`. The GNU-first-then-BSD `stat` probe is correct and its ordering is documented (`test-mcp-daemon.sh:87-89`); `sleep 0.3` (`common.sh:644`) is accepted by both BSD and GNU sleep.
- **The argv handoff works**: `mempalace.mcp_server` parses `--transport/--host/--port` at import via `parse_known_args` (`:295-301`), and `MEMPALACE_MCP_IDLE_HOURS=0` genuinely disables the idle watchdog (`:696-705`) — the open question ADR 0016 flagged for DEV resolves in the affirmative.
- **`uninstall_daemon_supervisor` uses the correct inverse verbs**, and `stop-mcp-server.sh` correctly refuses to use them.

---

### Revised summary

Blocking is now **F1–F5, F12, F13, F14** — eight. The three additions all sit in the security and lifecycle story the PR leads with: the credential reaches a `0644` file (F12), the fail-closed guard is bypassable by whitespace at this commit (F13), and the daemon respawn-loops on any non-default ChromaDB port (F14).

The pattern across F12/F13/F14/F16 is worth naming, because it is the same one as F1/F2: each is a place where the *stated* guarantee and the *tested* property are different quantities. The launcher's comment promises the token never touches a world-readable file, and the suite checks the unit file — while the registration writes `0644`. The guard promises fail-closed on an empty token, and the suite checks the exit status — which the ChromaDB deadline also supplies. The drift check promises the installed launcher matches the repository, and it compares a file it just rewrote. Tightening the assertions onto the properties actually claimed would have caught most of this diff's defects, and is probably worth more than any individual fix below.

Issue #709, opened 2026-08-04
🐛 Both CI parity checks pass blind to env-block divergence — GitLab `lint-specs` gets no `BASE_REF`


hcross:
## Problem

`scripts/build-ci.sh --check` and `scripts/check-ci-parity.sh` both report the GitHub Actions and GitLab pipelines in agreement while their **environment blocks differ**. Neither check inspects env vars, so a divergence that changes what a job actually does passes green.

Surfaced as an `arch` finding during the cold review of PR #708, which introduces the first case where it bites.

## The concrete instance

PR #708 wires a `BASE_REF` environment variable into the GitHub `lint-specs` job, conditional on pull-request context, so the spec linter compares against the pull request's real target branch. The GitLab job gets no equivalent, because **`ci/ci-capabilities.yml` has no way to express an env var** — its `requires:` vocabulary is closed to `{runtime, tools, history-depth}`.

Consequence: a merge request targeting `release/**` on GitLab compares against `origin/main` rather than against its actual target. The linter still runs and still fails closed, so nothing is silently green — but it answers a question about the wrong branch.

Both parity checks pass on that diff:

```
build-ci.sh --check      -> OK: .gitlab-ci.yml matches the CI capability reference.
check-ci-parity.sh       -> OK: reference, GitHub Actions, GitLab agree on the portable
                            capability set; every pipeline job is traceable.
```

## Why the checks miss it

They compare the *command lists* of each job and the capability reference, in order. An env block is not a command, so it is outside what either check models. The reference format cannot express one, so there is nothing for the checks to compare even in principle.

That makes this two defects stacked, and they need fixing in order:

1. **`ci/ci-capabilities.yml` cannot express a job-scoped environment variable.** Until it can, GitLab cannot receive `BASE_REF` and the divergence cannot be removed.
2. **Neither parity check verifies env blocks**, so any future divergence of this kind will also pass green. Fixing (1) without (2) leaves the next one undetected.

## Why it is worth fixing rather than accepting

The multi-engine CI parity contract exists so that a capability verified on one engine is verified on all of them. A parity check that passes while the engines behave differently is the same defect class as a green test that does not test what it claims — the pattern that cost issue #623 three review iterations. Here the check reports agreement that does not hold.

## Not urgent, and honestly scoped

[...]

PR #664, opened 2026-07-24
📝 Fold draft→approved status transition into the spec-PR commit

hcross:
This PR folds the initial `draft` → `approved` spec status transition into the spec-PR's own commit, so a merged spec lands on `main` already carrying `status: approved` instead of requiring a second, metadata-only follow-up PR. It rewrites the *Recording a status transition* section of `docs/spec-format.md` into two documented cases and updates the *Lifecycle states* table's `approved` row to match.

## How to read this PR?

A single file changed — `docs/spec-format.md` (one file, `54 insertions(+)`, `12 deletions(-)` per `git show 45506fb --stat`). Read the two hunks in order:

1. **The *Lifecycle states* table, `approved` row** (first hunk) — the one-line contract change. The "Accompanying artifact" cell now states that the merged spec-PR's own (squash) commit already carries `status: approved`, so no separate status-transition PR is opened. The Trigger and Authority cells are unchanged.
2. **The *Recording a status transition* section rewrite** (second hunk) — the section is split from one case into two: (a) the initial `draft` → `approved` transition, recorded inside the spec-PR's own commit; (b) every later transition (`approved` → `implemented`/`superseded`/`archived`), which stays a post-merge metadata-only edit.

Non-obvious decisions to note while reading:

- **Mode-invariance.** The `draft` → `approved` rule is stated to apply in all four interaction modes (`FULL`/`INTERMEDIATE`/`MINIMAL`/`AUTO`) because the merge-authorization gate fires in every mode — no mode lacks a real approval event to record.
- **Ordering guard.** The transition SHALL be recorded *only after* the merge-authorization approval is granted; recording `approved` ahead of that gate is defined as a violation.
- **New commit, not amend.** The documented merge mechanic creates a new commit rather than `git commit --amend`: under squash-merge both routes collapse to the same single commit on `main`, but a new commit needs only a plain `git push`, whereas an amend would force `git push --force-with-lease` and risk clobbering a concurrent push on the shared spec-PR branch.

One fitting irony, worth a sentence: this ticket (#610) is itself an instance of the overhead it eliminates. Specs 0099, 0100, and 0101 — all processed in this same session — still show the old two-PR-shaped world: each merged spec-PR left its spec at `status: draft` with no follow-up transition PR ever opened in practice (the separate-PR step was already being informally skipped), and this PR is what makes that skip *correct by rule* going forward instead of a silent deviation.

## How to test this PR?

[...]

PR #764, opened 2026-08-08
🐛 Charge a base-branch draft spec to the change that can fix it

### LINKED ISSUE #732: 🚧 A spec merged in draft blocks lint-specs on every PR in the repository
> **Filed directly rather than tagged in MemPalace.** The `harness-report`
> fallback for an unreachable memory service applies (`SKILL.md` → *Fallback*,
> case (b)): the MemPalace MCP server was disconnected at tagging time, so there
> was no live server to retry against. Payload reproduced verbatim below for the
> Curator.

## What happened

`specs/0112-spec-id-reservation.md` reached `main` via PR #727 carrying
`status: draft`. The spec-0109 invariant forbids exactly that, and
`scripts/lib/spec-linter.js:479` enforces it **against the base branch** — so
from that moment, `lint-specs` failed on *every* pull request in the repository,
regardless of whether the PR touched `specs/` at all.

**Blocked window: 10 h 28.**

| Event | Time (UTC) |
|---|---|
| PR #727 merges spec 0112 with `status: draft` | 2026-08-07 22:23:27 |
| PR #729 (an ADR — touches only `docs/`) merged with `lint-specs` red | 2026-08-08 07:50:24 |
| PR #731 restores the invariant (`draft` → `approved`) | 2026-08-08 08:51:38 |

PR #729 changed two files, `docs/adr/0016-*.md` and `docs/index.json`. It could
not go green for a reason it had no way to influence, in a file it did not
touch, owned by a different in-flight ticket (#726). It was merged with a red
check on the owner's explicit go-ahead — a workaround that had to be explained
and authorised, which is the safeguard-friction recognition signal.

## Why this is a defect and not bad luck

The guard is correct about the invariant and wrong about **who pays for
violating it**. It converts a local fault into a global halt: one ticket's
in-flight artifact stops every other agent in the repository, including agents
working on unrelated subsystems who cannot fix the cause and must either wait or
merge red.

Worse, the window is **structural, not accidental**. The spec-PR workflow merges
the spec artifact first and transitions its status in a *separate, later* PR.
Between those two merges the invariant is violated by construction. No amount of
author discipline closes that gap — #727 and #731 were the same well-behaved
ticket doing exactly what the workflow prescribes, ten hours apart. Any spec-PR
opens the same window; this one was simply long enough to be noticed.

Multi-agent operation makes it worse, not better: the more parallel sessions,
the higher the chance that someone else's PR lands inside the window.

## Suggested fixes

Ordered by how directly each addresses the cause.

1. **Scope the guard to the PR that owns the fault.** A PR that does not add,
   modify, or own the offending spec should not fail on its status. Base-branch
   violations could be attributed to the PR that introduced them, or reported as
   an annotation on unrelated PRs rather than a failing check. A guard that
   cannot be satisfied by the PR it fails is misattributing blame — this alone
   removes the cross-ticket coupling even if the window stays open.
2. **Close the window at its source.** Make the status transition atomic with
   the spec-PR merge — either the spec-PR merges already `approved` (the status
   its own merge is the documented trigger for), or the merge performs the
   transition. Then the invariant is never violated and the guard never fires
   spuriously. This is the deeper fix; it also removes an entire follow-up PR
   per spec.
3. **Fail only the owner, warn everyone else.** If the window must stay open,
   the check can stay red for the owning PR and degrade to a non-blocking
   annotation elsewhere, so the signal survives without the halt.

Fix 1 and fix 2 are complementary: 2 removes the occurrence, 1 removes the blast
radius when something else re-opens a similar window later.

## Friction payload

```yaml
FRICTION: A spec merged in draft blocks lint-specs on every PR in the repository

writer_agent: claude-code
subcategory: spec-status-invariant-global-block
room: process
canonical: https://github.com/crewrig/crewrig
severity: high
evidence:
  - 36c543d:specs/0112-spec-id-reservation.md:4   # status: draft reaches main
  - scripts/lib/spec-linter.js:479                # base-branch check that halts every PR
  - specs/0109-spec-status-invariant-on-main.md   # the invariant being enforced
  - https://github.com/crewrig/crewrig/pull/729   # unrelated PR merged red
  - https://github.com/crewrig/crewrig/pull/731   # restores the invariant, 10h28 later
suggestion: >
  Scope the base-branch status check to the PR that owns the offending spec —
  a PR that cannot fix a violation should not be failed by it. Separately, make
  the draft→approved transition atomic with the spec-PR merge so the window
  never opens: the spec-PR merge is already the documented trigger for
  `approved`, yet the workflow defers it to a second PR, leaving the invariant
  violated by construction in between.
```

hcross:
A non-delta spec carrying `status: draft` on the base branch failed **every** open pull request in the repository, including ones that touched no spec and had no way to go green. This charges the failure to the change that can actually record the correct status, and keeps it unconditional on the base branch's own build, where the invariant belongs.

Implements `specs/0109-spec-status-invariant-on-main.delta-02.md` (R2 as replaced, R9–R13). Closes #732.

## How to read this PR?

Three files. Read them in this order — the second is where the behaviour is, the first is why it is shaped that way.

1. **`specs/0109-spec-status-invariant-on-main.delta-02.md`** — already on `main` (PR #758), not in this diff, but read its preamble first if you have not. Two of the three fixes the originating friction report proposed were **already in force** and are deliberately not rebuilt here: the atomic `draft` → `approved` transition (PR #664, two weeks before the incident) and the detection itself, which failed the `main` build four minutes after the offending merge. What was missing is *attribution*.
2. **`scripts/lib/spec-linter.js`** — one new helper (`changedPaths`) and one rewritten block. Identification of an offender is untouched; only the consequence is partitioned. The comment above `changedPaths` carries the two decisions worth arguing with: no second environment variable, and no two-dot-diff fallback.
3. **`scripts/tests/test-spec-linter.sh`** — cases 35–38, on two new git fixtures. Cases 35–37 share `$GITFIX2` and **their order is load-bearing**: each advances that fixture's work-tree state, from clean, to an unrelated edit, to an edit of the offender itself. Reordering them silently changes what each measures, which is why the comment says so.

Three non-obvious points:

- **The asymmetry is the safety property.** `changed === null` (attribution unavailable) and `changed.size === 0` (nothing to attribute to) both make **every** offender blocking. Only a non-empty set can exempt one. So no state of the repository — shallow clone, unrelated histories, detached checkout — turns the check green while the invariant is violated.
- **`size === 0` is how the base branch keeps failing.** On a `push` to `main`, `HEAD` *is* the derived base, so the merge base is `HEAD` and the diff against a clean tree is empty. Requirement 10 is not a special case bolted on for CI; it falls out of the same measurement.
- **The diff has no second revision argument on purpose.** It compares the merge base to the *work tree*, so committed, staged and unstaged edits all count — matching what the linter reads. A status corrected but not yet committed would otherwise be attributed to nobody.

## How to test this PR?

```sh
# The suite needs node_modules resolvable from the worktree root it computes.
# A fresh worktree has none, and 25 of 39 cases then fail on
# `npx markdownlint: could not determine executable to run`.
ln -s /path/to/main/checkout/node_modules node_modules   # or: npm install

bash scripts/tests/test-spec-linter.sh        # expect: 39 passed, 0 failed
BASE_REF=crewrig/main node scripts/lib/spec-linter.js    # whole corpus
npx markdownlint docs/spec-format.md -c .markdownlintrc
```

**The live negative control, which is the test that matters.** At the time of writing, `specs/0114-shared-worktree-agent-isolation.md` is on `main` carrying `status: draft` (landed by `a0291d2`, PR #759) — a second, independent occurrence of this ticket's incident, 24 hours after the first. So the fix can be exercised against production data rather than a fixture:

```sh
# This branch's linter — same corpus, same base:
BASE_REF=crewrig/main node scripts/lib/spec-linter.js
#   [WARN] … specs/0114-shared-worktree-agent-isolation.md … does NOT fail this run
#   Linting passed!

# main's linter — same corpus, same base:
git show crewrig/main:scripts/lib/spec-linter.js > scripts/lib/.control.js
BASE_REF=crewrig/main node scripts/lib/.control.js specs
#   [FAIL] … specs/0114-shared-worktree-agent-isolation.md
#   Linting failed: 1 files contain errors.
rm scripts/lib/.control.js
```

Run the control from **inside** the worktree, not `/tmp` — `js-yaml` resolves relative to the script's own path, and a copy in `/tmp` dies on `MODULE_NOT_FOUND` before it reaches the check.

**Falsification.** Each new case was verified to fail when the behaviour it covers is removed, rather than asserted to:

| Mutation | Requirement removed | Case that goes red |
|---|---|---|
| `blockEveryOffender = true` | R9 — the bystander partition | 36 |
| `blockEveryOffender = changed === null` | R10 — an empty set stops blocking | 35 (and 28, 29 with it) |
| `changed \|\| new Set(['__none__'])`, `[ERROR]` suppressed | R11 — unavailable attribution exempts | 38 |

**Known limit, stated rather than glossed:** Case 37 does *not* fail if the implementation reverts to unconditional blocking — under that behaviour it still passes. It falsifies "never fail"; Case 36 falsifies "always fail". Neither alone pins the replaced R2's discriminator; the pair does.

### `scripts/lib/spec-linter.js`

**Added `changedPaths(baseRef)`**, beside the existing `baseBranchPaths()`. Runs `git merge-base <baseRef> HEAD`, then `git diff --name-only -z <merge-base>` with no second revision, and returns a `Set` of repo-relative paths — or `null` if either command fails or the merge base comes back empty.

Two rejected alternatives are recorded at the call site because both are reasonable and both are wrong here:

- *A second environment variable* (`SPEC_STATUS_SCOPE=strict|attributed`) wired per CI event. Rejected: it would have to be provisioned identically in two engines that already drift on exactly this pair of jobs (#709), and it would restate a fact the linter can measure.
- *A two-dot `git diff <baseRef>` fallback when `merge-base` fails.* Rejected: against a base that has advanced it reports base-only changes as the change under test's — mis-attribution, which is worse than an honest `null` that R11 handles by blocking.

**`resolveBaseContext()`** now returns `changed: changedPaths(ref)` alongside `basePaths` and `relPathByFile`, so the offender block reads one context object instead of re-probing git. It is called only on the enforced path, so the not-in-a-git-work-tree skip (Case 33) never touches git.

**The offender block** keeps its identification filter verbatim and partitions the result: `blocking` when `changed` contains the path, `bystanders` otherwise, with `blockEveryOffender = changed === null || changed.size === 0` collapsing both degenerate states onto "everything blocks". `[FAIL]` keeps its exact previous wording and still increments `totalErrors`; `[WARN]` is new, names the files and the base ref, states the violation is on the base branch and not this change's to fix, points at where it *is* fixed, and does not touch `totalErrors`. On the `null` path an `[ERROR]` line naming R11 precedes the `[FAIL]` block so the failure is explained rather than mysterious.

### `scripts/tests/test-spec-linter.sh`

**`$GITFIX2`** — one non-delta spec at `draft` plus one unrelated tracked file, committed on `main`, no remote (these cases pass `BASE_REF` explicitly). Cases 35 → 37 walk its work tree: clean (R10, exit 1), unrelated edit (R9, exit 0 with `[WARN]` and asserting no `[FAIL]`), edit of the offender (replaced R2, exit 1). Cases 36 and 37 both assert on *presence and absence* of a marker, because an exit code alone cannot distinguish "warned correctly" from "stopped checking".

**`$GITFIX3`** — an orphan branch as the base ref, the cheapest reproduction of "resolves but shares no ancestor", so `git merge-base` genuinely fails (R11, exit 1, stderr naming the cause). The orphan commit stages `specs` explicitly and **not** `git add -A`: adding everything tracks the harness's own untracked scaffolding, and the `checkout main` that follows then *deletes* `.markdownlintrc`, so markdownlint dies on a missing config and the case fails for a reason that has nothing to do with the check. The comment in the file says this, because the symptom is thoroughly misleading.

### `docs/spec-format.md`

Requirement 12. In *Lifecycle states → No spec on `main` is a draft*, the enforcement paragraph is reworded from "fails and names" to "names" (naming and failing are now separate decisions), and a new paragraph states which build fails, which only warns, and — the point of the requirement — that **a `[WARN]` met on a pull request is a statement about `main`, not about that pull request**, cleared by editing the named spec on the base branch and by nothing in the change under test.

### Not triggered, checked rather than assumed

- **CLI matrix:** none of the three paths is a trigger surface under `AGENTS.md` → *CLI Matrix Maintenance*. No `docs/cli-matrix.md` edit is due.
- **CI wiring:** unchanged in both engines. Attribution reads the `BASE_REF` the check already reads; no new variable, no new parity surface.
- **Version bump:** `metadata.provenance.version` lives on skill and agent sources; none is touched. The spec-side bump (`2.0.0`) shipped in #758.
- **Build outputs / core-paths manifest:** nothing under `artifacts/`, no core-layer path added, removed or reclassified.

### DIFF
--- a/docs/spec-format.md
+++ b/docs/spec-format.md
@@ -211,15 +211,32 @@ to use `status` for the one question it exists to answer — "merged and in
 force" or "proposed but never landed".
 
 The rule is enforced mechanically rather than by recall. The spec linter
-(`scripts/lib/spec-linter.js`, run as `task spec:lint` in CI) fails and names
-every non-delta spec that is **already present on the change's base branch**
-while carrying `status: draft`. Presence on the base branch is the
-discriminator, so a spec the change under test *introduces* is not flagged: it
-is legitimately `draft` until the *Merge mechanic* below flips it. Delta-specs
-are exempt too — what `status` a delta should carry is deliberately unsettled,
-per
+(`scripts/lib/spec-linter.js`, run as `task spec:lint` in CI) names every
+non-delta spec that is **already present on the change's base branch** while
+carrying `status: draft`. Presence on the base branch is the discriminator for
+being named, so a spec the change under test *introduces* is not named at all:
+it is legitimately `draft` until the *Merge mechanic* below flips it.
+Delta-specs are exempt too — what `status` a delta should carry is deliberately
+unsettled, per
 [`specs/0109-spec-status-invariant-on-main.md`](../specs/0109-spec-status-invariant-on-main.md)
 → *Out of scope*.
+
+**Which build the violation fails, and which it only warns.** Being named and
+being failed are two different things, decided by two different questions. A
+change that **modifies** a named spec is failed by it (`[FAIL]`), because it can
+record the correct status in the same edit. A change that does not — a pull
+request touching no spec at all, say — gets the same spec named as a
+**non-blocking** warning (`[WARN]`) and passes: a check a change has no power to
+satisfy must not block it. The build that does fail in that case is the base
+branch's own, where `HEAD` is the base and there is therefore no change to
+charge the violation to. So **a `[WARN]` met on a pull request is a statement
+about `main`, not about that pull request**: it is cleared by correcting the
+named spec's status on the base branch, and nothing in the change under test
+will make it go away. When the linter cannot determine which files the change
+modifies, it reports every named spec as blocking and says so — an
+indeterminate answer never becomes an exemption.
+[`specs/0109-spec-status-invariant-on-main.delta-02.md`](../specs/0109-spec-status-invariant-on-main.delta-02.md)
+is the contract.
 
 ### Recording a status transition
 
--- a/scripts/lib/spec-linter.js
+++ b/scripts/lib/spec-linter.js
@@ -163,6 +163,47 @@ function baseBranchPaths(baseRef, relPaths) {
     return new Set(result.stdout.split('\0').filter(Boolean));
 }
 
+// changedPaths(baseRef) — the repo-relative paths the change under test
+// modifies, as a Set, or `null` when that set cannot be derived. This is the
+// discriminator spec 0109 delta-02 adds on top of presence-on-base: a
+// base-branch violation fails the change that CAN cure it (R2 as replaced) and
+// is reported non-blocking to every change that cannot (R9).
+//
+// Derived from the repository at check time, never configured (R11, mirroring
+// what R3 already obliges for the set of files examined). Deliberately NOT a
+// second environment variable alongside BASE_REF: the value is measurable here,
+// and a second variable would have to be provisioned identically in two CI
+// engines that already drift on exactly this pair of jobs (issue #709).
+//
+// `git merge-base` then `git diff` against it, rather than a two-dot diff
+// against the base tip: against a base that has advanced, the two-dot form
+// reports base-only changes as this change's, which is MIS-attribution rather
+// than degraded attribution. There is no fallback to it for that reason — a
+// `null` here is an honest "unknown", which R11 answers by blocking.
+//
+// The diff compares the merge base to the WORK TREE (no second revision
+// argument), so committed, staged and unstaged edits all count. That matches
+// what the linter reads: it lints the work tree, so attribution must measure the
+// work tree too, or a status corrected but not yet committed would be attributed
+// to nobody.
+//
+// Empty is meaningful and distinct from `null`: an empty set means the tree
+// under test introduces nothing relative to the base — which is the state this
+// runs in on the base branch's own build, where `HEAD` IS the derived base — and
+// R10 answers it by blocking on every offender.
+function changedPaths(baseRef) {
+    const mergeBase = gitCapture(['merge-base', baseRef, 'HEAD']);
+    const mergeBaseSha = mergeBase.stdout.trim();
+    if (mergeBase.status !== 0 || mergeBaseSha === '') {
+        return null;
+    }
+    const diff = gitCapture(['diff', '--name-only', '-z', mergeBaseSha]);
+    if (diff.status !== 0) {
+        return null;
+    }
+    return new Set(diff.stdout.split('\0').filter(Boolean));
+}
+
 // resolveBaseContext(files) — decide whether the base-branch status check runs,
 // and if so against which paths. Three outcomes, deliberately distinct:
 //
@@ -244,7 +285,7 @@ function resolveBaseContext(files) {
         process.exit(2);
     }
 
-    return { enforced: true, ref, basePaths, relPathByFile };
+    return { enforced: true, ref, basePaths, relPathByFile, changed: changedPaths(ref) };
 }
 
 function lintFile(filePath) {
@@ -479,7 +520,8 @@ function run() {
         }
     }
 
-    // Base-branch `status: draft` check (spec 0109 R1/R2). A non-delta spec
+    // Base-branch `status: draft` check (spec 0109 R1/R2, R2 as replaced by
+    // delta-02, plus delta-02 R9/R10/R11). A non-delta spec
     // already present on the base branch has by definition had its own spec-PR
     // merged, which is the trigger docs/spec-format.md assigns to `approved` —
     // so `draft` on it is a contradiction, not a lagging value. Delta-specs are
@@ -492,14 +534,49 @@ function run() {
             !isDelta && status === 'draft' && baseContext.basePaths.has(baseContext.relPathByFile.get(file))
         );
         if (offenders.length > 0) {
-            console.error(`\n[FAIL] Non-delta specs present on the base branch (${baseContext.ref}) carry 'status: draft':`);
-            for (const { file } of offenders) {
-                console.error(`  - ${file}`);
+            // Whose violation is it? (delta-02 R2/R9/R10/R11.) Identification
+            // above is unchanged; only the consequence is attributed. `null`
+            // means attribution is unavailable (R11) and an empty set means
+            // there is no change to attribute to (R10) — both make every
+            // offender blocking, so only a NON-EMPTY set can exempt one. That
+            // asymmetry is the whole safety property: no state of the
+            // repository turns the check green while the invariant is violated.
+            const changed = baseContext.changed;
+            const blockEveryOffender = changed === null || changed.size === 0;
+            const blocking = blockEveryOffender
+                ? offenders
+                : offenders.filter(({ file }) => changed.has(baseContext.relPathByFile.get(file)));
+            const bystanders = blockEveryOffender
+                ? []
+                : offenders.filter(({ file }) => !changed.has(baseContext.relPathByFile.get(file)));
+
+            if (changed === null) {
+                console.error(`\n[ERROR] Base-branch status check (spec 0109 R11): the set of files this`);
+                console.error(`        change modifies could not be derived from the repository, so every`);
+                console.error(`        spec below is reported as this change's to fix rather than exempt.`);
             }
-            console.error(`  A spec reaches the base branch only by its own merged spec-PR, which is`);
-            console.error(`  the trigger docs/spec-format.md assigns to 'approved'. Record the status`);
-            console.error(`  that reflects the spec's true state (metadata-only edit).`);
-            totalErrors++;
+
+            if (blocking.length > 0) {
+                console.error(`\n[FAIL] Non-delta specs present on the base branch (${baseContext.ref}) carry 'status: draft':`);
+                for (const { file } of blocking) {
+                    console.error(`  - ${file}`);
+                }
+                console.error(`  A spec reaches the base branch only by its own merged spec-PR, which is`);
+                console.error(`  the trigger docs/spec-format.md assigns to 'approved'. Record the status`);
+                console.error(`  that reflects the spec's true state (metadata-only edit).`);
+                totalErrors++;
+            }
+
+            if (bystanders.length > 0) {
+                console.error(`\n[WARN] Non-delta specs present on the base branch (${baseContext.ref}) carry 'status: draft':`);
+                for (const { file } of bystanders) {
+                    console.error(`  - ${file}`);
+                }
+                console.error(`  This change modifies none of them, so the violation lives on the base`);
+                console.error(`  branch rather than in this change and is not this change's to fix — it`);
+                console.error(`  does NOT fail this run. Correcting it is an edit to the named spec on the`);
+                console.error(`  base branch, whose own build fails on it (spec 0109 R1/R9/R10).`);
+            }
         }
     }
 
--- a/scripts/tests/test-spec-linter.sh
+++ b/scripts/tests/test-spec-linter.sh
@@ -680,10 +680,12 @@ run_base_case "Case 34 — an uncanonicalizable linted path is named, not a stac
 # same way — `git symbolic-ref` rather than `git init -b`, so the fixture does
 # not depend on the host's init.defaultBranch.
 #
-# Case ORDER is load-bearing here: the three cases share one fixture and each
-# advances its work-tree state, from clean (35) to an unrelated edit (36) to an
-# edit of the offender itself (37). Reordering them silently changes what each
-# measures.
+# The three cases share this one fixture but do NOT depend on each other's
+# order: each one RESETS the work tree (`git checkout -- .`) and then applies
+# only the state it needs. An accumulating variant would pass just as green while
+# measuring something other than what the case names claim the moment anyone
+# inserts or reorders a case — an order-dependent test that silently changes
+# meaning is the same class of defect as the check being fixed here.
 # -------------------------------------------------------------------------
 GITFIX2="$TMP_ROOT/gitfix2"
 mkdir -p "$GITFIX2/specs" "$GITFIX2/docs"
@@ -711,6 +713,7 @@ printf '# Unrelated\n\nBody.\n' > "$GITFIX2/docs/unrelated.md"
 # empty-set branch and the base branch's build goes green while `main` violates
 # the invariant.
 # -------------------------------------------------------------------------
+git -C "$GITFIX2" checkout -q -- .
 run_base_case "Case 35 — no change relative to the base ref blocks on every offender (R10)" \
   "$GITFIX2" "specs" 1 "main" \
   "specs/0210-attributed.md" "[WARN]"
@@ -723,6 +726,7 @@ run_base_case "Case 35 — no change relative to the base ref blocks on every of
 # is present AND no `[FAIL]` block is — because exit 0 alone could not
 # distinguish "warned correctly" from "stopped checking".
 # -------------------------------------------------------------------------
+git -C "$GITFIX2" checkout -q -- .
 printf '# Unrelated\n\nBody, edited by the change under test.\n' > "$GITFIX2/docs/unrelated.md"
 run_base_case "Case 36 — an offender the change does not touch warns, and does not fail (R9)" \
   "$GITFIX2" "specs" 0 "main" \
@@ -735,6 +739,7 @@ run_base_case "Case 36 — an offender the change does not touch warns, and does
 # pins the replaced R2's discriminator: same offender, same base branch, and the
 # outcome turns only on whether the change touches the file.
 # -------------------------------------------------------------------------
+git -C "$GITFIX2" checkout -q -- .
 printf '\nEdited by the change under test.\n' >> "$GITFIX2/specs/0210-attributed.md"
 run_base_case "Case 37 — an offender the change does touch fails (R2 as replaced)" \
   "$GITFIX2" "specs" 1 "main" \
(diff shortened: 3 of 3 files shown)


hcross:
## Verdict: REQUEST CHANGES

One `tech` finding. Nothing about the mechanism — the attribution logic is
correct and well evidenced. The finding is in the test suite's fixture handling.

**Reviewer-independence disclosure, stated first because it qualifies everything
below.** This review was NOT cold-spawned. `AGENTS.md` → *Agent Team Protocol*
calls for a `pr-reviewer` spawned without authoring context; this session runs
under an instruction not to spawn subagents, so the same session that authored
the diff reviewed it. The shared-identity posting path was taken as well
(`gh api user` → `hcross`, PR author → `hcross`, so a plain comment rather than a
review event, per the skill's rung 2 and issue #692). Treat the verdict as a
structured self-audit against the skill's checklist, not as independent
confirmation. A genuine cold pass would still be worth having.

### CI status

All **13** required checks **pass** on `3e66fe4` (run `31281226716`). No failing
check, none pending.

The `lint-specs` job is worth quoting, because it is the change demonstrating
itself on production data rather than on a fixture:

```text
Results: 39 passed, 0 failed
[WARN] Non-delta specs present on the base branch (origin/main) carry 'status: draft':
  - specs/0114-shared-worktree-agent-isolation.md
Linting passed!
```

`specs/0114-shared-worktree-agent-isolation.md` is on `main` at `status: draft`
as of this review (landed by `a0291d2`, PR #759 — an independent second
occurrence of the incident this ticket exists for). Under `main`'s linter that
same corpus and base produce `[FAIL] … Linting failed: 1 files contain errors.`
This PR is green under the exact condition that is currently failing every other
pull request in the repository.

### Correctness

- `scripts/lib/spec-linter.js:544-545` — the two degenerate states collapse
  correctly: `changed === null` (underivable) and `changed.size === 0` (nothing to
  attribute to) both set `blockEveryOffender`, so only a non-empty set can exempt
  an offender. The asymmetry is the safety property, and it is the right way
  round — no repository state turns the check green while the invariant is
  violated.
- `scripts/lib/spec-linter.js:196-205` (`changedPaths`) — `git diff` is invoked
  with no second revision, so it compares the merge base to the **work tree**.
  That matches what the linter reads (it lints the work tree), which is what makes
  Case 30's pre-existing behaviour — a status corrected in the tree clears the
  violation on the same run — survive the change. Verified: Case 30 still passes.
- `scripts/lib/spec-linter.js:189-193` — no two-dot fallback when `git merge-base`
  fails. Correct: against an advanced base, `git diff <baseTip>` reports
  base-only changes as the change under test's, which would blame files the change
  never touched. `null` → block is the honest answer.
- Path-space agreement checked rather than assumed. `relPathByFile` values are
  `path.relative(repoRoot, …)`; `git diff --name-only` output is repo-root-relative,
  so `changed.has(relPathByFile.get(file))` compares like with like. Measured on
  this branch, including the adversarial config: with `diff.relative=true` and cwd
  at the repository root, output stays root-relative
  (`docs/spec-format.md`, `scripts/lib/spec-linter.js`, `scripts/tests/test-spec-linter.sh`).
  Note for the record — with cwd at a *subdirectory* and `diff.relative=true` the
  output becomes cwd-relative, and `git ls-tree` (the pre-existing
  `baseBranchPaths`, `spec-linter.js:155-164`) returns **empty** from a
  subdirectory even with default config. Neither is reachable: `spec-linter.js:473`
  passes `-c .markdownlintrc`, which is cwd-relative, so any run with cwd ≠ repo
  root dies at markdownlint before the base-branch check executes. **Not a defect,
  and deliberately not fixed here** — the check's path correctness rests on
  cwd == repo root, which today is guaranteed only as a side effect of
  markdownlint's config lookup. That accidental guarantee is worth a follow-up
  note; widening this diff for an unreachable hazard is not.
- Exit-status semantics preserved: the `[WARN]` branch never touches
  `totalErrors`, so a bystander-only run reaches `Linting passed!` and exit 0.
  Verified by Case 36 (`expected exit 0`) and by the CI log above.
- Rename and deletion paths reasoned through: a renamed or deleted base-branch
  draft is absent from the work tree, so it is never linted and never identified
  as an offender. No new behaviour needed.

### Convention compliance

- **Gitmoji** — commit `3e66fe4` and the PR title both use `🐛`, no
  conventional-commit prefix. Compliant.
- **Spec-PR ordering** — `specs/0109-…delta-02.md` merged first (PR #758,
  `863079f`); the implementation branch `fix/0109-spec-status-invariant-on-main`
  was cut afterwards from that commit. Compliant.
- **Branch naming** — `fix/<NNNN>-<slug>` with `0109` matching the spec id and
  `spec-status-invariant-on-main` matching its slug. Compliant.
- **Closing keyword** — the PR body carries `Closes #732` and the commit repeats
  it. `#732` is this ticket's shared logbook anchor and, per
  `docs/spec-pr-workflow.md` → *Independence rule*, the **implementation** PR is
  the one that carries it. Correct placement; the spec-PR correctly carried none
  (`gh pr view 758 --json closingIssuesReferences` → `[]`).
- **Version bump** — not triggered. `metadata.provenance.version` lives on skill
  and agent sources; this diff touches none. `check-skill-versions` passes.
- **CLI matrix** — none of the three changed paths is a trigger surface under
  `AGENTS.md` → *CLI Matrix Maintenance*. `check-ci-parity` and
  `gitlab-ci-check` both pass, consistent with no CI wiring change.
- **Pre-edit guard** — issue #732 pre-existed, work happened on a feature branch
  in the dedicated worktree `.worktrees/732`. Compliant.
- **PR body** — all four required sections present per `docs/pr-format.md`.

### Test coverage

Four cases added, `scripts/tests/test-spec-linter.sh:672-786`. Suite: **39
passed, 0 failed**.

Each was falsified by mutation rather than asserted to be falsifiable — the
evidence is in the logbook comment on #732:

| Requirement removed | Case that goes red |
|---|---|
| R9 — bystander partition | 36 |
| R10 — empty set stops blocking | 35, and 28/29 with it |
| R11 — underivable attribution exempts | 38 |

Two coverage observations, neither blocking:

- Case 37 does not fail under a revert to unconditional blocking. The PR body
  states this limit explicitly rather than implying the case is stronger than it
  is; Case 36 is what falsifies "always fail". The pair brackets the
  discriminator. Correctly disclosed.
- The pre-existing Cases 28/29 turn out to also cover R10's path (they run a
  clean tree where `HEAD` is the base). That is coverage overlap, not redundancy
  to remove — 28/29 assert on identification, 35 on attribution.

### Linter findings

`lint-markdown.sh docs/spec-format.md` — exit **0**, no output.

`lint-shell.sh scripts/tests/test-spec-linter.sh` — exit **1**, three SC2086
(info) findings, verbatim:

```text
In scripts/tests/test-spec-linter.sh line 67:
  output=$( ( cd "$TMP_ROOT" && CREWRIG_REPO_DIR="$env_repo_dir" node "$LINTER_JS" $files 2>&1 ) ) || actual_exit=$?
                                                                                   ^----^ SC2086 (info): Double quote to prevent globbing and word splitting.

In scripts/tests/test-spec-linter.sh line 544:
    output=$( ( cd "$workdir" && env -u BASE_REF node "$LINTER_JS" $targets 2>&1 ) ) || actual_exit=$?
                                                                   ^------^ SC2086 (info): Double quote to prevent globbing and word splitting.

In scripts/tests/test-spec-linter.sh line 546:
    output=$( ( cd "$workdir" && BASE_REF="$base_ref" node "$LINTER_JS" $targets 2>&1 ) ) || actual_exit=$?
                                                                        ^------^ SC2086 (info): Double quote to prevent globbing and word splitting.
```

**All three are pre-existing and outside this diff.** The diff's only hunk in that
file is `@@ -669,6 +669,117 @@`; lines 67, 544 and 546 sit in the pre-existing
`run_case` / `run_base_case` helpers, where the unquoted expansion is deliberate —
those helpers accept several targets as one string and rely on word splitting. No
new shellcheck finding is introduced. Not a finding against this PR.

### Findings

**Finding 1**

class: tech

`scripts/tests/test-spec-linter.sh:672-746` — Cases 35, 36 and 37 share
`$GITFIX2` and depend on execution **order**, because each *accumulates* work-tree
state on top of the previous one: 35 needs a clean tree, 36 adds an edit to
`docs/unrelated.md`, 37 appends to `specs/0210-attributed.md` on top of 36's edit.
The header comment states this ("Case ORDER is load-bearing here"), which is
honest but is documentation standing in for a guarantee. A future contributor
inserting a case between 35 and 36, or reordering them while tidying, gets three
cases that still *pass* while measuring something other than what their names
claim — and an order-dependent test that silently changes meaning is the same
class of defect as the check this PR is fixing. Remediation is cheap and removes
the coupling rather than documenting it: have each case **set** the exact
work-tree state it needs (`git -C "$GITFIX2" checkout -- .` first, then apply only
its own mutation) instead of inheriting the previous case's. One line per case;
the comment then describes intent rather than warning about a hazard.

### Considered and dismissed, so the next reader need not re-derive them

- **`changedPaths()` is computed eagerly** in `resolveBaseContext`
  (`spec-linter.js:296`), spawning two `git` subprocesses on every run even when
  no offender exists — the overwhelmingly common case. Dismissed: it is two
  subprocesses beside the three `resolveBaseContext` already spawns, the value
  belongs to the context object the block reads, and making it lazy would trade a
  measurable simplicity for an unmeasurable saving.
- **The `[WARN]` block prints on every pull request for as long as `main` violates
  the invariant.** Dismissed: it is non-blocking by construction, and suppressing
  it would leave a reader who *does* look with no explanation for why the base
  branch's build is red. The plan records this as an accepted risk.
- **`[WARN]` goes to stderr alongside `[FAIL]`.** Dismissed: every other finding
  in this linter goes to stderr, and the prefix discriminates. Splitting streams
  by severity would be a new convention introduced in passing.

hcross:
## Verdict: APPROVE

REVIEW iter:2 on `33f9e82`. **Zero findings of any class.** The single `tech`
finding from iter:1 is fixed, and the fix was falsified before and verified after
rather than assumed.

Same independence disclosure as iter:1: this review was not cold-spawned, and the
shared-identity posting path applies (`gh api user` → `hcross` = PR author), so
this plain comment is the binding verdict artifact per the skill's rung 2.

### CI status

All **13** required checks **pass** on `33f9e82`, the exact head commit reviewed.
None failing, none pending. `lint-specs` on that commit:

```text
Results: 39 passed, 0 failed
[WARN] Non-delta specs present on the base branch (origin/main) carry 'status: draft':
Linting passed!
```

The `[WARN]`-and-pass is still the live demonstration: `main` continues to carry a
`draft` spec, and this branch is green under the condition that fails every other
open pull request.

### Correctness of the iter:1 remediation

The incremental diff `3e66fe4..33f9e82` is **9 insertions, 4 deletions in one
file** — three `git -C "$GITFIX2" checkout -q -- .` lines and the header comment
rewritten from a hazard warning into a statement of intent. No production code
touched; `scripts/lib/spec-linter.js` and `docs/spec-format.md` are byte-identical
to iter:1.

The finding was verified to be real before being fixed, by reversing the three
case blocks against the same fixture:

| Variant, blocks reversed to 37 → 36 → 35 | Result |
|---|---|
| iter:1 accumulating version | **Case 36 FAILS** (exit 1 where 0 required) — 38 passed, 1 failed |
| iter:2 reset version | 39 passed, 0 failed |

The mechanism is exactly as diagnosed: reversed, Case 37's append to
`specs/0210-attributed.md` persisted into Case 36, so Case 36 no longer had an
*untouched* offender and was correctly failed by the linter while asserting exit
0. So the iter:1 finding was not a style objection — the suite genuinely measured
something other than what two of its case names claimed under any reordering. It
now survives reordering.

Declared order also still passes: **39 passed, 0 failed**.

### Convention compliance

- Commit `33f9e82` uses Gitmoji `✅`, appropriate for a test-only change, and its
  body records the falsification rather than asserting the fix.
- Everything verified in iter:1 is unchanged: branch naming, spec-PR ordering,
  `Closes #732` on the implementation PR only, no version-bump trigger, no
  CLI-matrix trigger surface, pre-edit guard satisfied.

### Test coverage

Unchanged in count — cases 35–38, suite at 39 — and strengthened in kind: each of
the three shared-fixture cases now establishes its own precondition instead of
inheriting the previous case's. The falsification matrix from iter:1 still holds
(R9 → 36, R10 → 35 with 28/29, R11 → 38), and the disclosed limit of Case 37 is
unchanged and still disclosed in the PR body.

### Linter findings

`lint-markdown.sh docs/spec-format.md` — exit **0**, no output (file untouched
since iter:1).

`lint-shell.sh scripts/tests/test-spec-linter.sh` — exit **1**, three SC2086
(info) findings at lines **67, 544, 546**. Identical set to iter:1: same three
lines, no new one. All three are in the pre-existing `run_case` / `run_base_case`
helpers, outside this PR's hunks, where the unquoted expansion is deliberate
(several targets passed as one string, relying on word splitting). The three
lines added by iter:2 introduce no new finding. Not a finding against this PR.

### Termination

`AGENTS.md` → *Retroactive review loop* → *Termination* requires an APPROVE
verdict, zero findings of any class, and green CI on the head commit reviewed. All
three hold on `33f9e82`: verdict APPROVE, zero findings, 13/13 pass. The
lifecycle terminates at MERGE, subject to the user's merge authorization.

PR #649, opened 2026-07-23
♻️ Align pr-logbook and harness-curator to CLI-first forge access

hcross:
Two framework skills still described a GitHub-MCP-first path for their forge writes, which does not exist in any post-0090 CLI-only deployment. This realigns `pr-logbook` and `harness-curator` to lead with the forge CLI (`gh`, with `glab`/`tea` analogues), demoting a host-provided MCP to an optional-if-present convenience.

## How to read this PR?

Two source edits carry the whole change; the four `pr-logbook` outputs are mechanical rebuilds.

1. Start with `artifacts/core/skills/pr-logbook/SKILL.md` — three spots: §4 (logbook-append), §6 (pre-push exec-bit note), §7 (Rule C close).
2. Then `artifacts/library/skills/harness-curator/SKILL.md` — the single "Open them yourself" bullet.
3. The four deployed `pr-logbook` outputs (`.agents/`, `.claude/`, `.gemini/`, `.github/`) are `build-components.sh` regenerations of source #1 with identical wording — skim one, trust the rest (parity is checked in CI).

Non-obvious decisions:

- **`harness-curator` is library-tier**: it builds to a gitignored `dist/`, so there is no committed deployed output to stage — only the source changes.
- **§6 push_files generalization** is an in-file consistency cleanup riding along. The exec-bit-loss note previously named the MCP `push_files` tool as the culprit; it was the last MCP-transport assumption left in the file, so it is generalized to transport-agnostic wording rather than left as a lone stale reference.
- **pr-reviewer is out of scope**: it carried the same stale mandate but was already realigned in #634, so this PR does not touch it. The wording here mirrors what #634 applied there.

## How to test this PR?

Prerequisites: `gh` authenticated, `markdownlint-cli` installed, branch checked out at head `c5ffcff`.

```sh
# 1. Version-bump gate — both modified sources must bump
bash scripts/check-skill-versions.sh                    # expect PASS

# 2. Build drift — deployed outputs match sources
bash scripts/build-components.sh --target all --check   # expect clean (no drift)

# 3. CLI parity
bash scripts/check-ci-parity.sh                         # expect OK

# 4. Lint the two changed sources
markdownlint artifacts/core/skills/pr-logbook/SKILL.md \
             artifacts/library/skills/harness-curator/SKILL.md   # expect clean

# 5. No residual forge-MCP mandate on the changed sources
git grep -n 'forge MCP' -- artifacts/core/skills/pr-logbook/SKILL.md \
[...]

PR #646, opened 2026-07-22
📝 Add byte-accounting discipline to architect skill

gfourny-sfeir:
Adds a byte-accounting discipline step (new "## Operating mode" step 4) to the architect skill so budget-constrained plans cite measured savings — `wc -c` on the actual candidate span — instead of line-count estimates, with an explicit safety margin and a pre-finalization self-check. This closes a harness-friction cluster in which an unmeasured PLAN overshot its budget and forced an unplanned DEV-stage extraction (#498, PR #500).

Closes #601

## How to read this PR?

Start with `artifacts/core/skills/architect/SKILL.md` — the only hand-edited file (39 insertions, 3 deletions per `git diff origin/main HEAD --stat`). The new content sits between the existing "### 3. Ripple-effect analysis" and the renumbered "### 5. Output formats" (was step 4); "### 6. Briefing sub-agents" is the renumbered old step 5. The four mirrors (`.claude/skills/architect/SKILL.md`, `.gemini/skills/architect/SKILL.md`, `.github/skills/architect/SKILL.md`, `.agents/skills/architect/SKILL.md`) are mechanical regenerations via `scripts/build-components.sh` — each shows the identical +39/-3 diff, no hand-editing needed there.

Also note the `provenance.version` bump from `1.1.2` to `1.2.0` (MINOR, additive) in the source file's frontmatter, per AGENTS.md → Version Bump Convention.

## How to test this PR?

1. `git diff origin/main HEAD -- artifacts/core/skills/architect/SKILL.md` and confirm the new "### 4. Byte-accounting discipline (size/budget-constrained plans)" section reads correctly and the old steps 4/5 are renumbered to 5/6.
2. Run `scripts/check-skill-versions.sh` (or `task check-skill-versions`) locally and confirm it passes given the version bump.
3. Diff each of the four mirrored copies under `.claude/`, `.gemini/`, `.github/`, `.agents/` against `artifacts/core/skills/architect/SKILL.md` and confirm they're byte-identical — this is what `scripts/build-components.sh` regenerates; no manual edits are expected there.
4. There is no runtime behavior to exercise: this is a prompt-only change to the architect skill's own instructions, verified in practice by the next PLAN that invokes it.

## Detailed description (for agents)

[...]

PR #634, opened 2026-07-22
🐛 Give pr-reviewer a CLI-only verdict-posting fallback ladder (#595)

gfourny-sfeir:
Moves the PR-verdict posting instruction out of `AGENTS.md` (orchestration-level, invisible to a cold-spawned reviewer) and into the `pr-reviewer` skill/agent sources as an explicit CLI-only fallback ladder. Closes the friction of #595 (recurred 12× after #106/PR #111 patched only the orchestration doc).

## How to read this PR?

Read the two sources first, then the generated outputs:

1. `artifacts/core/skills/pr-reviewer/SKILL.md` — **step 6** is the substance: a preflight identity check (`gh api user` vs PR author) branching into a 3-rung ladder (distinct → formal `gh pr review`; shared → plain `gh pr comment` with a `## Verdict: …` header; denied → return to orchestrator, no permission-laundering).
2. `artifacts/core/agents/pr-reviewer/AGENT.md` — description + step 5 + the canonical-artifact note updated to match; no longer names the dropped GitHub MCP.
3. The 8 generated files under `.claude/`, `.gemini/`, `.github/`, `.agents/` are pure `build-components.sh` output — skim to confirm parity, no hand edits.

**Non-obvious decision:** #595 suggested a 4-rung ladder keeping "GitHub MCP if reachable" as rung 1. Dropped deliberately — #624 made forge access CLI-only and removed the framework forge MCP (`artifacts/core/rules/60-tools.md`: *"The framework ships no forge MCP server for any CLI"*). Keeping an MCP rung would resurrect a mechanism the architecture explicitly retired.

## How to test this PR?

```sh
git fetch origin && git checkout fix/595-pr-reviewer-posting-fallback

# Version-bump CI gate (both sources bumped)
BASE_REF=origin/main bash scripts/check-skill-versions.sh   # expect: OK

# Build is deterministic — regenerating must yield no diff
bash scripts/build-components.sh && git status --short       # expect: clean

# Ladder is present in the generated Claude output
grep -c "Shared identity" .claude/skills/pr-reviewer/SKILL.md # expect: >= 1
```

## Detailed description (for agents)

**Modified sources (2):**

- `artifacts/core/skills/pr-reviewer/SKILL.md` — step 6 rewritten from the unconditional "Use the GitHub MCP `pull_request_review_write` tool" into a CLI-only ladder with an up-front identity resolution. Version `1.1.5` → `1.1.6` (PATCH — friction fix).
[...]

PR #648, opened 2026-07-22
📝 Add Rule 5 — orchestrator edit-fence for delegated deliverables

gfourny-sfeir:
Adds Rule 5 to `docs/agent-team-protocol.md` → *Team Communication*, fencing the orchestrator from editing a deliverable file still delegated to a sub-agent until completion is actually confirmed. Realizes spec `specs/0095-orchestrator-deliverable-edit-fence.md` (spec-PR #647, merged) per PLAN v2 (user-validated: https://github.com/crewrig/crewrig/issues/602#issuecomment-5049538090).

## How to read this PR?

Single-file diff: `docs/agent-team-protocol.md`. Read the new Rule 5 block in place, right after the existing Rule 4 and before *Team Shutdown* (`docs/agent-team-protocol.md`, +35/-0 — see diff). Rule 5 is the write-side counterpart to Rule 3: Rule 3 already establishes that an idle notification does not prove a teammate skipped its Rule 1 report on the *read* side; Rule 5 draws the corresponding consequence for the orchestrator's own edits on the *write* side — the same unreliable idle-notification signal must not be acted upon either. The remediation path for a still-needed change to a still-delegated file is two-pronged: instruct the live sub-agent via `SendMessage` to make the change itself, or confirm conclusion (a landed Rule 1 result, or a Rule 3 step 2 side-effect check) before the orchestrator takes ownership directly.

## How to test this PR?

This is a documentation-only behavioral rule with no runtime behavior to execute. Verification is a read-through: open `docs/agent-team-protocol.md` and confirm Rule 5 reads coherently in context against Rule 1 (`SendMessage` result reporting), Rule 3 (idle-notification unreliability, side-effect check), and *Team Shutdown* (conclusion check) — the cross-references it depends on.

## Detailed description (for agents)

Closes the `concurrent-deliverable-edit` friction cluster (issue #602): a Harness Curator report of an incident (ticket #569) where the orchestrator's own completion edits raced a delegated sub-agent's edits to the same file, producing duplicate content requiring manual reconciliation.

[...]

PR #770, opened 2026-08-09
✨ Give every forge artifact a single writer at a time

### LINKED ISSUE #735: 🐛 Concurrent agent writes to a forge artifact are a silent last-writer-wins race
> **Friction report — filed directly as a GitHub issue.** The MemPalace MCP server disconnected mid-session, so `harness-report` → `mempalace_add_drawer` was unreachable. Same write-path fallback as spec 0103 / #685 / #730. Category `process`, severity `med`, subcategory `forge-artifact-concurrent-write`.

## Friction

`docs/agent-team-protocol.md` mandates **worktree isolation** so concurrent agents cannot clobber each other's file edits. Nothing in the protocol protects a **forge artifact** — a pull-request body, an issue body, a review comment. Two agents on the same ticket can both be authorised to write one PR body, and:

- there is no lock and no ownership rule;
- there is no conflict detection — `gh pr edit` does not take a base revision;
- **`gh pr edit` reports success on a lost update**, so the losing writer has no signal at all.

The result is a silent last-writer-wins race on the one artifact an independent reviewer reads.

## Live occurrence

**2026-08-08**, ticket #726, PR #734. Both writers were acting correctly under their own instructions:

| Time | Actor | Action |
|---|---|---|
| ~11:29 | `pr-logbook` | Opened PR #734 |
| ~11:31 | team-lead | Edited the body (plan-version citations + a governance blockquote); `gh pr edit` reported success; read-back immediately after confirmed the edit was live |
| 11:31:59 | `pr-logbook` | Republished its own revision — **overwriting the team-lead's edit** |
| ~11:33 | team-lead | Asserted to the user that its version was live, from the earlier read-back |
| ~11:34 | `pr-logbook` | Read the live body, found the team-lead's blockquote absent, and reported the lost update |

Neither agent did anything wrong. The team-lead had told `pr-logbook` "do not re-apply"; that message arrived *after* `pr-logbook` had already published, because it had independently caught the same problem from an earlier briefing. Both fixes were correct; one silently replaced the other.

## Cost

Two costs, and the second is the expensive one.

1. **A stale assertion to the user.** The team-lead verified its edit, the verification expired, and it reported a state that was no longer true. A read-back is only valid until the next writer.
2. **A partial regression that survived**, and was caught only by an explicit read-back. The surviving body flagged **three** post-plan behaviours where the overwritten version flagged **five**. The two lost items — a malformed-carrier refusal and an `--id`-on-occupied-ref semantics — were still documented elsewhere in the body, but no longer flagged as surface that post-dates the approved plan. A conformance reviewer would therefore have checked them against a plan revision that has no entry for them, which is the exact failure the flag exists to prevent.

A cold reviewer was mid-read while this was happening.

## Suggested direction (not prescriptive)

1. **Single-writer ownership per forge artifact**, declared when the artifact is created — the agent that opens a PR owns its body until it hands ownership over explicitly. Cheapest fix; needs only a protocol sentence.
2. **Mandatory read-back before asserting state**, with the assertion carrying the observed `updated_at`. Does not prevent the race, but makes it detectable rather than silent.
3. **Optimistic concurrency where the forge supports it.** The GitHub API does not offer a compare-and-swap on a PR body, but `updated_at` can be read before and after a write to detect that someone else wrote in between.

Direction 1 alone would have prevented this occurrence.

## Evidence

- PR #734 `updated_at` history: `11:31:59Z` (the overwriting write), `11:35:23Z` (the repair).
- Issue #726, the logbook comments recording the coordination decisions.
- `docs/agent-team-protocol.md` → *Worktree Isolation* — the guarantee that exists for files and not for forge artifacts.
- Adjacent but distinct: #730 (`idle_notification` timing), and the ticket's own logbook note on a decision reversed in a direct message leaving the durable record stating the reversed version. All three are the same family — **a correction that exists in one place while the record says another** — with different mechanisms.

hcross:
Two agents on one ticket can both write a pull-request body, an issue body, or a comment through the same forge, and neither learns that it overwrote the other — worktree isolation protects none of it, because a forge artifact lives outside every worktree. This PR realizes spec 0115 and its delta-01 by giving every such artifact exactly one writer at a time: a role, scoped to the ticket, that changes hands only through an explicit handover performed by the writer itself.

Closes #735

## How to read this PR?

Read `docs/agent-team-protocol.md` first — everything else follows from it.

1. **`docs/agent-team-protocol.md`** — the new **Rule 6 — Single writer per forge artifact**, inserted immediately after Rule 5 and before `## Team Shutdown`, plus the section preamble at `:336` going from "Five rules" to "Six rules". Rule 6 is **90 lines**, against the cold review's ~85–90 target and Rule 5's 34. Eleven bolded leads, one per obligation, so a reader consulting it mid-task can jump to the one they need. The two paragraphs worth reading twice are *The writer is a role on a ticket, not an agent instance* (why writership survives a session) and *Relation to Rule 5, and to worktree isolation* (the conditional/unconditional asymmetry that makes an instance's conclusion the one event that must **not** move writership).
2. **`artifacts/core/skills/pr-logbook/SKILL.md`** — a six-bullet cross-cutting section, so the role that actually opens pull requests observes the rule with only its own contract loaded (spec R10).
3. **`artifacts/core/skills/pr-reviewer/SKILL.md`** — a shorter section: a reviewer's entire verdict *is* an assertion about an artifact's content.
4. **The eight build outputs** under `.claude/`, `.gemini/`, `.github/`, `.agents/` — mechanical, generated by `scripts/build-components.sh`, staged in the same commit per `AGENTS.md` → *Built components*. Skim or skip.

**Two non-obvious decisions, both deliberate:**

- **`AGENTS.md` is not touched.** Rule 5 — which spec R11 makes Rule 6's counterpart — is itself absent from `AGENTS.md`'s *Critical rules* list (`AGENTS.md:256-262`), so Rule 6 inherits exactly Rule 5's discoverability, which is the parity R11 asks for. `AGENTS.md` → *Agent Team Protocol* already points every team at this document for the team-communication rules in aggregate. (The byte budget agrees but is the weaker argument: 21 879 B against a `-ge 22000` gate leaves 120 B of headroom, and this PR spends none of it.)
- **`docs/cli-matrix.md` is consulted but unchanged.** The trigger surface is hit via `artifacts/**` and the *Protocol-only exemption* does not apply. Both skills are already covered generically by matrix row 3 (*Skill definitions directory (built output)*); this change adds prose to existing components and introduces no path, command, hook, or per-CLI configuration, so no cell moves. Four merged precedents did the same — `7d56c91` (#646), `bf785d8` (#649), `df22b0b` (#634), `c51e938` (#652), each touching `artifacts/core/skills/*/SKILL.md` with zero `docs/cli-matrix.md` hunks (verified in this worktree with `git show --name-only`). The absent hunk is a conclusion, not an omission.

## How to test this PR?

No new check ships — spec R12 forecloses mechanical enforcement, and the four existing CI jobs cover every mechanical consequence. From a worktree on this branch:

```sh
# 1. Built outputs match their sources (CI: check-components)
bash scripts/build-components.sh --target all --check
# → exit 0 — "OK: All generated files match source."

# 2. Both skill sources carry a version bump (CI: check-skill-versions)
bash scripts/check-skill-versions.sh
# → exit 0 — pr-logbook and pr-reviewer both OK

# 3. AGENTS.md still under budget (CI: check-agents-size)
bash scripts/check-agents-size.sh
# → exit 0 — 21879 bytes, threshold 22000, unchanged by this PR

# 4. Markdown, exactly as CI runs it (CI: lint-markdown)
npx markdownlint-cli "**/*.md" --ignore node_modules --ignore extension-skeleton \
  --ignore communication --ignore 'extensions/**/commands/*.md'
# → exit 0

# 5. Specs (CI: lint-specs)
node scripts/lib/spec-linter.js specs/
# → exit 0 — 0114's draft-status breakage is cured on main

# 6. Byte-level hygiene: no tool scaffolding leaked into any authored file
/usr/bin/grep -nE '</?(content|invoke|parameter|function_calls|antml)' \
  docs/agent-team-protocol.md artifacts/core/skills/pr-{logbook,reviewer}/SKILL.md
# → exit 1, no match. Run this rather than trusting a linter: the spec linter
#   returned exit 0 on a file that shipped literal closing-tag scaffolding
#   earlier on this very ticket. A green check answers its own question,
#   not the one you are asking.
```

The behavioural obligation itself is verified by reading, not by executing — that is what R12 chose. Read Rule 6 against the fourteen requirements of `specs/0115-forge-artifact-single-writer.md` as amended by `specs/0115-forge-artifact-single-writer.delta-01.md`; the traceability table in `## PLAN v2` on #735 maps each one to its paragraph.

### `docs/agent-team-protocol.md`

**Added — Rule 6, 90 lines, after Rule 5 and before `## Team Shutdown`.** Eleven paragraphs, each opening with a bolded lead naming its obligation:

| Paragraph | Carries |
|---|---|
| *Rule 6 — Single writer per forge artifact* | R2 (definition + new-comment carve-out), R3 first clause |
| *The writer is a role on a ticket, not an agent instance* | R3 as replaced by delta-01 — role-scoped, ticket-bound, an instance's conclusion ends nothing; no per-instance register wanted (delta *Out of scope* 6) |
| *Determining the writer, without asking it* | R4 — a forward walk of the writership chain, resolving without an exchange with the current writer and defaulting to "you are not the writer". Starts at the creating role's `**Writer:**` line and applies each handover oldest first, accepting one only where its left-hand role is the writer the walk has reached |
| *Already-published comments resolve at rung 3, by design* | The PLAN's decision 6, answering the v1 review's Finding 3 — comments are out of the recording mechanism **intentionally**, no per-comment writer line is owed |
| *A non-writer does not write* | R5 as replaced — the remediation set closed at two branches, naming the three limbs it refuses (a confirmed conclusion, a landed completion report, an observed side effect) |
| *Two live agents of one role — the orchestrator's obligation* | R13 as amended — binds the **orchestrator**, not the two agents, since each resolves the ladder to its own role and cannot observe the other; R7/R8 still bind each writing agent |
| *Handover, and who may perform one* | R6 + R14 — only the current writer may hand over; a nomination from anyone else confers nothing however explicitly stated or recorded; the recipient determines per R4 and treats itself as non-writer unless the determination names **its role**. The walk above enforces this procedurally, so an agent following the steps literally cannot reach a nominee (REVIEW iter:1, `911a059`) |
| *Assert only what you have just observed* | R7, **both halves** — an unsuperseded observation *and* the last-modification marker carried with the assertion |
| *Detect the lost update; a success report is not proof* | R8 + R9 — content-level comparison as the obligation, the forge's marker as the cheap form where one exists, and the logbook report when the artifact moved |
| *Relation to Rule 5, and to worktree isolation* | R11 — the counterpart declaration, the conditional/unconditional asymmetry, and that worktree isolation affords a forge artifact nothing |

The two record forms — `**Writer:**` and `**Writership handover:**` — ship as one fenced `markdown` block serving both rungs, rather than a fence per rung. Both lines ride inside logbook comments the ticket already gets; no separate register comment is created, which is the cost decision the PLAN's decision 1 turns on.

**Modified — `:336`.** `Five rules govern…` → `Six rules govern…`. The preamble is not the text of Rules 1–5, so it sits outside spec 0115's out-of-scope bullet; `088af38` set the identical precedent under spec 0095's identically-worded bullet.

**Not modified.** Rules 1–5 and *Worktree Isolation* keep their existing text, per that same out-of-scope bullet. The only new coupling is the outward reference Rule 6 makes to Rule 5.

**On length.** The cold re-review recorded Rule 6's drafted text at 119 lines against Rule 5's 24 and asked for ~85–90 without dropping a paragraph. Delivered at 90, then **99** after REVIEW iter:1 added the chain-validation clause and reconciled a comment's writer — the reviewer relaxed the constraint for that fix. The cuts are the four spans it named — the derivation of role-scoping (which delta-01 argues at length), the measurement behind "the authorship field names the account", the cost rationale for rejecting per-comment writer lines, and the PR #734 re-narration. Every obligation survives; only the *why* was thinned, and it remains recoverable from the spec, the delta, and the plan.

### `artifacts/core/skills/pr-logbook/SKILL.md` — `1.1.7` → `1.2.0`

New `## Cross-cutting: single writer per forge artifact`, placed after the existing version-bump cross-cutting section. Six bullets: your role owns what it creates (and names itself on a `**Writer:**` line); you do not rewrite what your role did not create; a handover is only good from the current writer (R14); observe immediately before you write; a success report is not proof nothing was lost; **assert only a fresh observation, and carry its marker**.

That last bullet is the v1 review's Finding 4, `pr-logbook` half. The draft carried only R7's freshness clause and left the marker to a bullet framed around *writing*, which is R8. The marker-with-the-assertion is the half that actually failed on PR #734, so it belongs in the assertion bullet — an agent holding only this contract is now told to carry it.

MINOR per `AGENTS.md` → *Version Bump Convention*: a new section, additive, no existing obligation withdrawn.

### `artifacts/core/skills/pr-reviewer/SKILL.md` — `1.1.7` → `1.2.0`

New `## Cross-cutting: assert only a fresh observation`, placed after `### 6. Post the review`. Spec R10's "at minimum" does not name this role; it is included because a reviewer's verdict *is* an assertion about an artifact's current content, and the friction report records a cold reviewer mid-read while PR #734's body was being republished under it.

The section carries the marker instruction **and** R9's marker-absent fallback — the v1 review's Finding 4, `pr-reviewer` half: "Where a forge reports no such marker, compare the body text itself against what you last read — the obligation is to know whether the artifact moved, not to read a particular field, and it holds on every forge the *Forge Access* policy admits." No `glab` or `tea` field name is written anywhere in this diff: the PLAN flagged those as unverified, and decision 4 makes the obligation content-level so an unknown field name degrades to a body comparison rather than to no detection.

### Build outputs — 8 files

`{.claude,.gemini,.github,.agents}/skills/{pr-logbook,pr-reviewer}/SKILL.md`, regenerated by `bash scripts/build-components.sh` and staged in the same commit per `AGENTS.md` → *Agent Team Protocol* → *Built components*. None was hand-edited; `--target all --check` passes.

### Not touched, deliberately

`AGENTS.md` (see *How to read this PR?*); `artifacts/core/agents/{pr-logbook,pr-reviewer}/AGENT.md` — both personas already route to their skill, and restating there costs two more version bumps and eight more outputs to carry a pointer that can drift; `docs/cli-matrix.md`; every file under `scripts/`; `.crewrig/core-paths.txt` — `docs/layers.md` gains and loses no path; `docs/index.json` — `docs/agent-team-protocol.md` stays `published=false`, and `build-docs-index.sh --check` confirms no drift.

### Traceability

All fourteen requirements are covered. R1 (`complexity: standard`) is a team-composition requirement rather than a file edit — `architect` reviewed at PLAN and the plan was re-reviewed cold before DEV. The remaining thirteen map to the paragraphs and bullets tabulated above; the full requirement→step table lives in `## PLAN v2` on #735.

---

**REVIEW iter:1** (`911a059`) — the determination's first rung accepted the most recent handover line without testing who wrote it, so a nomination from a non-writer resolved to the nominee: delta-01 Scenario 5 failed against the shipped procedure, which is the incident R14 exists to prevent. The norm was in the document three paragraphs below, but a procedure that terminates with an answer is never abandoned partway for a principle stated later. The determination now walks the chain forward from the creating role's `**Writer:**` line, accepting a handover only where its left-hand role is the writer the walk has reached. Also reconciled a comment's two writers — its author at publication, nobody thereafter. The scenario walk that verifies it is on the [logbook](https://github.com/crewrig/crewrig/issues/735#issuecomment-5230736155).

### DIFF
--- a/.agents/skills/pr-logbook/SKILL.md
+++ b/.agents/skills/pr-logbook/SKILL.md
@@ -7,7 +7,7 @@ metadata:
   provenance:
     canonical: "https://github.com/crewrig/crewrig"
     feedback: "https://github.com/crewrig/crewrig"
-    version: "1.1.7"
+    version: "1.2.0"
 ---
 
 
@@ -175,6 +175,41 @@ SemVer applies:
 A "version-only bump" PR is not a thing — the version bump always
 accompanies the content edit. See `artifacts/FORMAT.md` →
 *Version semantics* for the contract.
+
+## Cross-cutting: single writer per forge artifact
+
+Not a step in the composition lifecycle — a rule that applies to every
+forge artifact you write. The full rule is `docs/agent-team-protocol.md`
+→ *Team Communication* → Rule 6; below is the part you must observe even
+when this skill is the only contract you have loaded.
+
+- **Your role owns what your role creates.** Opening a pull request makes
+  `pr-logbook` the writer of its body for that ticket — the role, not
+  your instance, so the writership survives your session. Name it once,
+  on a `**Writer:**` line inside the logbook entry you already write for
+  that artifact; see Rule 6 for the exact form.
+- **You do not rewrite what your role did not create.** Another role's PR
+  body, issue body, or comment is not yours to republish; route the
+  change through its writer, or take a recorded handover first. Posting a
+  *new* comment is always allowed — it creates a new artifact you own.
+  Already-published comments are written by nobody, by design.
+- **A handover is only good from the current writer.** Being told an
+  artifact is yours confers nothing unless the teller is its current
+  writer. Determine the writer per Rule 6 before you write, and treat
+  yourself as not being it unless that determination names your role.
+- **Observe immediately before you write.** Capture the artifact's
+  last-modification marker where the forge reports one
+  (`gh pr view <n> --json updatedAt`; the `glab` / `tea` equivalents
+  elsewhere), or its body text where it does not.
+- **A success report is not proof nothing was lost.** `gh pr edit`
+  reports success on a lost update. If the artifact moved between your
+  observation and your write, say so on the logbook — an unreported
+  overwrite costs more than the overwrite.
+- **Assert only a fresh observation, and carry its marker.** When you
+  tell the user or a teammate what an artifact currently says, base it on
+  an observation no later write has superseded and quote the
+  last-modification marker the forge reported at that observation. A
+  read-back is valid until the next writer, not longer.
 
 ## Grounding discipline
 
--- a/.agents/skills/pr-reviewer/SKILL.md
+++ b/.agents/skills/pr-reviewer/SKILL.md
@@ -7,7 +7,7 @@ metadata:
   provenance:
     canonical: "https://github.com/crewrig/crewrig"
     feedback: "https://github.com/crewrig/crewrig"
-    version: "1.1.7"
+    version: "1.2.0"
 ---
 
 
@@ -164,6 +164,22 @@ The three events map as follows:
   merge.
 - `COMMENT` — observations without a verdict (e.g. when the diff is
   outside the reviewer's domain).
+
+## Cross-cutting: assert only a fresh observation
+
+Your verdict asserts what a pull request currently contains. Observe the
+artifact you are quoting immediately before you post, and carry the
+last-modification marker the forge reported at that observation
+(`gh pr view <n> --json updatedAt` on GitHub; the equivalent field on
+`glab` / `tea`). Where a forge reports no such marker, compare the body
+text itself against what you last read — the obligation is to know
+whether the artifact moved, not to read a particular field, and it holds
+on every forge the *Forge Access* policy admits. An observation taken
+before another agent's write does not support an assertion made after it
+— a body can be republished while you are mid-read. Your verdict comment
+is a new artifact and yours to write; the pull-request body belongs to
+the role that opened it. Full rule: `docs/agent-team-protocol.md` →
+*Team Communication* → Rule 6.
 
 ## Scripts
 
--- a/.claude/skills/pr-logbook/SKILL.md
+++ b/.claude/skills/pr-logbook/SKILL.md
@@ -11,7 +11,7 @@ metadata:
   provenance:
     canonical: "https://github.com/crewrig/crewrig"
     feedback: "https://github.com/crewrig/crewrig"
-    version: "1.1.7"
+    version: "1.2.0"
 ---
 
 
@@ -179,6 +179,41 @@ SemVer applies:
 A "version-only bump" PR is not a thing — the version bump always
 accompanies the content edit. See `artifacts/FORMAT.md` →
 *Version semantics* for the contract.
+
+## Cross-cutting: single writer per forge artifact
+
+Not a step in the composition lifecycle — a rule that applies to every
+forge artifact you write. The full rule is `docs/agent-team-protocol.md`
+→ *Team Communication* → Rule 6; below is the part you must observe even
+when this skill is the only contract you have loaded.
+
+- **Your role owns what your role creates.** Opening a pull request makes
+  `pr-logbook` the writer of its body for that ticket — the role, not
+  your instance, so the writership survives your session. Name it once,
+  on a `**Writer:**` line inside the logbook entry you already write for
+  that artifact; see Rule 6 for the exact form.
+- **You do not rewrite what your role did not create.** Another role's PR
+  body, issue body, or comment is not yours to republish; route the
+  change through its writer, or take a recorded handover first. Posting a
+  *new* comment is always allowed — it creates a new artifact you own.
+  Already-published comments are written by nobody, by design.
+- **A handover is only good from the current writer.** Being told an
+  artifact is yours confers nothing unless the teller is its current
+  writer. Determine the writer per Rule 6 before you write, and treat
+  yourself as not being it unless that determination names your role.
+- **Observe immediately before you write.** Capture the artifact's
+  last-modification marker where the forge reports one
+  (`gh pr view <n> --json updatedAt`; the `glab` / `tea` equivalents
+  elsewhere), or its body text where it does not.
+- **A success report is not proof nothing was lost.** `gh pr edit`
+  reports success on a lost update. If the artifact moved between your
+  observation and your write, say so on the logbook — an unreported
+  overwrite costs more than the overwrite.
+- **Assert only a fresh observation, and carry its marker.** When you
+  tell the user or a teammate what an artifact currently says, base it on
+  an observation no later write has superseded and quote the
+  last-modification marker the forge reported at that observation. A
+  read-back is valid until the next writer, not longer.
 
 ## Grounding discipline
 
--- a/.claude/skills/pr-reviewer/SKILL.md
+++ b/.claude/skills/pr-reviewer/SKILL.md
@@ -12,7 +12,7 @@ metadata:
   provenance:
     canonical: "https://github.com/crewrig/crewrig"
     feedback: "https://github.com/crewrig/crewrig"
-    version: "1.1.7"
+    version: "1.2.0"
 ---
 
 
@@ -169,6 +169,22 @@ The three events map as follows:
   merge.
 - `COMMENT` — observations without a verdict (e.g. when the diff is
   outside the reviewer's domain).
+
+## Cross-cutting: assert only a fresh observation
+
+Your verdict asserts what a pull request currently contains. Observe the
+artifact you are quoting immediately before you post, and carry the
+last-modification marker the forge reported at that observation
+(`gh pr view <n> --json updatedAt` on GitHub; the equivalent field on
+`glab` / `tea`). Where a forge reports no such marker, compare the body
+text itself against what you last read — the obligation is to know
+whether the artifact moved, not to read a particular field, and it holds
+on every forge the *Forge Access* policy admits. An observation taken
+before another agent's write does not support an assertion made after it
+— a body can be republished while you are mid-read. Your verdict comment
+is a new artifact and yours to write; the pull-request body belongs to
+the role that opened it. Full rule: `docs/agent-team-protocol.md` →
+*Team Communication* → Rule 6.
 
 ## Scripts
 
--- a/.gemini/skills/pr-logbook/SKILL.md
+++ b/.gemini/skills/pr-logbook/SKILL.md
@@ -7,7 +7,7 @@ metadata:
   provenance:
     canonical: "https://github.com/crewrig/crewrig"
     feedback: "https://github.com/crewrig/crewrig"
-    version: "1.1.7"
+    version: "1.2.0"
 ---
 
 
@@ -175,6 +175,41 @@ SemVer applies:
 A "version-only bump" PR is not a thing — the version bump always
 accompanies the content edit. See `artifacts/FORMAT.md` →
 *Version semantics* for the contract.
+
+## Cross-cutting: single writer per forge artifact
+
+Not a step in the composition lifecycle — a rule that applies to every
+forge artifact you write. The full rule is `docs/agent-team-protocol.md`
+→ *Team Communication* → Rule 6; below is the part you must observe even
+when this skill is the only contract you have loaded.
+
+- **Your role owns what your role creates.** Opening a pull request makes
+  `pr-logbook` the writer of its body for that ticket — the role, not
+  your instance, so the writership survives your session. Name it once,
+  on a `**Writer:**` line inside the logbook entry you already write for
+  that artifact; see Rule 6 for the exact form.
+- **You do not rewrite what your role did not create.** Another role's PR
+  body, issue body, or comment is not yours to republish; route the
+  change through its writer, or take a recorded handover first. Posting a
+  *new* comment is always allowed — it creates a new artifact you own.
+  Already-published comments are written by nobody, by design.
+- **A handover is only good from the current writer.** Being told an
+  artifact is yours confers nothing unless the teller is its current
+  writer. Determine the writer per Rule 6 before you write, and treat
+  yourself as not being it unless that determination names your role.
+- **Observe immediately before you write.** Capture the artifact's
+  last-modification marker where the forge reports one
+  (`gh pr view <n> --json updatedAt`; the `glab` / `tea` equivalents
+  elsewhere), or its body text where it does not.
+- **A success report is not proof nothing was lost.** `gh pr edit`
+  reports success on a lost update. If the artifact moved between your
+  observation and your write, say so on the logbook — an unreported
+  overwrite costs more than the overwrite.
+- **Assert only a fresh observation, and carry its marker.** When you
+  tell the user or a teammate what an artifact currently says, base it on
+  an observation no later write has superseded and quote the
+  last-modification marker the forge reported at that observation. A
+  read-back is valid until the next writer, not longer.
 
 ## Grounding discipline
 
--- a/.gemini/skills/pr-reviewer/SKILL.md
+++ b/.gemini/skills/pr-reviewer/SKILL.md
@@ -7,7 +7,7 @@ metadata:
   provenance:
     canonical: "https://github.com/crewrig/crewrig"
     feedback: "https://github.com/crewrig/crewrig"
-    version: "1.1.7"
+    version: "1.2.0"
 ---
 
 
@@ -164,6 +164,22 @@ The three events map as follows:
   merge.
 - `COMMENT` — observations without a verdict (e.g. when the diff is
   outside the reviewer's domain).
+
+## Cross-cutting: assert only a fresh observation
+
+Your verdict asserts what a pull request currently contains. Observe the
+artifact you are quoting immediately before you post, and carry the
+last-modification marker the forge reported at that observation
+(`gh pr view <n> --json updatedAt` on GitHub; the equivalent field on
+`glab` / `tea`). Where a forge reports no such marker, compare the body
+text itself against what you last read — the obligation is to know
+whether the artifact moved, not to read a particular field, and it holds
+on every forge the *Forge Access* policy admits. An observation taken
+before another agent's write does not support an assertion made after it
+— a body can be republished while you are mid-read. Your verdict comment
+is a new artifact and yours to write; the pull-request body belongs to
+the role that opened it. Full rule: `docs/agent-team-protocol.md` →
+*Team Communication* → Rule 6.
 
 ## Scripts
 
--- a/.github/skills/pr-logbook/SKILL.md
+++ b/.github/skills/pr-logbook/SKILL.md
@@ -7,7 +7,7 @@ metadata:
   provenance:
     canonical: "https://github.com/crewrig/crewrig"
     feedback: "https://github.com/crewrig/crewrig"
-    version: "1.1.7"
+    version: "1.2.0"
 ---
 
 
@@ -175,6 +175,41 @@ SemVer applies:
 A "version-only bump" PR is not a thing — the version bump always
 accompanies the content edit. See `artifacts/FORMAT.md` →
 *Version semantics* for the contract.
+
+## Cross-cutting: single writer per forge artifact
+
+Not a step in the composition lifecycle — a rule that applies to every
+forge artifact you write. The full rule is `docs/agent-team-protocol.md`
+→ *Team Communication* → Rule 6; below is the part you must observe even
+when this skill is the only contract you have loaded.
+
+- **Your role owns what your role creates.** Opening a pull request makes
+  `pr-logbook` the writer of its body for that ticket — the role, not
+  your instance, so the writership survives your session. Name it once,
+  on a `**Writer:**` line inside the logbook entry you already write for
+  that artifact; see Rule 6 for the exact form.
+- **You do not rewrite what your role did not create.** Another role's PR
+  body, issue body, or comment is not yours to republish; route the
+  change through its writer, or take a recorded handover first. Posting a
+  *new* comment is always allowed — it creates a new artifact you own.
+  Already-published comments are written by nobody, by design.
+- **A handover is only good from the current writer.** Being told an
+  artifact is yours confers nothing unless the teller is its current
+  writer. Determine the writer per Rule 6 before you write, and treat
+  yourself as not being it unless that determination names your role.
+- **Observe immediately before you write.** Capture the artifact's
+  last-modification marker where the forge reports one
+  (`gh pr view <n> --json updatedAt`; the `glab` / `tea` equivalents
+  elsewhere), or its body text where it does not.
+- **A success report is not proof nothing was lost.** `gh pr edit`
+  reports success on a lost update. If the artifact moved between your
+  observation and your write, say so on the logbook — an unreported
+  overwrite costs more than the overwrite.
+- **Assert only a fresh observation, and carry its marker.** When you
+  tell the user or a teammate what an artifact currently says, base it on
+  an observation no later write has superseded and quote the
+  last-modification marker the forge reported at that observation. A
+  read-back is valid until the next writer, not longer.
 
 ## Grounding discipline
 
--- a/.github/skills/pr-reviewer/SKILL.md
+++ b/.github/skills/pr-reviewer/SKILL.md
@@ -7,7 +7,7 @@ metadata:
   provenance:
     canonical: "https://github.com/crewrig/crewrig"
     feedback: "https://github.com/crewrig/crewrig"
-    version: "1.1.7"
+    version: "1.2.0"
 ---
 
 
@@ -164,6 +164,22 @@ The three events map as follows:
   merge.
 - `COMMENT` — observations without a verdict (e.g. when the diff is
   outside the reviewer's domain).
+
+## Cross-cutting: assert only a fresh observation
+
+Your verdict asserts what a pull request currently contains. Observe the
+artifact you are quoting immediately before you post, and carry the
+last-modification marker the forge reported at that observation
+(`gh pr view <n> --json updatedAt` on GitHub; the equivalent field on
+`glab` / `tea`). Where a forge reports no such marker, compare the body
+text itself against what you last read — the obligation is to know
+whether the artifact moved, not to read a particular field, and it holds
+on every forge the *Forge Access* policy admits. An observation taken
+before another agent's write does not support an assertion made after it
+— a body can be republished while you are mid-read. Your verdict comment
+is a new artifact and yours to write; the pull-request body belongs to
+the role that opened it. Full rule: `docs/agent-team-protocol.md` →
+*Team Communication* → Rule 6.
 
 ## Scripts
 
--- a/artifacts/core/skills/pr-logbook/SKILL.md
+++ b/artifacts/core/skills/pr-logbook/SKILL.md
@@ -11,7 +11,7 @@ metadata:
   provenance:
     canonical: "${CANONICAL_REPO}"
     feedback: "${CANONICAL_REPO}"
-    version: "1.1.7"
+    version: "1.2.0"
 claude:
   allowed-tools:
     - Read
@@ -183,6 +183,41 @@ SemVer applies:
 A "version-only bump" PR is not a thing — the version bump always
 accompanies the content edit. See `artifacts/FORMAT.md` →
 *Version semantics* for the contract.
+
+## Cross-cutting: single writer per forge artifact
+
+Not a step in the composition lifecycle — a rule that applies to every
+forge artifact you write. The full rule is `docs/agent-team-protocol.md`
+→ *Team Communication* → Rule 6; below is the part you must observe even
+when this skill is the only contract you have loaded.
+
+- **Your role owns what your role creates.** Opening a pull request makes
+  `pr-logbook` the writer of its body for that ticket — the role, not
+  your instance, so the writership survives your session. Name it once,
+  on a `**Writer:**` line inside the logbook entry you already write for
+  that artifact; see Rule 6 for the exact form.
+- **You do not rewrite what your role did not create.** Another role's PR
+  body, issue body, or comment is not yours to republish; route the
+  change through its writer, or take a recorded handover first. Posting a
+  *new* comment is always allowed — it creates a new artifact you own.
+  Already-published comments are written by nobody, by design.
+- **A handover is only good from the current writer.** Being told an
+  artifact is yours confers nothing unless the teller is its current
+  writer. Determine the writer per Rule 6 before you write, and treat
+  yourself as not being it unless that determination names your role.
+- **Observe immediately before you write.** Capture the artifact's
+  last-modification marker where the forge reports one
+  (`gh pr view <n> --json updatedAt`; the `glab` / `tea` equivalents
+  elsewhere), or its body text where it does not.
+- **A success report is not proof nothing was lost.** `gh pr edit`
+  reports success on a lost update. If the artifact moved between your
+  observation and your write, say so on the logbook — an unreported
+  overwrite costs more than the overwrite.
+- **Assert only a fresh observation, and carry its marker.** When you
+  tell the user or a teammate what an artifact currently says, base it on
+  an observation no later write has superseded and quote the
+  last-modification marker the forge reported at that observation. A
+  read-back is valid until the next writer, not longer.
 
 ## Grounding discipline
 
--- a/artifacts/core/skills/pr-reviewer/SKILL.md
+++ b/artifacts/core/skills/pr-reviewer/SKILL.md
@@ -13,7 +13,7 @@ metadata:
   provenance:
     canonical: "${CANONICAL_REPO}"
     feedback: "${CANONICAL_REPO}"
-    version: "1.1.7"
+    version: "1.2.0"
 claude:
   allowed-tools:
     - Read
@@ -175,6 +175,22 @@ The three events map as follows:
   merge.
 - `COMMENT` — observations without a verdict (e.g. when the diff is
   outside the reviewer's domain).
+
+## Cross-cutting: assert only a fresh observation
+
+Your verdict asserts what a pull request currently contains. Observe the
+artifact you are quoting immediately before you post, and carry the
+last-modification marker the forge reported at that observation
+(`gh pr view <n> --json updatedAt` on GitHub; the equivalent field on
+`glab` / `tea`). Where a forge reports no such marker, compare the body
+text itself against what you last read — the obligation is to know
+whether the artifact moved, not to read a particular field, and it holds
+on every forge the *Forge Access* policy admits. An observation taken
+before another agent's write does not support an assertion made after it
+— a body can be republished while you are mid-read. Your verdict comment
+is a new artifact and yours to write; the pull-request body belongs to
+the role that opened it. Full rule: `docs/agent-team-protocol.md` →
+*Team Communication* → Rule 6.
 
 ## Scripts
 
--- a/docs/agent-team-protocol.md
+++ b/docs/agent-team-protocol.md
@@ -333,7 +333,7 @@ non-blocking observation, not a blocking finding.
 
 ## Team Communication
 
-Five rules govern how teammates report back inside a team and how the team-lead interprets their signals.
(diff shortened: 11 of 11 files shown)


hcross:
## Verdict: REQUEST CHANGES

Cold review of PR #770 at head `acfd631`, against `specs/0115-forge-artifact-single-writer.md` as amended by `specs/0115-forge-artifact-single-writer.delta-01.md` (both read at the head ref, not from a local checkout).

One blocking finding: the writer-determination ladder does not implement R14, the requirement delta-01 was written to add. Everything else in the diff verifies clean — build fidelity, version bumps, byte hygiene, and the remaining thirteen requirements.

---

## CI status

All **12** required checks pass on `acfd631`. `check-components` was pending at review start and finished **pass** (5m46s, job `93222504397`).

`build` · `check-agents-size` · `check-ci-parity` · `check-components` · `check-extension-version-bump` · `check-feedback-routing` · `check-skill-versions` · `check-spec-id-reserved` · `gitlab-ci-check` · `lint-markdown` · `lint-specs` · `test-harness-curate` — pass. No failures, none pending.

The lifecycle's CI-green half of the termination criterion is met on the reviewed head.

---

## Correctness

### Requirement coverage — 14 requirements, 13 met

Mapped against the merged delta's *replacement* R3/R5 and its added R13/R14, not the originals.

| Req | Where realized | Verdict |
|---|---|---|
| R1 `complexity: standard` | spec frontmatter; team composition, not a file edit | met |
| R2 forge-artifact definition + new-comment carve-out | `docs/agent-team-protocol.md:462-467` | met |
| R3 *(replaced)* role-scoped, ticket-bound, conclusion ends nothing | `:469-474` — all four clauses present, incl. delta *Out of scope* 6 ("No per-instance identifier or register is wanted") | met |
| R4 determination without an exchange, default to non-writer | `:476-486` | **see Finding 1** |
| R5 *(replaced)* two-branch set closed, three refused limbs named | `:500-507` — "not on a confirmed conclusion, not on a landed completion report, not on an observed side effect" | met |
| R6 handover only; five non-transfer vectors | `:517-525` — assumption, elapsed time, `idle_notification`, apparent inactivity, peer finished something related — all five | met |
| R7 **both halves** | `:527-531` — unsuperseded observation *and* the marker carried with the assertion | met |
| R8 lost-update detection; success report is not proof | `:533-541` | met |
| R9 forge-neutrality + marker-absent fallback | `:536-539` degrades to body comparison; no `glab`/`tea` field name invented anywhere in the diff | met |
| R10 role contracts carry it | `pr-logbook` + `pr-reviewer` skills + the protocol doc (which *is* the orchestrator contract — no `team-lead` agent exists under `artifacts/core/agents/`) | met |
| R11 counterpart to Rule 5 + worktree isolation affords nothing | `:543-551` | met |
| R12 no lock/lease/mutex/CAS | nothing mechanical added; no new CI check | met |
| R13 binds the **orchestrator** | `:509-515` | met |
| R14 handover valid only from the current writer | `:519-523` states the norm | **Finding 1** |

**On R13 specifically** — the brief flagged that `## PLAN v2` drafted this as binding "the role's agents collectively" while the merged delta binds the orchestrator. The shipped text at `:511-514` reads "An orchestrator that brings a further agent of that role live does not direct more than one of them to write a given artifact", with the rationale the delta gives (each resolves the ladder to its own role and cannot see the other is live). **The diff followed the spec, not the plan.** Correct.

**On the two halves the earlier review found missing** — both are now present:

- R7's marker half in `pr-logbook`: bullet 6, "**Assert only a fresh observation, and carry its marker.**" (`artifacts/core/skills/pr-logbook/SKILL.md:216-220`).
- R9's marker-absent fallback in `pr-reviewer`: "Where a forge reports no such marker, compare the body text itself against what you last read" (`artifacts/core/skills/pr-reviewer/SKILL.md:185-188`).

---

### Finding 1 — the ladder resolves a void nomination to the nominee

`class: tech` · **blocking** · `docs/agent-team-protocol.md:479-480`

Rung 1 of the writer-determination ladder reads:

> 1. **A recorded handover** — the most recent `**Writership handover:**` line on the ticket's logbook for that artifact names the writer.

Unconditional. It accepts the most recent handover line whatever its origin. R14 requires the opposite:

> A handover under requirement 6 SHALL be effective only when the party handing writership over is the artifact's current writer. […] An agent offered writership SHALL determine the artifact's current writer per requirement 4 before writing, and SHALL treat itself as not being the writer unless that determination names its role.

Walk delta-01 Scenario 5 against the shipped text. `pr-logbook` is the writer of a PR body. `team-lead`, not the writer, records `**Writership handover:** PR #N body — pr-logbook → architect` on the logbook — explicitly, on the record, exactly where a determination reads, which is the scenario's premise. `architect` then resolves the ladder before writing. Rung 1 matches: the most recent handover line names `architect`. The determination returns `architect`, `architect` writes.

The scenario requires the opposite outcome:

> Then it determines the current writer per requirement 4 and **finds the first agent's role named**
> And it does not write the body

This is the 2026-08-08 incident the delta was written after — a nomination that "satisfied requirement 6 on its face" and "was void, because its sender had never been the writer." A recipient following Rule 6's ladder literally reproduces it.

The Handover paragraph at `:519-521` does state the norm ("A nomination […] by an agent that is not the current writer confers nothing"), and one can argue rung 1's "a recorded handover" is implicitly defined by it. Two reasons that does not close the gap:

1. The ladder is presented as a self-contained procedure ("An agent about to write resolves writership in this order") and terminates at rung 1 with an answer. The qualifier lives three paragraphs later, at `:517`.
2. Even a reader who cross-applies has **no procedure** for deciding whether a given handover line came from the then-current writer. Rule 6 never says to walk the chain back to the `**Writer:**` line.

The fix is small, because the record form already carries the data. `:490` shows `**Writership handover:** PR #734 body — pr-logbook → team-lead` — the handing-over role is in the left-hand position. Rung 1 needs to count a handover only when its left-hand side names the writer established by the preceding determination, so the chain validates from the rung-2 `**Writer:**` line forward. No new field, no new record, no spec change.

---

### Finding 2 — a new comment's writer, stated two ways

`class: tech` · non-blocking · `docs/agent-team-protocol.md:464-465` vs `:493-496`

`:464-465` — "Publishing a **new** comment creates a new artifact whose writer is its author" (R2, and spec 0115's *additional logbook comment stays unrestricted* scenario, whose `Then` is "the peer is the writer of the comment it has just created").

`:495-496` — "a published comment has no determinable writer and is written by nobody".

Both are defensible under different senses of "writer" — answerability at creation versus determinability afterwards — but Rule 6 never distinguishes them, and they sit twenty-eight lines apart in one rule. A reader asking "am I the writer of the comment I just posted?" gets two answers.

Behaviourally harmless: every path leads to "correct a published comment by posting a new one" (`:497-498`), which is unambiguous and matches `docs/plan-format.md` → *Append-only revisions* (verified present, `docs/plan-format.md:120`). Raised because it rides the same iteration Finding 1 forces, and because a normative document that answers its own question twice invites the wrong one to be quoted. A half-sentence at `:495` scoping the claim to *editing* would settle it.

---

### On Rule 6's length — 90 lines, and my judgment

Measured, not taken from the PR body: Rule 5 spans `:427-460` = **34 lines**; Rule 6 spans `:462-551` = **90 lines**. The doc's additions (92) reconcile as 90 + 1 blank separator + 1 rewritten preamble line.

**Nothing was lost in the 119 → 90 compression.** The requirement table above is the evidence — every obligation from R2 through R14 traces to a shipped paragraph, including the clauses most at risk in a cut: R5's three refused limbs, R6's five non-transfer vectors, R7's second half, and delta *Out of scope* 6. What was thinned is rationale, and it remains in the spec, the delta, and the plan.

**90 lines is defensible, and I would not ask for another cut.** Rule 6 carries thirteen requirements against Rule 5's one; per obligation it is *tighter* than Rule 5, not looser. Eleven bolded leads make it navigable — an agent needing the handover rule jumps to `:517` without reading the rest, which is the property that matters for a document consulted mid-task rather than read through. Pushing toward Rule 5's 34 would mean dropping obligations, which is what the compression instruction forbade. The one structural improvement I would take is Finding 1's, which *adds* a clause.

---

## Convention compliance

**Gitmoji** — title `✨ Give every forge artifact a single writer at a time`; commit `acfd631` matches. No conventional-commit prefix.

**Version Bump Convention** — both modified skill sources bump `1.1.7` → `1.2.0`. MINOR is right: additive sections, no obligation withdrawn. `scripts/check-skill-versions.sh` exits 0 and names both sources OK.

**Built components** — all eight generated files are staged in the same commit. `bash scripts/build-components.sh --target all --check` exits 0: *"OK: assembly verification passed. OK: All generated files match source."* This is a regeneration-and-diff, so it is conclusive that none of the eight was hand-edited. Parity is symmetric across all four CLI surfaces — `.claude`, `.gemini`, `.github`, `.agents` — 2 skills × 4 CLIs.

**PR body format** — *How to read this PR?*, *How to test this PR?*, and *Detailed description (for agents)* all present. `Closes #735` present and correct; `gh pr view --json closingIssuesReferences` would resolve to the feature issue, which is the intended target here (not a spec-PR).

**Branch naming** — `feat/0735-forge-artifact-single-writer`, matching `<prefix>/<NNNN>-<slug>`.

**`docs/index.json`** — `scripts/build-docs-index.sh --check` exits 0, no drift.

### `docs/cli-matrix.md` — judged, not accepted

The PR argues consultation-without-change. Checked against `docs/cli-matrix-maintenance.md` rather than taken on faith:

- **Trigger surface is hit.** `artifacts/**`, `.claude/**`, `.gemini/**` all touched. The PR body concedes this.
- **The *Protocol-only exemption* does not apply**, and the PR body says so. It is limited to diffs whose only trigger-surface file is a top-level entry point (`AGENTS.md`, `CLAUDE.md`, `GEMINI.md`). Correct self-assessment.
- **No cell moves.** Row 3 (`docs/cli-matrix.md:26`) covers *Skill definitions directory (built output)* generically, with `core` → `.{claude,gemini,github,agents}/skills/<name>/SKILL.md` already enumerated for all four CLIs. This diff adds prose to two existing components and introduces no path, command, hook, or per-CLI configuration.
- **No parity gap arises**, so the *Gap-acceptance evidence rule* is not engaged — that rule governs adding a `Parity gaps` entry, and none is added because all four CLIs receive identical content in this same PR. That is the *Parity check* default ("implement symmetrically in the same PR"), satisfied.
- **The four precedents are real.** Verified by `git show --name-only` in this worktree: `7d56c91` (#646), `bf785d8` (#649), `df22b0b` (#634), `c51e938` (#652) each touch `artifacts/core/skills/*/SKILL.md` with zero `docs/cli-matrix.md` hunks. `088af38` (#648) likewise landed Rule 5 in this same document.

The obligation's text is "consult […] and update it in the same diff", and read at its most literal an update is unconditional. But the clause enumerates what an update *is* — "new row, edited cell, or refreshed `Parity gaps` entry" — and none of the three has a referent here; the section's own stated purpose is that "Drift is a parity bug", and there is no drift when the matrix already describes the surface correctly and all four CLIs move together. **The argument holds. No finding.**

---

## Test coverage

No new automated check ships, and none is owed: spec R12 forecloses mechanical enforcement, and the *Out of scope* list rejects any lock, lease, mutex, or CAS. The obligation is behavioural, verified by reading — the same posture as spec 0095 requirement 6.

The four existing jobs cover every mechanical consequence of this diff, and all four are green: `check-components` (build fidelity), `check-skill-versions` (version bump), `check-agents-size` (`AGENTS.md` untouched, so unchanged), `lint-markdown`.

I re-ran the PR body's six verification commands independently. All reproduce as documented.

**Coverage gap worth naming, not a finding:** nothing mechanically prevents Finding 1's defect class — a ladder whose rungs under-implement a requirement is invisible to every check in the repo, because R12 chose prose over enforcement. That is the accepted trade of this spec, not a defect of this PR. It does mean cold reading is the only gate, which is why Finding 1 is blocking rather than deferred.

---

## Linter findings

| Script | Targets | Result |
|---|---|---|
| `lint-markdown.sh` | all 11 changed `.md` | exit 0, no output |
| `lint-skill.sh` | 2 skill sources, `BASE_REF=origin/main` | exit 0, no output |

`lint-shell.sh`, `lint-python.sh`, `lint-json.sh` — not applicable, no `.sh` / `.py` / `.json` in the diff.

### Byte-level hygiene — checked directly, not via a green lint

The brief is right that a green spec-linter proves nothing here: it validates structure, not stray markup, and passed at exit 0 earlier tonight on a file carrying leaked `</content>` / `</invoke>` scaffolding. So this was run at byte level over all eleven files, whole-file rather than added-lines-only:

- **Tool scaffolding** — `grep -nE '</?(content|invoke|parameter|function_calls|antml|thinking|system-reminder)'` → no match in any of the eleven.
- **Control characters, BOM, zero-width, non-breaking space** — `LC_ALL=C grep` for `[\x01-\x08\x0B\x0C\x0E-\x1F\x7F]`, `\xEF\xBB\xBF`, `\xE2\x80\x8B`, `\xE2\x80\x8E`, `\xC2\xA0` → no match.
- **Trailing whitespace / tabs at EOL** → no match. **CRLF** → 0 carriage returns per file, all eleven.

One methodology note, since it nearly produced a false clean: the first pass used `for f in $FILES`, which under `zsh` does **not** word-split, so `grep` received all eleven paths concatenated as one argument and reported nothing. It was re-run with `while IFS= read -r f` before the result above was trusted. A scan that finds nothing is worth exactly as much as the proof it was actually looking.

---

## Summary

| # | Class | Blocking | Finding |
|---|---|---|---|
| 1 | `tech` | **yes** | Ladder rung 1 (`:479-480`) accepts any recent handover line, so a void nomination resolves to the nominee — defeats R14 and fails delta-01 Scenario 5 |
| 2 | `tech` | no | A new comment's writer is stated two ways (`:464-465` vs `:495-496`) without reconciliation |

Both route to **DEV**. Both are text edits inside `docs/agent-team-protocol.md`; neither touches a skill source, so no further version bump and no rebuild follow from them. Finding 1's fix adds a clause to rung 1 rather than removing one, so Rule 6's length is not a constraint on it.

The rest of the change is sound, and the parts most likely to be wrong are right: the diff followed the merged delta over the plan on R13, both halves of R7 and R9's fallback landed in the role contracts, the eight build outputs are regenerated rather than hand-edited, and the `cli-matrix` argument survives being checked against its own governing document.

---

*Reviewed cold at head `acfd631`. PR body observed at `updatedAt` `2026-08-09T08:50:23Z`; the body is `pr-logbook`'s artifact, not mine — this verdict comment is a new artifact and the only thing this review writes.*

hcross:
## Verdict: APPROVE

Iteration 2, cold re-review at head `911a059`. One file moved since `acfd631` (`docs/agent-team-protocol.md`, +101/−1 against the previous +92/−1); the ten skill files are byte-identical to iteration 1, so build fidelity, version bumps, byte hygiene on those files, and the `cli-matrix` argument stand as upheld and are not re-derived here.

**Both iteration-1 findings are answered.** Finding 1's fix is correct and, more importantly, correct in the way that matters: it does not over-tighten. Finding 2 is properly reconciled.

One finding survives, non-blocking, and it is two words wide. Under `AGENTS.md` → *Retroactive review loop* the pass still does not terminate — termination needs APPROVE **and** zero findings of any class — so this routes to DEV and the next pass should close the loop.

---

## CI status

All **12** checks pass on `911a059`. `check-components` was in progress at review start and finished **pass** (5m48s, job `93224335438`), against the ~5m45s baseline.

`build` · `check-agents-size` · `check-ci-parity` · `check-components` · `check-extension-version-bump` · `check-feedback-routing` · `check-skill-versions` · `check-spec-id-reserved` · `gitlab-ci-check` · `lint-markdown` · `lint-specs` · `test-harness-curate` — pass. No failures, none pending.

CI-green on the reviewed head is satisfied.

---

## 1. Does the forward walk satisfy delta-01 Scenario 5?

Walked independently against the shipped text at `:480-490`, not taken from the author's account.

`pr-logbook` creates the PR body, so the logbook carries `**Writer:** pr-logbook — PR #N body`. `team-lead`, not the writer, states on the record that writership now belongs to `architect` — the scenario's premise, rendered in the nominator's own voice as `**Writership handover:** PR #N body — team-lead → architect`. `architect` is about to write and resolves:

- **Step 1** (`:480-482`) — start at `pr-logbook`. "That role is the writer until a valid handover moves it."
- **Step 2** (`:483-488`) — apply handovers oldest first. The single line's left-hand role is `team-lead`; the walk has reached `pr-logbook`. Left-hand ≠ walk position, so the line "is void …, contributes nothing, and does not halt the walk". Nothing moves.
- Walk ends at `pr-logbook`.

Against the scenario's three `Then` clauses:

| Scenario requires | Shipped text yields |
|---|---|
| "finds the first agent's role named" | `pr-logbook` ✓ |
| "does not treat the second agent's statement as a handover" | void, contributes nothing ✓ |
| "it does not write the body" | determination does not name `architect`'s role, and `:530-532` bars it ✓ |

**Scenario 5 passes.** The inversion is the right fix, and the author's reasoning for inverting rather than patching the old rung 1 holds up: "chain-validating against the preceding determination" is circular unless the chain has an origin, and creation is the only origin R3 supplies.

---

## 2. Has validation over-tightened?

This was the risk worth the most scrutiny — a rule that silently rejects a *legitimate* handover leaves the artifact unwritable by anyone, and nothing would surface it. Four walks:

| Case | Walk | Result |
|---|---|---|
| Spec 0115 explicit-handover scenario — `pr-logbook` hands to a peer | start `pr-logbook`; line `pr-logbook → team-lead`, left-hand = position → moves | `team-lead` is writer at the moment of the write ✓; the previous writer determining now finds `team-lead`, not itself, so it "does not write the body again without a handover back" ✓ |
| Chained handover | `pr-logbook` → `team-lead` → `architect`, each left-hand matching the walk position in turn | `architect` ✓ |
| Handover back | `pr-logbook → team-lead`, then `team-lead → pr-logbook` | `pr-logbook` ✓ |
| **Void line interleaved with valid ones** | `pr-logbook → team-lead` (valid, walk at `team-lead`); `architect → developer` (void, skipped); `team-lead → developer` (valid) | `developer` ✓ |

The fourth is the one that decides it, and the text handles it explicitly: a void line "contributes nothing, and **does not halt the walk**: the most recent line is not the answer unless it survives" (`:487-488`). A stray nomination dropped mid-chain therefore cannot poison a later legitimate handover. Without that clause the fix would have been worse than the defect. It is there.

**No over-tightening.** A handover is rejected only when its left-hand role is not the current writer — which is precisely the definition of not being a legitimate handover. There is no input where a valid handover is refused.

One adjacent case I checked and am **not** raising: if the creating role never posts a `**Writer:**` line, step 1 has no anchor, no handover can match, and the artifact is unwritable by everyone. That is stricter than the old rung 1, which could have taken a bare handover line as an origin. But a handover with no prior `**Writer:**` line is a transfer of writership nobody demonstrably held — exactly what R14 voids — so tightening there is correct. And the "nobody can determine it" outcome is R4's own mandated default, not something this fix introduced.

---

## 3. The residual — a forged line in the writer's voice

Assessed on the merits, as asked, rather than accepted because it was disclosed.

The walk rejects `team-lead → architect` (nominator's own voice) but accepts `pr-logbook → architect` posted by `team-lead` — a line asserting a handover `pr-logbook` never performed. Left-hand matches the walk position, so it applies.

**Not a finding.** Three reasons, in order of weight:

1. **The document closes it by stipulation, in the shipped text and not merely in the PR body.** `:485-486` reads "its left-hand role is the writer the walk has reached, **that role being the one who posts it**". The left-hand position is *defined* as the poster. A line whose left-hand names a role that did not post it is therefore not a permitted-but-unvalidated record — it is a false statement about who acted, which spec 0115's third scenario already classifies as auditable misconduct ("the audit does not require the lost content to be recoverable to establish the violation").
2. **No procedure available to the protocol closes it.** Distinguishing a genuine `pr-logbook → architect` from a forged one requires authorship attribution at role granularity. Rule 6 itself records why the forge cannot supply it (`:492-493`: "the forge's own authorship field names the *account*, not the role"), and under a shared maintainer account that is structural, not incidental. R9 additionally forbids depending on a capability only one forge offers.
3. **R12 already fixes the posture.** The spec forecloses locks, leases, and mechanical enforcement; the whole rule is honour-system by design. An agent that forges a handover line has departed from the protocol no less than one that simply overwrites the body without looking — a case Rule 6 equally cannot prevent and was never asked to.

Raising this would produce a `spec`-class finding whose only available remedy is "and do not lie on the record", which R14 already says. That is not worth a SPECS iteration.

---

## 4. Prose integrity — read end to end

Rule 6 read as prose, `:462-560`, plus mechanical checks for the defect class that slipped twice during authoring. No dropped noun, no duplicated clause, no orphaned referent: adjacent-duplicate-word scan over the rule returns empty, and no three-word sequence repeats more than twice. Byte-level re-scan of the changed file is clean — no tool scaffolding, no control characters, BOM, zero-width or bidi marks, no non-breaking spaces, no trailing whitespace, zero carriage returns. `lint-markdown.sh` exits 0.

Two spots I checked specifically and cleared:

- `:492-493` "**Both lines** ride in comments the ticket already gets" — this sentence was lifted out of the old rung 2 and now stands alone after the list. Its referent still resolves (the `**Writer:**` line of step 1 and the `**Writership handover:**` line of step 2), and the fenced example directly below shows both. Fine.
- `:500-507` — Finding 2's reconciliation. "A comment's writer is its author when it is published, but nothing records that, so every later determination resolves at [step] 3 and the comment is written by nobody from then on." The two propositions are now one temporal sequence rather than two competing assertions, and it agrees with `:464-465`. **Finding 2 is closed.**

### Finding — `rung` outlived the ladder it belonged to

`class: tech` · non-blocking · `docs/agent-team-protocol.md:500` and `:503`

This commit renamed the determination procedure from a ladder to a walk, thoroughly: `ladder` now appears **zero** times in the file, `walk` appears five, and two of those five are deliberate rewrites of former `ladder` occurrences (`:522`, `:531`, both in the incremental diff). Two references to the old metaphor were left behind:

- `:500` — "**Already-published comments resolve at rung 3, by design.**" (a bolded paragraph lead, the most-scanned text in the rule)
- `:503` — "so every later determination resolves at rung 3"

A *rung* is a step of a ladder. With the ladder gone from the document's vocabulary, a reader meets a word the rule never introduces and has to map it onto a numbered list that is now described as a walk. At `acfd631` this was internally consistent (`ladder` ×2, `rung` ×2); the inconsistency is introduced by this commit, so it is a regression in the diff under review rather than pre-existing debt.

Nothing behavioural turns on it — the pointer still resolves, and "step 3" is the correct target under the new structure exactly as it was under the old. `markdownlint` cannot see it, which is why it is here.

**Fix:** `rung 3` → `step 3` at both sites. Two words, one file, no skill source touched, so no version bump and no rebuild follow.

For the avoidance of doubt on scope: the `rung` at `artifacts/core/skills/pr-reviewer/SKILL.md:130` and at `docs/solo-merge-classifier-workaround.md:29` belong to the *verdict-posting* fallback ladder, which is a different ladder and still called one. Both are correctly untouched.

---

## 5. Rule 6 at 99 lines

Measured at head: Rule 6 spans `:462-560` = **99 lines**; Rule 5 spans `:427-460` = **34**. The nine added lines reconcile exactly against the two fixes — seven for the forward walk and its void-line clause, two for the comment sequence.

**Still defensible; I am not flagging it.** I said at 90 not to cut further, and I would not trade either of these nine-line purchases back. The seven bought the difference between a rule that fails delta-01 Scenario 5 and one that passes it; the two closed a normative document answering its own question twice. A shorter Rule 6 with those defects would be the worse artifact.

Nothing was padded back: the four trimmed spans stay trimmed — the derivation of role-scoping, the authorship-field measurement, the per-comment cost rationale, and the PR #734 re-narration are all still absent, verified by reading rather than taken from the author's statement. The structure that makes the length tolerable is intact — **ten** bolded paragraph leads, one per obligation, thirteen counting the three inside the walk's numbered list, still jump-navigable mid-task. Rule 6 carries thirteen requirements to Rule 5's one; at 99 lines it remains the tighter of the two per obligation.

*Correction to my iteration-1 review:* I wrote "eleven bolded leads" there. That figure was inherited from the PR body and is wrong — the count is ten paragraph leads (a naive `grep -cE '^\*\*[A-Z]'` returns twelve because it also matches the two `**Writer:**` / `**Writership handover:**` lines inside the fenced example at `:496-497`). Nothing turns on the number; correcting it because I asserted it without deriving it, which is the one thing the grounding rule asks a reviewer not to do.

---

## Summary

| # | Class | Blocking | Finding |
|---|---|---|---|
| 1 | `tech` | no | `rung 3` at `:500` and `:503` survives a rename that purged `ladder` from the file — orphaned metaphor in a bolded lead |

Iteration-1 Finding 1 (**closed**) — the forward walk satisfies delta-01 Scenario 5, verified by independent walk, and does not reject any legitimate handover across four cases including a void line interleaved mid-chain.
Iteration-1 Finding 2 (**closed**) — the comment's writer is now one temporal sequence, consistent with `:464-465`.

Verdict is **APPROVE**: nothing blocking remains, and the change is sound. Under the lifecycle's termination criterion the pass still surfaces one finding, so it routes to DEV rather than to MERGE. A two-word edit should make iteration 3 terminate.

---

*Reviewed cold at head `911a059`, file read from the forge at that ref rather than from a local checkout. PR body observed at `updatedAt` `2026-08-09T09:10:18Z`; the body is `pr-logbook`'s artifact — this verdict comment is a new artifact and the only thing this review writes.*

hcross:
## Verdict: APPROVE

Iteration 3, terminal check at head `67ccd76`. **Zero findings.**

The diff since `911a059` is two words in one file, so this review is scoped to match it. Everything settled across iterations 1 and 2 — requirement coverage against both spec files, the forward walk's correctness, the four over-tightening walks, build fidelity, version bumps, byte hygiene, the `cli-matrix` argument, the forged-line residual, and the 99-line judgement — stands, and a two-word substitution cannot disturb any of it.

### The change is confined

`git diff 911a059..67ccd76` → `1 file changed, 2 insertions(+), 2 deletions(-)`, both in `docs/agent-team-protocol.md`, both the substitution asked for. Nothing else moved.

### The two paragraphs read correctly

This was the one thing worth attention here, since prose damage during a small edit has bitten this ticket three times and no CI check sees it. Read at head:

> **Already-published comments resolve at step 3, by design.** A `**Writer:**` line is posted for a pull-request or issue body, never per comment. A comment's writer is its author when it is published, but nothing records that, so every later determination resolves at step 3 and the comment is written by nobody from then on — effectively append-only […]

Both substitutions are noun-phrase-for-noun-phrase in identical grammatical position, and both sentences carry through cleanly — no dropped noun, no duplicated clause, no broken agreement. The adjacent-duplicate-word scan over Rule 6 returns empty and `lint-markdown` exits 0.

The pointer is also now correct in the document's own vocabulary, not merely consistent: the list is introduced at `:476-478` as a forward walk, and its third element at `:489-490` is "**Otherwise, you are not the writer.**" — exactly what a comment resolves to, having no `**Writer:**` line for step 1 to anchor on and no handovers for step 2 to apply. `step 3` names it in the same terms the list uses.

`rung` and `ladder` both now return **0** occurrences in the file. The `step 3` at `:136` is pre-existing and belongs to *Worktree Isolation*, unrelated to Rule 6 and untouched. The two verdict-posting-ladder occurrences outside this file remain correctly scoped out, as they describe a different ladder that is still called one.

### CI

Observed directly on `67ccd76`: **12 of 12 pass, zero failures, none pending.** PR reports `MERGEABLE` / `CLEAN`, 11 files.

---

## Termination

| Criterion | State |
|---|---|
| REVIEW verdict is APPROVE | ✅ |
| Zero findings of any class | ✅ |
| CI green on the reviewed head | ✅ 12/12 on `67ccd76` |

All three conditions of `AGENTS.md` → *Retroactive review loop* are met. The lifecycle terminates at MERGE on this head. Three iterations, inside the guardrail of five.

Both findings this review raised were real and both were answered without collateral: iteration 1's blocking defect (the ladder resolving a void nomination to the nominee) produced the forward walk that satisfies delta-01 Scenario 5 while still admitting every legitimate handover, and iteration 2's orphaned metaphor is now gone. Spec 0115 and its delta-01 are realized, and the document that carries them is internally consistent.

---

*Reviewed cold at head `67ccd76`, file read from the forge at that ref rather than from a local checkout. PR body observed at `updatedAt` `2026-08-09T09:20:13Z`; the body is `pr-logbook`'s artifact — this verdict comment is a new artifact and the only thing this review writes.*

PR #780, opened 2026-08-09
🐛 Run git at the repository root so the base-branch check cannot silently pass

### LINKED ISSUE #768: 🐛 The base-branch spec check silently passes when run from a subdirectory
> **Corrected 2026-08-09, before any implementation.** This issue was first filed
> claiming the hazard was *unreachable* ("correct only because markdownlint fails
> first"). **That claim is false**, and it was falsified by measurement rather than
> by review. The reachability probe is below. This is a live, silent wrong-green,
> not hardening against a hypothetical.

## The defect

The base-branch `status: draft` check derives two sets from git, and both are
resolved **relative to the process working directory**:

- `baseBranchPaths()` (`scripts/lib/spec-linter.js:155-164`) passes
  repo-root-relative paths as pathspecs to `git ls-tree`, which interprets a
  pathspec **relative to the current directory** unless `--full-name` is given.
  From a subdirectory it matches nothing and returns an empty set — so the check
  finds **no offenders at all** and reports success.
- `changedPaths()` (`spec-linter.js:166-205`, added by #732) runs
  `git diff --name-only <merge-base>`, which is root-relative under default config
  but becomes **cwd-relative, silently omitting files outside cwd**, when
  `diff.relative=true`.

## Reachability, measured

One throwaway repository, one spec present on the base branch carrying
`status: draft`, under `sub/specs/`. `sub/` carries its own `.markdownlintrc`, so
markdownlint finds a config and does **not** abort first — which is the only thing
that was hiding this.

```console
$ cd "$FIX/sub" && BASE_REF=main node …/scripts/lib/spec-linter.js specs
Running markdownlint-cli on 1 files...
Running semantic validation...

Linting passed!
[exit 0]                       # ← the violation is on the base branch and is NOT reported

$ cd "$FIX"     && BASE_REF=main node …/scripts/lib/spec-linter.js sub/specs
  A spec reaches the base branch only by its own merged spec-PR, which is
  the trigger docs/spec-format.md assigns to 'approved'. Record the status
  that reflects the spec's true state (metadata-only edit).

Linting failed: 1 files contain errors.
[exit 1]                       # ← same repo, same spec, same base ref
```

Nothing exotic is required: no unusual git config, no code change, no CI change.
A nested `.markdownlintrc` beside a nested `specs/` directory is enough — and
nested spec directories are an established pattern in this repository
(`specs/org/`, per `specs/0071-org-specs-lint-exclusion.md`).

## Why this is the worst class of defect for this particular check

`specs/0109-spec-status-invariant-on-main.md` exists to remove the "green does not
mean checked" failure mode, and its own implementation reproduces it. The check
does not error, warn, or skip with a notice — the `[SKIP]` path
(`spec-linter.js:~215`) is reserved for "not inside a git work tree" and is not
reached here. It prints `Linting passed!` and exits `0`, which is
indistinguishable from a clean corpus.

## Suggested fix

Pin git to the repository root **structurally**, so that a git call added to this
file later inherits the correct behaviour instead of requiring its author to
notice path-sensitivity. A module-level root, set once the work tree is located,
used as `gitCapture`'s default `cwd`:

```js
let gitCwd = null;                    // set once `rev-parse --show-toplevel` succeeds

function gitCapture(args, cwd) {
    const result = spawnSync('git', args, {
        encoding: 'utf8',
        cwd: cwd !== undefined ? cwd : (gitCwd || undefined),
    });
    …
}
```

The discovery call (`rev-parse --show-toplevel`) must stay unpinned — it is what
answers "which repository am I in".

Rejected alternative: per-call flags (`git ls-tree --full-name`,
`git diff --no-relative`). They state the intent at the point of use but leave the
next author to remember them, which is the fragility that produced this bug;
`--no-relative` also introduces a git ≥ 2.28 floor.

Rejected alternative: assert `cwd == repoRoot` and fail loudly. It eliminates the
silence but codifies a constraint instead of removing one, and leaves the check's
correctness dependent on another subsystem's behaviour — the property this issue
is about.

## Coverage

Reachable through the CLI, so it needs **no module exports**: an ordinary
`run_base_case` in `scripts/tests/test-spec-linter.sh` with a fixture whose specs
and `.markdownlintrc` sit under a subdirectory, run from that subdirectory,
expecting the offender to be named. That case exits `0` before the fix and `1`
after. (The original filing claimed exports were the real cost of covering this.
Also wrong — the probe above is the counter-example.)

Related: #732, `specs/0109-spec-status-invariant-on-main.delta-02.md`.

hcross:
The base-branch `status: draft` check silently passed when the linter ran from a subdirectory: `git ls-tree` resolved repo-root-relative pathspecs against the working directory, matched nothing, and no spec was ever identified as present on the base branch. Pinning git to the repository root — structurally, not per call site — makes the check trustworthy wherever it runs.

Implements `specs/0109-spec-status-invariant-on-main.delta-03.md` (R14–R16). Closes #768.

## How to read this PR?

Two files, and the interesting part is *why the fix has the shape it has* rather than what it does.

1. **`scripts/lib/spec-linter.js`** — a module-level `gitCwd`, assigned once in `resolveBaseContext()` right after the work tree is located, used as `gitCapture`'s default `cwd`. Three lines of behaviour; the comment above `gitCwd` carries the argument.
2. **`scripts/tests/test-spec-linter.sh`** — cases 39 and 40 on a new `$GITFIX4` whose specs **and** `.markdownlintrc` both sit under `sub/`.

Three things worth pausing on:

- **Why a default and not an argument at each call site.** `--full-name` on `ls-tree` and `--no-relative` on `diff` would satisfy R14 for the two invocations that carry them. R15 rejects that, on evidence rather than taste: this defect exists *because* two path-sensitive invocations were each written without their author classifying them as path-sensitive. A remedy requiring that classification to be made correctly next time does not remove the failure mode. As a default, invocation number eight inherits the guarantee without knowing it exists. (`--no-relative` also imposes a git ≥ 2.28 floor, and an unrecognised flag is a hard error, not a degradation.)
- **The one invocation that must stay unpinned.** `rev-parse --show-toplevel` runs while `gitCwd` is still `null`, so it resolves against the caller — which is what lets it answer *which repository* the run is about. Pinning it would be circular. This falls out of assignment order rather than needing an exemption anyone maintains.
- **The nested `.markdownlintrc` in the fixture is load-bearing, not incidental.** `-c .markdownlintrc` is resolved against the working directory, so a subdirectory run without a config there dies at markdownlint and never reaches the check. That is precisely why I first filed #768 claiming the hazard was unreachable — the reasoning assumed the only config in existence is the root one. The fixture exists to keep the path reachable.

## How to test this PR?

```sh
ln -s /path/to/main/checkout/node_modules node_modules   # or npm install
bash scripts/tests/test-spec-linter.sh                   # expect: 41 passed, 0 failed
BASE_REF=crewrig/main node scripts/lib/spec-linter.js    # whole corpus
```

**The probe from the issue, before and after.** One throwaway repository, one non-delta spec on the base branch at `status: draft`, under `sub/specs/`, with `sub/.markdownlintrc` present:

```console
$ cd "$FIX/sub" && BASE_REF=main node <main-checkout>/scripts/lib/spec-linter.js specs
Linting passed!                                    [exit 0]   ← the violation is NOT reported

$ cd "$FIX/sub" && BASE_REF=main node <this-branch>/scripts/lib/spec-linter.js specs
  A spec reaches the base branch only by its own merged spec-PR, …
Linting failed: 1 files contain errors.            [exit 1]   ← named
```

**Falsification.** Both new cases were shown to fail when the behaviour they cover is removed, and the two mutations discriminate between a complete and a partial fix:

| Mutation | Case 39 | Case 40 |
|---|---|---|
| `gitCwd = repoRoot` assignment removed | **FAIL** (exit 0) | **FAIL** (exit 0) |
| `ls-tree` pinned, `git diff` left unpinned (`process.cwd()`) | pass | **FAIL** (exit 0) |

The second row is why Case 40 exists: with `ls-tree` pinned the offender *is* identified, but an unpinned `git diff` reports `specs/0240-nested-layout.md` relative to `sub/`, so it never matches `sub/specs/0240-nested-layout.md`, the offender degrades to a bystander, and the run passes with a `[WARN]`. Hence Case 40 asserts on the **absence** of `[WARN]`, not on the exit code alone.

Mutations were run under `env -i bash --noprofile --norc` with `/bin/cp -f` restores verified by `diff -q`. Not ceremony: in the sibling ticket #732 an aliased `cp -i` silently no-opped three restores while exiting `0`, so three different mutations were measured against the same file and produced identical results.

### `scripts/lib/spec-linter.js`

**Added `let gitCwd = null;`** above `gitCapture`, with the invariant stated once: git in this file runs at the repository root.

**`gitCapture(args)` → `gitCapture(args, cwd)`.** Resolution order is explicit argument → `gitCwd` → `undefined`. The `undefined` rung preserves today's behaviour exactly for every pre-work-tree and non-repository path — Case 33 (outside a git work tree) and the whole flat `$TMP_ROOT` fixture set exercise it, and all of them still pass unchanged.

**Assigned `gitCwd = repoRoot` in `resolveBaseContext()`**, immediately after `fs.realpathSync(toplevel…)` and before the first path-sensitive read. That single line covers all seven non-discovery invocations — `resolveBaseRef()`'s four (`remote`, two `rev-parse --verify`, `fetch`), `baseBranchPaths()`'s `ls-tree`, and `changedPaths()`'s `merge-base` and `diff` — without naming any of them. `repoRoot` is already canonicalised via `realpathSync` for the symlinked-checkout case, and pinning to the canonical form is what keeps it consistent with the `relPathByFile` comparison built on the same form.

Only two of those seven were path-sensitive; the other five are pinned as a side effect and are unaffected (`git fetch` writes to the repository rather than reading paths — the only *writing* call, called out because it is the one where an unexamined cwd change could plausibly matter, and Cases 28–38 exercise the resolution path it lives in).

### `scripts/tests/test-spec-linter.sh`

**`$GITFIX4`** — `sub/specs/0240-nested-layout.md` at `status: draft` committed on `main`, with `sub/.markdownlintrc` and a `sub/node_modules` symlink. No remote; both cases pass `BASE_REF` explicitly.

- **Case 39** (R14) — run with `sub/` as the working directory, target `specs`, expect exit 1 with the offender named. This is the defect, reachable through the ordinary CLI entry point.
- **Case 40** (R14, attribution half) — same run with `diff.relative=true` and the offending spec modified in the tree, expect exit 1 and no `[WARN]`. Discriminates the partial fix, as tabulated above. Its comment states plainly that it does **not** isolate the `diff` derivation from the `ls-tree` one — the only working directory where `diff` misbehaves is one where `ls-tree` has already returned nothing, so isolation is not observable through the command line. Recorded so the case is not read as stronger than it is.

Suite: **41 passed, 0 failed**.

### Not triggered, checked rather than assumed

- **Docs:** none. Unlike delta-02's R12, delta-03 imposes no documentation obligation — nothing a spec author or reader does changes, so `docs/spec-format.md` is untouched.
- **CLI matrix:** `scripts/lib/` and `scripts/tests/` are not trigger surfaces. No `docs/cli-matrix.md` edit due; `check-ci-parity` and `gitlab-ci-check` unaffected.
- **CI wiring:** unchanged in both engines. No new variable, no new parity surface.
- **Version bump:** no skill or agent source touched. The spec-side bump (`2.1.0`) shipped in #774.
- **Build outputs / core-paths manifest:** none; nothing under `artifacts/`, no layer reclassified.
- **Behaviour change visible to a user:** the linter starts *reporting* offenders it previously missed. No repository-nested spec layout exists on `main` today (all specs sit at `specs/`, and `specs/org/` is manifest-excluded), so nothing here changes state — confirmed by the whole-corpus run passing.

### DIFF
--- a/scripts/lib/spec-linter.js
+++ b/scripts/lib/spec-linter.js
@@ -100,11 +100,39 @@ function isExcludedSpecPath(relPath, entries) {
 // legitimately `draft` until its own merge mechanic flips it, while a spec
 // already on the base branch carrying `draft` is the contradiction R1 forbids.
 
-// gitCapture(args) — run git, capturing both streams. Returns the exit status
-// and stdout; never throws and never leaks git's own stderr into the linter's
-// output, so callers can probe git freely (including outside a repository).
-function gitCapture(args) {
-    const result = spawnSync('git', args, { encoding: 'utf8' });
+// gitCwd — the directory every git invocation in this file runs in, assigned once
+// the work tree has been located (see resolveBaseContext). It enforces one
+// invariant: **git in this file runs at the repository root** (spec 0109
+// delta-03 R14).
+//
+// A default rather than an argument at each call site, because R15 requires the
+// mechanism to be structural. The defect delta-03 fixes existed precisely
+// because two path-sensitive invocations — `ls-tree`'s pathspecs and `diff`'s
+// output paths — were each written without their author classifying them as
+// path-sensitive; from a subdirectory the first returned an empty set and the
+// check reported `Linting passed!` on a live violation. A remedy that depends on
+// that classification being made correctly next time does not remove the failure
+// mode. As a default, an invocation added later inherits the guarantee without
+// having to know it exists.
+//
+// One invocation is necessarily exempt: the one that DISCOVERS the work tree runs
+// while this is still `null`, so it resolves against the caller's directory —
+// which is what lets it answer which repository the run is about. Pinning it
+// would be circular.
+let gitCwd = null;
+
+// gitCapture(args, cwd) — run git, capturing both streams. Returns the exit
+// status and stdout; never throws and never leaks git's own stderr into the
+// linter's output, so callers can probe git freely (including outside a
+// repository). `cwd` overrides the module default for a single call; omitting it
+// (the normal case) takes `gitCwd`, and `undefined` inherits this process's
+// directory — which is what every pre-work-tree and non-repository path does,
+// exactly as before this file had a default at all.
+function gitCapture(args, cwd) {
+    const result = spawnSync('git', args, {
+        encoding: 'utf8',
+        cwd: cwd !== undefined ? cwd : (gitCwd !== null ? gitCwd : undefined),
+    });
     return {
         status: result.status,
         stdout: typeof result.stdout === 'string' ? result.stdout : '',
@@ -247,6 +275,13 @@ function resolveBaseContext(files) {
     // (a symlinked checkout, /tmp on macOS) would otherwise resolve "outside"
     // the repository and be silently exempted from the check.
     const repoRoot = fs.realpathSync(toplevel.stdout.trim());
+
+    // From here on, every git invocation runs at the repository root (R14/R15).
+    // Assigned immediately after the root is known and before the first
+    // path-sensitive read, so resolveBaseRef(), baseBranchPaths() and
+    // changedPaths() are all covered without this line naming any of them —
+    // which is the point: the next one is covered too.
+    gitCwd = repoRoot;
 
     const { ref, error } = resolveBaseRef();
     if (error) {
--- a/scripts/tests/test-spec-linter.sh
+++ b/scripts/tests/test-spec-linter.sh
@@ -788,6 +788,16 @@ run_base_case "Case 38 — attribution that cannot be derived blocks, and says s
 # -------------------------------------------------------------------------
 # Cases 39-40 (delta-03 R16, covering R14/R15) — WORKING-DIRECTORY INDEPENDENCE.
 #
+# Both cases assert on `Non-delta specs present on the base branch` — the banner
+# only this check emits — and NOT on the offending file's path. The path is also
+# printed by markdownlint on any rule violation in the fixture, and markdownlint's
+# own failure is exit 1, so a fixture that drifts out of conformance would satisfy
+# a path-plus-exit-code assertion while the check never executed. Measured: with
+# an MD001 violation injected and run against a linter WITHOUT the fix, exit is 1
+# and the path appears. For cases guarding a check whose defect was passing
+# without checking, that is the one false-pass that must not be possible. The
+# fixture holds exactly one spec, so asserting the banner loses no identity.
+#
 # A fixture whose specs AND markdownlint configuration both sit under `sub/`.
 # The nested configuration is not incidental: `-c .markdownlintrc` is resolved
 # against the working directory, so without a config there the run dies at
@@ -818,9 +828,11 @@ render_spec "0240" "nested-layout" "draft" > "$GITFIX4/sub/specs/0240-nested-lay
 # reported `Linting passed!` with exit 0 on a live violation. Remove the
 # `gitCwd = repoRoot` assignment and this case goes red.
 # -------------------------------------------------------------------------
+git -C "$GITFIX4" checkout -q -- .
+git -C "$GITFIX4" config --unset-all diff.relative 2>/dev/null || true
 run_base_case "Case 39 — an offender is identified from a subdirectory (R14)" \
   "$GITFIX4/sub" "specs" 1 "main" \
-  "specs/0240-nested-layout.md" "-"
+  "Non-delta specs present on the base branch" "-"
 
 # -------------------------------------------------------------------------
 # Case 40 (R14, the attribution half) — same subdirectory run, with
@@ -839,12 +851,17 @@ run_base_case "Case 39 — an offender is identified from a subdirectory (R14)" 
 # observable through the command line, since the only working directory where
 # `diff` misbehaves is one where `ls-tree` has already returned nothing. Stated
 # rather than implied, so the case is not read as stronger than it is.
+#
+# Like cases 35-37 on $GITFIX2, these two do NOT depend on each other's order:
+# each resets the work tree and the `diff.relative` setting, then applies only
+# what it measures. A third case added here inherits neither.
 # -------------------------------------------------------------------------
+git -C "$GITFIX4" checkout -q -- .
 git -C "$GITFIX4" config diff.relative true
 printf '\nEdited by the change under test.\n' >> "$GITFIX4/sub/specs/0240-nested-layout.md"
 run_base_case "Case 40 — attribution stays root-anchored under diff.relative (R14)" \
   "$GITFIX4/sub" "specs" 1 "main" \
-  "specs/0240-nested-layout.md" "[WARN]"
+  "Non-delta specs present on the base branch" "[WARN]"
 
 # -------------------------------------------------------------------------
 # Summary
--- a/scripts/tests/test-spec-linter.sh
+++ b/scripts/tests/test-spec-linter.sh
@@ -786,6 +786,67 @@ run_base_case "Case 38 — attribution that cannot be derived blocks, and says s
   "could not be derived" "[WARN]"
 
 # -------------------------------------------------------------------------
+# Cases 39-40 (delta-03 R16, covering R14/R15) — WORKING-DIRECTORY INDEPENDENCE.
+#
+# A fixture whose specs AND markdownlint configuration both sit under `sub/`.
+# The nested configuration is not incidental: `-c .markdownlintrc` is resolved
+# against the working directory, so without a config there the run dies at
+# markdownlint and never reaches the check. That is exactly why this defect was
+# first believed unreachable, and it is what these cases exist to keep reachable.
+# -------------------------------------------------------------------------
+GITFIX4="$TMP_ROOT/gitfix4"
+mkdir -p "$GITFIX4/sub/specs"
+cp "$ROOT_DIR/.markdownlintrc" "$GITFIX4/sub/"
+ln -s "$ROOT_DIR/node_modules" "$GITFIX4/sub/node_modules"
+render_spec "0240" "nested-layout" "draft" > "$GITFIX4/sub/specs/0240-nested-layout.md"
+(
+  cd "$GITFIX4" || exit 1
+  git init -q
+  git symbolic-ref HEAD refs/heads/main
+  git config user.email "test@example.com"
+  git config user.name "Test"
+  git config commit.gpgsign false
+  git add sub/specs
+  git commit -q -m "draft non-delta spec on the base branch, under sub/"
+)
+
+# -------------------------------------------------------------------------
+# Case 39 (R14) — the offender is identified when the linter runs from a
+# SUBDIRECTORY. Before delta-03, `git ls-tree` resolved the repo-root-relative
+# pathspec against the working directory, matched nothing, and returned an empty
+# set — so no file was ever identified as present on the base branch and the run
+# reported `Linting passed!` with exit 0 on a live violation. Remove the
+# `gitCwd = repoRoot` assignment and this case goes red.
+# -------------------------------------------------------------------------
+run_base_case "Case 39 — an offender is identified from a subdirectory (R14)" \
+  "$GITFIX4/sub" "specs" 1 "main" \
+  "specs/0240-nested-layout.md" "-"
+
+# -------------------------------------------------------------------------
+# Case 40 (R14, the attribution half) — same subdirectory run, with
+# `diff.relative=true` and the offending spec MODIFIED in the tree. It must fail,
+# because a change that touches the offender is the change that pays
+# (delta-02 R2 as replaced).
+#
+# This is the case that discriminates a PARTIAL fix, which is why it exists
+# alongside Case 39: pin `ls-tree` alone and the offender is identified, but
+# `git diff` still reports `specs/0240-nested-layout.md` relative to `sub/`, the
+# `changed` set never matches `sub/specs/0240-nested-layout.md`, the offender
+# degrades to a bystander, and the run passes with a [WARN]. Hence the assertion
+# on the ABSENCE of [WARN] rather than on the exit code alone.
+#
+# It does not isolate the `diff` derivation from the `ls-tree` one — that is not
+# observable through the command line, since the only working directory where
+# `diff` misbehaves is one where `ls-tree` has already returned nothing. Stated
+# rather than implied, so the case is not read as stronger than it is.
+# -------------------------------------------------------------------------
+git -C "$GITFIX4" config diff.relative true
+printf '\nEdited by the change under test.\n' >> "$GITFIX4/sub/specs/0240-nested-layout.md"
+run_base_case "Case 40 — attribution stays root-anchored under diff.relative (R14)" \
+  "$GITFIX4/sub" "specs" 1 "main" \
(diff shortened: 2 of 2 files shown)


hcross:
## Verdict: REQUEST CHANGES

Two `tech` findings, both in the **new test cases**, both measured rather than
reasoned. The production change is correct and I found nothing against it.

**Reviewer-independence disclosure, first because it qualifies everything below.**
Not cold-spawned: this session runs under an instruction not to spawn subagents, so
the author reviewed the diff. Shared-identity posting path applies (`gh api user`
→ `hcross` = PR author), so this plain comment is the binding verdict per the
skill's rung 2. Treat it as a structured self-audit, not independent confirmation.

### CI status

All **13** required checks **pass** on `020a5e3`. None failing, none pending.
`lint-specs` reports `Results: 41 passed, 0 failed` then `Linting passed!`. Diff is
exactly the two files claimed: `scripts/lib/spec-linter.js` (+40),
`scripts/tests/test-spec-linter.sh` (+61).

### Correctness

- **Assignment order is right, and it is the thing that makes the design work.**
  `spec-linter.js:284` sets `gitCwd = repoRoot` after `realpathSync` and **before**
  `resolveBaseRef()` at line 286, so all four of that function's invocations are
  covered along with `baseBranchPaths()` and `changedPaths()`. Verified by reading
  the region, not inferred from the comment.
- **The discovery invocation stays unpinned by construction**, not by an
  exemption: `rev-parse --show-toplevel` runs while `gitCwd` is `null`, so
  `gitCapture` falls to `undefined` and inherits. Circularity avoided without a
  special case anyone has to maintain.
- **The `undefined` rung preserves today's behaviour exactly** for every
  non-repository path. Case 33 (outside a git work tree) and the whole flat
  `$TMP_ROOT` fixture set exercise it; all still pass.
- **`repoRoot` is the canonicalised form**, consistent with the `relPathByFile`
  comparison built on the same `realpathSync` output. Pinning to the physical path
  is correct here, and the pre-existing comment above it already records why
  canonicalisation matters (symlinked checkout, `/tmp` on macOS).
- **`git fetch` is now pinned too.** It writes to the repository rather than
  reading paths, so cwd is immaterial; called out because it is the only *writing*
  invocation in the file and therefore the one where an unexamined cwd change could
  plausibly matter. Cases 28–38 exercise the resolution path it lives in.
- **The fix is verified against the defect, not only against fixtures.** Same
  probe, same repository, same base ref: `main`'s linter exits 0 with
  `Linting passed!` on a live violation; this branch's exits 1 and names it.

### Convention compliance

- Gitmoji `🐛` on both commit and PR title; no conventional-commit prefix.
- Spec-PR ordering respected: delta-03 merged first (#774, `cde9fc2`), the
  implementation branch cut afterwards.
- Branch `fix/0109-spec-status-invariant-on-main` follows `<prefix>/<NNNN>-<slug>`
  with `0109` the spec id, per `docs/spec-pr-workflow.md` → *Ordering rule*. Note
  for the record: this repeats the branch name used for #732's implementation
  against the same spec. Permitted (the earlier branch is deleted local and remote)
  and traceable, but two implementations of one spec id are now indistinguishable
  by branch name alone. Not a finding — the convention says spec id — but worth
  knowing that sibling sessions use the *issue* number instead
  (`feat/0736-…` for spec 0114), so the repository has two live practices.
- `Closes #768` on the implementation PR only; the spec-PR carried none
  (`closingIssuesReferences` → `[]` on #774).
- No version-bump trigger, no CLI-matrix trigger surface, no build output, no
  core-paths change. Pre-edit guard satisfied: issue existed, feature branch,
  dedicated worktree `.worktrees/768`.

### Test coverage

Two cases added, suite **41 passed, 0 failed**. Both falsified under
`env -i bash --noprofile --norc` with `/bin/cp -f` restores checked by `diff -q`:

| Mutation | Case 39 | Case 40 |
|---|---|---|
| `gitCwd = repoRoot` removed | FAIL (exit 0) | FAIL (exit 0) |
| `ls-tree` pinned, `diff` unpinned via `process.cwd()` | pass | FAIL (exit 0) |

The second row is the good part: Case 40 discriminates a *partial* fix, which is
the failure mode a reviewer would otherwise have to take on trust. And Case 40's
comment states plainly that it does not isolate the `diff` derivation from the
`ls-tree` one, because the only working directory where `diff` misbehaves is one
where `ls-tree` has already returned nothing.

The two findings below are the gap in this coverage.

### Linter findings

`lint-shell.sh scripts/tests/test-spec-linter.sh` — exit **1**, three SC2086
(info) findings at lines **67, 544, 546**. Identical set to the pre-existing
baseline; the diff's hunk in that file starts at line 785, so no new finding is
introduced. All three sit in the pre-existing `run_case` / `run_base_case` helpers
where the unquoted expansion is deliberate. Not a finding against this PR.

`lint-markdown.sh` — not applicable, no Markdown in the diff.

### Findings

**Finding 1**

class: tech

`scripts/tests/test-spec-linter.sh:821-823` and `845-847` — **both new cases can
pass for the wrong reason.** Each asserts exit `1` plus the string
`specs/0240-nested-layout.md` in the output. That path is also printed by
**markdownlint** on any rule violation in the fixture spec, and markdownlint's
failure is itself exit `1` — so a fixture that drifts out of markdownlint
conformance satisfies both assertions while the base-branch check never executes.
Measured rather than argued: with an `MD001`/`MD022` violation injected into the
fixture and run against a linter **without** this PR's fix, exit is `1`, the path
appears, and both of Case 39's assertions hold while no offender is ever
identified.

```console
[exit 1]
specs/0240-nested-layout.md:11 error MD001/heading-increment …
  => case would report PASS while the base-branch check NEVER RAN: YES
```

This matters more here than it would elsewhere: these are the cases guarding a
check whose defect *was* passing without checking. Remediation: assert on a string
only this check emits — `Non-delta specs present on the base branch` — instead of
the file path. The fixture holds exactly one spec, so identity is not lost, and
markdownlint cannot produce that string.

**Finding 2**

class: tech

`scripts/tests/test-spec-linter.sh:843-847` — **`$GITFIX4` accumulates state across
its two cases, so they depend on execution order.** Case 40 sets
`git config diff.relative true` and appends to the offending spec; Case 39 relies
on neither being set. Reversing them makes Case 39 run with `diff.relative=true`
and a modified spec, which is not what its name claims to measure, and any third
case added to `$GITFIX4` later silently inherits both.

This is the **same finding, of the same class, that iter:1 raised against #732**
(cases 35–37 accumulating on `$GITFIX2`) and that was fixed there by having each
case establish the state it needs. Repeating the pattern two tickets later, in a
file that now carries the corrected idiom eight lines above, is the stronger
argument for fixing it than the fragility itself. Remediation: mirror the existing
idiom — reset before each case (`git -C "$GITFIX4" checkout -q -- .` plus
`git -C "$GITFIX4" config --unset-all diff.relative || true`), then apply only the
state that case measures.

### Considered and dismissed

- **`let gitCwd` is mutable module state, so `gitCapture` is order-dependent.**
  Calling `baseBranchPaths()` or `changedPaths()` before `resolveBaseContext()`
  would silently use the unpinned path. Dismissed: both are module-private and
  called from exactly one place, inside `resolveBaseContext` itself, after the
  assignment. Adding a `gitCwd !== null` assertion would guard a call graph that
  does not exist, and R15 requires the correct behaviour to be the *default* rather
  than something each caller verifies.
- **`--full-name` / `--no-relative` would be more local and need no module state.**
  Dismissed by R15, on the evidence in the delta: this defect exists because two
  path-sensitive invocations were each written without their author classifying
  them as such.
- **The behaviour change could newly redden a repository with nested specs.** True
  and intended — that is the fix. No such layout exists on `main` (`specs/org/` is
  manifest-excluded), and the whole-corpus run passes, so nothing changes state
  here.

hcross:
## Verdict: APPROVE

REVIEW iter:2 on `a25d004`. **Zero findings of any class.** Both iter:1 findings are
fixed, and each fix was verified to do what it claims rather than asserted to.

Same independence disclosure as iter:1: not cold-spawned, shared-identity posting
path, so this plain comment is the binding verdict artifact.

### CI status

All **13** required checks **pass** on `a25d004`, the exact head commit reviewed.
None failing, none pending. `lint-specs` reports `Results: 41 passed, 0 failed` then
`Linting passed!`.

### Correctness of the iter:1 remediations

Incremental diff `020a5e3..a25d004` is **19 insertions, 2 deletions in one file**.
`scripts/lib/spec-linter.js` is byte-identical to iter:1 — the production change was
not touched, which is right: both findings were in the tests.

**Finding 1 — the false-pass path is closed, measured on both sides.**

The assertion moved from the offending file's path to
`Non-delta specs present on the base branch`, the banner only this check emits.
Re-running the same adversarial probe — an `MD001`/`MD022` violation injected into
the fixture, against a linter **without** the fix:

| Assertion | Before iter:2 | After iter:2 |
|---|---|---|
| exit `1` | satisfied (by markdownlint) | satisfied (by markdownlint) |
| the required string appears | **satisfied** → false pass | **absent** → case correctly fails |

So the case can no longer report success on a run where the check never executed.
That was the one false pass that had to be impossible here, the defect under repair
being precisely a check that passed without checking.

**Finding 2 — order independence, verified by reversing.**

Each case now resets the work tree and unsets `diff.relative` before applying only
the state it measures, mirroring the `$GITFIX2` idiom eight lines above. Reversed
execution order (40 → 39) against the same fixture: **41 passed, 0 failed**.

**The falsification matrix was re-measured after the assertion change**, not
assumed to carry over:

| Mutation | Case 39 | Case 40 |
|---|---|---|
| `gitCwd = repoRoot` removed | **FAIL** (exit 0) | **FAIL** (exit 0) |
| `ls-tree` pinned, `diff` unpinned | pass | **FAIL** (exit 0) |

Both rows unchanged from iter:1, so the stronger assertion did not cost
discriminating power — including Case 40's ability to catch a partial fix, which is
the row that matters most.

All runs under `env -i bash --noprofile --norc`, restores by `/bin/cp -f` with
`diff -q` confirming the linter returned byte-identical.

### Convention compliance

Commit `a25d004` uses Gitmoji `✅`, appropriate for a test-only change, and its body
records the measurement rather than asserting the fix. Everything verified in
iter:1 is unchanged: spec-PR ordering, branch naming, `Closes #768` on the
implementation PR only, no version-bump or CLI-matrix trigger, pre-edit guard
satisfied.

### Test coverage

Unchanged in count (39, 40; suite at 41) and stronger in kind: the assertions now
key on a string that only the subject-under-test can emit. The disclosed limit of
Case 40 — that it does not isolate the `diff` derivation from the `ls-tree` one —
is unchanged and still stated in the file.

### Linter findings

`lint-shell.sh scripts/tests/test-spec-linter.sh` — exit **1**, three SC2086 (info)
findings at lines **67, 544, 546**: the pre-existing baseline, unchanged in count
and location, all outside this PR's hunks. No new finding introduced.

### Termination

`AGENTS.md` → *Retroactive review loop* → *Termination* requires APPROVE, zero
findings of any class, and green CI on the head commit reviewed. All three hold on
`a25d004`: APPROVE, zero findings, 13/13 pass. Terminates at MERGE.

PR #791, opened 2026-08-10
✨ Make a per-component install reach every served tier

### LINKED ISSUE #789: 🐛 manage-*-component.sh cannot install an org-tier component — first existing staging root wins
## Summary

`scripts/manage-claude-component.sh` can never install an `org`-tier skill on any
installation where the `community` tier holds at least one component. Its
staging-root resolution binds `SRC_DIR` to the first directory that merely
**exists**, instead of searching the roots for the requested component. Since
`dist/community/.claude/skills` exists as soon as the community tier is built,
`dist/org` is never reached.

The operator gets a misleading error that points at a missing build rather than
at the real cause (wrong staging root selected).

ADR-0011 and spec 0019 give `community` and `org` identical install semantics —
built always, home-installed on opt-in — so a per-component entry point that can
only ever reach one of the two silently implements half the contract.

## Reproduction

Reproduced on a fork at `fd0ec50`, macOS, Bash 3.2.

```console
$ bash scripts/build-components.sh
...
--- Tier: org (output root: /path/to/repo/dist/org) ---
Building skill: activity-journal
  Generated: /path/to/repo/dist/org/.claude/skills/activity-journal/SKILL.md
  ...

$ ls dist/community/.claude/skills dist/org/.claude/skills
dist/community/.claude/skills:
nano-banana

dist/org/.claude/skills:
activity-journal

$ bash scripts/manage-claude-component.sh install claude-skills activity-journal
Error: 'activity-journal' not found
$ echo $?
1
```

The build is correct — the component is staged exactly where ADR-0011 says it
should be. Only the resolution is wrong.

The documented Taskfile surface inherits the defect verbatim:

```console
$ task install-claude-component TYPE=claude-skills NAME=activity-journal
Error: 'activity-journal' not found
```

## Root cause

`scripts/manage-claude-component.sh:49-53`:

```bash
SRC_DIR=""
for staging in "$REPO_DIR/dist/community/.claude/skills" "$REPO_DIR/dist/org/.claude/skills"; do
  if [ -d "$staging" ]; then
    SRC_DIR="$staging"
    break          # <-- binds to the first EXISTING root, not the one holding $NAME
  fi
done
```

`SRC_DIR` is then consumed at line 128 by a per-component guard:

```bash
[ -d "$SRC_DIR/$NAME" ] || { echo "Error: '$NAME' not found"; exit 1; }
```

So the loop answers "which root exists?" while the guard asks "which root holds
the component?" — two different questions. The `break` makes the second
unanswerable for any root after the first.

## Parity across the four sibling scripts

`manage-copilot-component.sh` already implements the correct pattern; the other
three do not. The bug is therefore a **parity gap**, not an isolated slip.

| Script | Site | Roots iterated | Verdict |
|---|---|---|---|
| `manage-copilot-component.sh` | `:119-125` | `artifacts/{core,library,community,org}/skills` | ✅ **correct** — `[ ! -d "$SRC_DIR" ] && continue`, then tests `[ -d "$SRC_DIR/$NAME" ]` per root. This is the reference pattern. |
| `manage-claude-component.sh` | `:50-53` | `dist/{community,org}/.claude/skills` | ❌ first existing root wins (`break`). Reproduced above. |
| `manage-antigravity-component.sh` | `:91-94` | `dist/{community,org}/.agents/skills` | ❌ identical pattern, identical consequence for its own overlay tiers. |
| `manage-workspace-component.sh` | `:30-32` | `artifacts/{community,org}/$TYPE` | ❌ variant: org is used **only if** `artifacts/community/$TYPE` does not exist. `artifacts/community/skills/` does exist (it ships `nano-banana`), so `artifacts/org/skills/` is unreachable. |

`manage-workspace-component.sh` is worth calling out separately: the comment
immediately above the defective code states the intent explicitly, and the code
does not realise it.

```bash
# Resolve the overlay source tier. Default is the community sandbox; the org
# tier (spec 0019 — only skills/ and agents/) is searched as a fallback so a
# single-component install covers org symmetrically with Copilot/Claude.
SRC_DIR="$REPO_DIR/artifacts/community/$TYPE"
if [ ! -d "$SRC_DIR" ] && [ -d "$REPO_DIR/artifacts/org/$TYPE" ]; then
  SRC_DIR="$REPO_DIR/artifacts/org/$TYPE"
fi
```

"Searched as a fallback" describes the intended behaviour; the guard implements
"used only when community is absent" instead. A directory-existence test cannot
express a component search.

## Impact

The only working paths to install a single `org` component on Claude Code today:

1. `bash scripts/setup-claude-interactive.sh` — the canonical entry point per
   `docs/cli-matrix.md` row 10. It offers the `org` opt-in prompt correctly, but
   it is heavyweight for one component: it also rewrites the layered context
   rules, re-registers MCP servers, and re-offers the transcript hooks.
2. A hand-rolled copy replicating `install_tier_to_home()` from
   `scripts/setup-claude-interactive.sh:375-404` —
   `cp -R dist/org/.claude/skills/<name> ~/.claude/skills/`. This is the
   workaround actually used.

Neither is the documented per-component surface, which is precisely what
`manage-*-component.sh` and `task install-*-component` exist to provide.

## Suggested fix

1. **Resolve the component, not the directory.** Adopt the
   `manage-copilot-component.sh:119-125` pattern in the three affected scripts:
   iterate every root, `continue` past absent ones, and test
   `[ -d "$root/$NAME" ]` inside the loop. For the bulk (no `$NAME`) mode, place
   every component from every root rather than only the first root's.
2. **Add an explicit tier selector** — e.g. `--tier org`, or a `TIER=` Taskfile
   variable — so an operator can disambiguate a component name present in more
   than one tier. Without it, precedence between `community` and `org` for a
   shared name stays implicit and order-dependent.
3. **Make the not-found message enumerate the roots actually searched.** The
   current wording sent the diagnosis in the wrong direction; listing the
   searched roots would have made the real cause self-evident.
4. **Consider a shared helper in `scripts/lib/`.** Four scripts implement four
   variants of "resolve a component across overlay tiers", one of which is
   correct. A single `resolve_component_across_tiers()` would make the parity
   rule structural instead of a convention each script re-derives.

## Secondary observation (not part of this bug)

The four scripts disagree on what they read: `claude` and `antigravity` resolve
from the built staging tree `dist/<tier>/`, while `copilot` and `workspace`
resolve from the sources under `artifacts/<tier>/`. That asymmetry may well be
deliberate, but it is worth confirming while touching these code paths — a fix
applied uniformly would otherwise have to straddle both conventions.

## Environment

- Fork of `crewrig/crewrig` at `fd0ec508e3711bd192c044563103c1127d3b2270`
- macOS (Darwin 25.5.0), Bash 3.2
- CLI affected: Claude Code (primary), Antigravity CLI and the generic workspace
  installer by inspection

hcross:
Make the four per-component install commands resolve a named component from every tier they serve, deliver it where the assisted setup of the same CLI delivers, and fail loudly instead of silently, implementing `specs/0119-overlay-tier-component-resolution.md`. The reported defect — an `org` component unreachable once `community` exists — turned out to be the smallest of five faults on this path, two of which made a command exit 0 having installed nothing usable.

Closes #789

## How to read this PR?

13 files, +1993 −212, in three commits that are worth reading in order:

1. **`9c499d6`** — the regression suite, failing. `scripts/tests/test-component-tier-resolution.sh` (1074 lines, 22 cases) plus its CI wiring in `.github/workflows/build.yml` and `ci/ci-capabilities.yml`. Read this first: it is the executable statement of what was broken.
2. **`2d07407`** — the implementation, 10 files.
3. **`c6da611`** — a fixture fix, 33 lines moved, no behaviour change. See *Detailed description* for why it is a separate commit.

**The load-bearing file is new: `scripts/lib/component-resolve.sh` (598 lines).** All four `manage-*-component.sh` commands now source it (`. "$(dirname "$0")/lib/component-resolve.sh"`), and the bulk of the −212 is the four copies of divergent resolution logic it replaces. Read it after the test and before the four commands, which become thin by comparison.

### The five defects, because the ticket title only names one

| # | Defect | Consequence |
|---|---|---|
| 1 | Resolution bound to the first staging root that merely *existed* | `org` unreachable once `community` was populated — **the reported one** |
| 2 | `manage-copilot-component.sh` wrote to the committed project tree, and read authoring sources while its setup reads compiled output | Every community/org install dirtied the operator's checkout |
| 3 | `for item in "$SRC_DIR"/*/` yields a trailing slash; BSD `cp -rf src/foo/ dest/` copies foo's **contents** | Unnamed install delivered **zero usable components** on three of four commands, **exit 0** |
| 4 | A glob missing `.gitkeep` stays literal; the failing test is the last command of its case arm; `set -e` aborts the wrapper's loop | `task install-workspace` installed **nothing at all**, silently, for every type |
| 5 | Copilot named-miss was silent, and its non-zero status was accidental | Status flipped to 0 when the last searched root was absent |

Defects 3 and 4 are the ones to scrutinise. Both are *silent success* — the command reports done and delivered nothing. Neither would ever have been filed as "an org component cannot be installed", and neither was in the original diagnosis.

### Three deviations from PLAN v4 — flagged deliberately, please contest

The `developer` raised these rather than burying them. Each is a judgment call REVIEW may reverse:

1. **`mcp-servers` on the Copilot command moved to all three overlay tiers**, where PLAN v4 step 6 said "unchanged". Argument for: leaving it community-only makes Copilot the sole command carrying that asymmetry, which pillar 5 (multi-CLI parity) prohibits, and the parallel bullets in steps 4/5/7 read "landing zones unchanged". Argument against: no requirement compels the wider set. **No test covers it either way** — that gap is the strongest reason to contest.
2. **Step 3's "unnamed request rebuilds first" is gated on types that resolve from compiled output.** The never-compiled types have no staging tree to refresh, so the rebuild would be a no-op with a cost. Case 11 depends on this gating.
3. **`artifacts/FORMAT.md` was edited** though it sat outside the ownership list the developer was given. It is PLAN v4 step 9 and R14's stated home. The edit also corrects "**Bash 4+** for associative arrays" to the **3.2.57** floor the repository actually enforces via `ci/bash32-forbidden.txt` and `scripts/check-bash32-portability.sh` — the old line contradicted a guard that has been failing builds on that basis.

### One correction to PLAN v4

**Step 11 is wrong: four types survive the truncation case, not five.** Found independently by the `tester`, confirmed by the `developer`. A colliding `skills` name poisons the build pre-pass for every compiled type, and `agents` resolves from compiled output too. The plan comment is immutable, so this is recorded here and on issue #789.

## How to test this PR?

Prerequisite: a POSIX shell and the repo checked out. No build or install is required — the suite constructs its own fixture tree and asserts it leaves the real repository untouched.

```sh
bash scripts/tests/test-component-tier-resolution.sh
```

Expected: `Result: 22 passed, 0 failed`, exit 0. The final case is a hygiene assertion (`the test leaves the real repository unchanged`) — if that one fails, the suite has polluted your checkout and you should treat the rest of the run as untrustworthy.

The five guards this change touches, each expected to exit 0:

```sh
for g in check-no-machine-paths check-bash32-portability check-test-wiring check-ci-parity check-core-paths; do
  bash scripts/$g.sh >/dev/null 2>&1; echo "$g: $?"
done
```

### Failure modes worth exercising

The suite covers these, but they are the ones to reproduce by hand if you want to see the bugs:

- **The reported defect.** Populate `dist/community/.claude/skills` with anything, put a component only in `dist/org/.claude/skills`, and request it by name. Before this change the command reported the component missing; now it installs.
- **Silent success (defect 3).** Run an unnamed install on `main` and inspect the destination: the directories arrive containing the *contents* of each component rather than the component. `cp` semantics differ between BSD and GNU here, so reproduce on macOS.
- **Bash 3.2 portability.** `/bin/bash` on macOS is 3.2.57 and is the enforced floor. `bash scripts/check-bash32-portability.sh` rejects `declare -A` and `mapfile`/`readarray`.

### Verification already performed

Re-run directly for this PR, real exit codes captured (not read through a pipe): suite **22/22, exit 0**; all five guards **exit 0**. The `team-lead` separately verified 22/22 under three configurations — `/bin/bash` 3.2.57, bash 5.3.15, and with the task runner genuinely absent. That third run is worth knowing about: the first attempt at it was a **false green**, because a PATH filter left `task` reachable and the run silently repeated configuration one. The re-run asserts the runner's absence before executing and aborts if it is present.

### Change surface

| File | Δ | Role |
|---|---|---|
| `scripts/lib/component-resolve.sh` | +598, new | Shared resolution/install library — the load-bearing addition |
| `scripts/tests/test-component-tier-resolution.sh` | +1074, new | 22-case regression suite |
| `scripts/manage-copilot-component.sh` | 131 ± | Landing zone and resolution basis corrected |
| `scripts/manage-workspace-component.sh` | 101 ± | Resolution basis corrected |
| `scripts/manage-claude-component.sh` | 96 ± | Resolution replaced by the library |
| `scripts/manage-antigravity-component.sh` | 93 ± | Resolution replaced by the library |
| `scripts/build-components.sh` | 43 ± | Collision pre-pass (R13) |
| `artifacts/FORMAT.md` | 32 ± | R14 documentation home; Bash-floor correction |
| `scripts/install-workspace.sh` | 26 ± | Defect 4 — silent no-op loop |
| `docs/cli-matrix.md` | 6 ± | Parity matrix maintenance |
| `.github/workflows/build.yml` | +3 | CI wiring for the suite |
| `ci/ci-capabilities.yml` | +1 | GitLab parity for the same |
| `.gitlab-ci.yml` | +1 | Generated from the above |

`scripts/lib/component-resolve.sh` is mode `100644`, matching the five other sourced libraries in `scripts/lib/` (`common.sh`, `provenance-carrier.sh`, `render-command.sh`, `tls-delegation.sh`, `tls-exec.sh`). It is sourced, never executed, so the absent exec bit is correct rather than a lost mode.

### Why the fixture fix is its own commit

`c6da611` renames the suite's fixture root from `home` to `cli-home` and adds `cli-home*/` to the fixture `.gitignore`. A directory named `home` reads as a machine path to `check-no-machine-paths`, so **two guards failed on a test file whose subject is component resolution** — a guard failure entirely unrelated to what the test tests.

It is separate because a test is not the implementer's to edit. The `tester` authored the suite; the `team-lead` made this correction under its own authorship rather than folding it into `2d07407`. Same reason `9c499d6` carries a committer distinct from its author: spec 0114's claim gate refuses a dirty tree, the residue belonged to an agent already released, and residue an agent did not author is not theirs to resolve by discarding.

### Staleness: worked around, not solved

`scripts/lib/component-resolve.sh:529-536` documents a limitation this PR does not fix:

> A compiled tree is stale by default and nothing in the repository can detect that — `ensure_tier_built` returns 0 the moment the staging directory exists, so it sees absence and never staleness.

`ensure_tier_built` (`scripts/lib/common.sh:447`) gates on directory presence alone, so a tier compiled before a component was added stays stale indefinitely. The workaround: a named miss is **not final** — the served overlay tiers are rebuilt once and the search retried, so a stale tree costs a rebuild rather than a false "not found". Detecting staleness properly needs a versioned or content-addressed `dist/`, which is not in this change.

### Spec conformance

Implements `specs/0119-overlay-tier-component-resolution.md` (merged at `061b5cf`). R13's build-time collision refusal lands in `scripts/build-components.sh`; R14's documentation home is `artifacts/FORMAT.md`; R15's install-time refusal and R16–R18's failure reporting live in the shared library. R20's regression coverage is the 22-case suite, wired into both CI engines.

### Lifecycle

DEV stage under ADR-0010, second and final PR for this ticket after spec-PR #790. Issue #789 is the logbook and closes on merge per Rule C — hence the closing keyword here, which the spec-PR deliberately omitted so the logbook would survive that merge. REVIEW follows; the three deviations and the PLAN step-11 correction above are the intended targets.


hcross:
## Verdict: APPROVE

Cold review — the reviewer was given the PR number and nothing else. Reviewed at `c6da611`, `updatedAt` `2026-08-10T18:58:50Z`, both re-observed immediately before posting.

Posted as a plain comment rather than a review event: the authenticated identity (`hcross`) equals the PR author, so GitHub refuses a self-review. Per the `pr-reviewer` skill's step-6 ladder this comment is the canonical verdict artifact.

No blocking findings. Three non-blocking findings, each classified, below.

### CI status

All 13 required checks **pass** at `c6da611` — `build`, `check-agents-size`, `check-ci-parity`, `check-components`, `check-extension-version-bump`, `check-feedback-routing`, `check-skill-versions`, `check-spec-id-reserved`, `gitlab-ci-check`, `grep-anti-patterns`, `lint-markdown`, `lint-specs`, `test-harness-curate`. Run 31421811626. No failing or pending job. `mergeStateStatus: CLEAN`.

Six guards re-run locally with exit codes captured directly from the command (not through a pipe): `check-no-machine-paths`, `check-bash32-portability`, `check-test-wiring`, `check-ci-parity`, `check-core-paths`, `check-skill-versions` — **all exit 0**.

### Correctness

Verified against `specs/0119-overlay-tier-component-resolution.md` at `061b5cf`, read directly rather than through the PR body's account of it.

**The suite is load-bearing, not decorative.** This is the check most likely to have sunk the PR, so it was run first. Four of the five defects were reintroduced one at a time into a throwaway clone and the suite re-run:

| Mutation | Suite result |
|---|---|
| M1 — `break` inside the outer loop of `resolve_component_in_roots` (defect 1, first-existing-root-wins) | **19/22, exit 1** — fails R7 org-while-community-populated, R15 ambiguity, R19 entry points |
| M2 — `"$root"/*/` trailing-slash glob restored in `enumerate_components_in_roots` (defect 3) | **16/22, exit 1** |
| M3 — `install-workspace.sh` loop bare under `set -e` (defect 4) | **21/22, exit 1** |
| M4 — R13 pre-pass disabled in `build-components.sh` (`if false`) | **18/22, exit 1** |

A caveat worth recording, because it nearly became a fabricated finding: the *first* M1 attempt left the suite green at 22/22. The mutation was inert — the inserted `break` landed after the outer loop's `done`, where `break` is a no-op, not inside it. The mutation was verified applied by `git diff` (`+    break`), which is exactly why it looked like a genuine coverage gap. Re-inserting at the correct line turned the suite red. A reviewer stopping at the `git diff` proof would have filed a coverage finding against a suite that does cover the reported defect.

**The collision guard keys on what the build actually writes.** `installed_targets` derives the name through `_component_declared_name` (`scripts/lib/component-resolve.sh:63-85`), which reads the frontmatter `name:`. `scripts/build-components.sh:411`, `:603` and `:700` each derive their output path from `name=$(yaml_field "$source" "name")`. Same key. A guard keyed on the directory or file name would have missed every component whose declared name differs from its path — the guard would have covered something the build does not produce.

**Every numeric claim in the new code and docs verifies exactly.** Sourcing the library against this repository's `artifacts/`:

- `component-resolve.sh:287-289` claims 172 installed targets → **172**.
- Same comment claims zero duplicates keyed on `(tier-class, install-target)` → **0**.
- Same comment claims nine keyed on name alone → **exactly 9**: `architect`, `developer`, `doc-writer`, `harness-curator`, `pr-logbook`, `pr-reviewer`, `security`, `spec-author`, `tester`. `artifacts/FORMAT.md`'s "`architect` … Eight other name pairs" is the same nine.

**The `docs/cli-matrix.md` row 4 correction is right, and the cell it replaces was wrong.** The new cell claims Antigravity agents build to `.agents/agents/<name>/AGENT.md` (directory). `scripts/build-components.sh:785` writes `check_or_write "$out_root/.agents/agents/$name/AGENT.md"`. The previous cell read `.antigravity/agents/<name>.md` (flat file), which matched neither `build_agents` nor row 1's `.agents/` root. Correcting it inside this PR is not scope creep: `installed_targets:349` emits `.agents/agents/$name` as a collision key, so the guard would have been unauditable against its own reference table.

**No regression from widening `*.json` globs to `*`.** `enumerate_components_in_roots` now enumerates `"$root"/*` where the old per-type loops globbed `*.json` for `mcp-servers` and `themes`. A stray non-JSON file would now reach `register_json_entry` / `merge_json_entry` and return 1, failing the unnamed install where it was previously ignored. Checked: `artifacts/community/{mcp-servers,themes,policies,hooks,commands}` each contain only `.gitkeep`, which the enumerator skips explicitly (`:213`), and no `library`/`org` counterpart directory exists. Inert at this commit, and the new behaviour (report rather than silently skip) is the better one.

**Exit-status contract holds at the call sites.** Every `component_install_*` call site uses `… || exit $?`, where `$?` is the driver's status, not a pipe's. The library's "return 0, carry the answer on stdout" contract (documented at `:15-32`) is the right call under the callers' `set -e`: a resolver returning 1 for "not found" would abort at `x=$(f)` and discard stdout — the silent failure the ticket exists to remove.

### Convention compliance

- **Gitmoji** — title `✨ Make a per-component install reach every served tier`; all three commit subjects carry gitmoji (`✅`, `✨`, `🐛`). Conforms to `AGENTS.md` → *Naming Convention*.
- **`docs/cli-matrix.md` co-maintenance** — mandatory here (the diff touches `artifacts/**`, `scripts/build-components.sh`, and four `manage-*.sh`). Rows 4, 12 and 15 updated in the same diff. Row 12 now records the per-command landing zone *and* basis, which is what makes R1/R2 auditable from the matrix rather than only from the scripts.
- **Branch naming** — `fix/0789-overlay-tier-component-resolution` matches `<prefix>/<NNNN>-<slug>`.
- **PR body** — carries *How to read this PR?*, *How to test this PR?* and *Detailed description (for agents)* per `docs/pr-format.md`.
- **Version-bump convention** — no skill or agent source is touched; `artifacts/FORMAT.md` is documentation and carries no `metadata.provenance`. `check-skill-versions` exits 0 locally and in CI.
- **Logbook** — `Closes #789` is correct here and deliberately contrasted with the spec-PR #790, which omitted it so the logbook survived that merge (Rule C).
- **File modes** — `scripts/lib/component-resolve.sh` is `100644`, matching exactly the five sourced siblings the PR body names (`common.sh`, `provenance-carrier.sh`, `render-command.sh`, `tls-delegation.sh`, `tls-exec.sh`). The new test is `100755`, matching every sibling under `scripts/tests/`.

### Test coverage

`scripts/tests/test-component-tier-resolution.sh` — 1074 lines, 22 cases. **Re-run in three configurations rather than transcribed**, each exit code read from the command itself:

| Configuration | Result |
|---|---|
| `/bin/bash` 3.2.57 (the enforced floor) | **22 passed, 0 failed — exit 0** |
| `bash` 5.3.x | **22 passed, 0 failed — exit 0** |
| `task` runner genuinely absent | **22 passed, 0 failed — exit 0** |

The third configuration was built to be falsifiable, since a previous attempt on this ticket produced a false green from a PATH filter that left the runner reachable. A shadow directory was populated with a symlink to every executable on the real `PATH` except `task` (2858 entries), `PATH` was set to it alone, and the run was **gated on three independent probes**: `command -v task`, `type task`, and `[ -x "$SHADOW/bin/task" ]` — all three required to be non-zero or the script aborts with exit 90 rather than running. A fourth assertion confirmed the shadow had not lost `bash git jq yq awk sed grep basename mktemp`, so a green could not come from a broken toolchain either. Case 20 takes its documented second driver in this configuration, extracting each entry's `cmd:` from `Taskfile.yml` instead of skipping — which is what keeps R19 asserted on a runner-less machine rather than only on a developer box.

Case 22 (`hygiene: the test leaves the real repository unchanged`) passed in all three runs, and `git status --porcelain` on the worktree is empty afterwards. The suite is self-fixturing and left no residue.

### Linter findings

`lint-markdown.sh` on `artifacts/FORMAT.md` and `docs/cli-matrix.md` — **exit 0**, no output.

`lint-shell.sh` could not be used as the skill documents it: invoked with the eight changed shell paths as positional arguments it reported

```
lint-shell.sh: scripts/lib/component-resolve.sh scripts/tests/... : not a regular file
```

treating the whole argument list as a single path, then exiting 1. `shellcheck -S warning` was run directly on each file instead. Findings, all triaged as non-blocking (shellcheck is not wired into this repository's CI — no workflow under `.github/workflows/` or `ci/` invokes it):

- `component-resolve.sh:423,426` — **SC2089/SC2090**. False positives. The single quotes in `"tier '$rec_tier' declares …"` are intentional literals in report text, and `:426` is correctly guarded as `${sources[@]+"${sources[@]}"}`. Confirmed behaviourally: the three R13 cases assert the report names both declaring tiers, and they pass.
- Four `manage-*.sh` — **SC2034 `REPO_DIR` appears unused**. False positive: `REPO_DIR` is consumed by the sourced library as `${REPO_DIR:?}` (`:130`, `:143`, `:474`). Persists under `shellcheck -x`, because the source path is computed (`$(dirname "$0")/lib/...`) and the `# shellcheck source=` directive does not resolve it.
- Four `manage-*.sh` — **SC2115** on `rm -rf "$dest_dir/$item_name"`. **Pre-existing**, present at `061b5cf:scripts/manage-copilot-component.sh:53`; not introduced here, and not reachable (both operands are non-empty by construction at every call site).

---

## Findings

**Finding 1 — `class: tech` (non-blocking).** R1/R2/R5/R11 are asserted for the `skills` type across all four commands (cases 19 and 20), but the Gemini command's `agents` type — the other type whose assisted setup resolves from the same compiled basis — has no equivalent assertion, and the `org` tier is exercised for `agents` on no command. Verified by inspection that the property holds: `setup-gemini-interactive.sh:316` sets `GEMINI_AGENTS_HOME="$GEMINI_HOME/agents"` and `:324` reads `dist/$tier/.gemini`, while `manage-workspace-component.sh` sets `DEST="$GEMINI_HOME/$TYPE"` and `component_set_staging_roots ".gemini/agents"` — same landing zone, same basis. So this is a coverage gap, not a defect, and R20's clause about a basis diverging from the assisted setup arguably reaches it. Cheap to close by parameterising case 19/20 over `agents`.

**Finding 2 — `class: tech` (non-blocking). Deviation 1, contested and upheld.** Moving Copilot `mcp-servers` to all three overlay tiers is the right call and I would keep it. The PR offers parity as the argument; the stronger one is that nothing forbids it. `specs/0019-artifact-build-install-scope.md` places **no type restriction on the `org` tier** — R1 compiles every tier present and R3 requires an `org` component be compiled identically to any other tier — so the widening does not "extend the `org` tier to further types" in the sense spec 0119's *Out of scope* bars. R5 is a floor ("SHALL serve every non-`core` tier that the assisted setup installs"), not a ceiling, and R7 keys success on "any served tier", which each command defines for itself. The absent test is also less alarming than the PR body suggests: `artifacts/library/mcp-servers` and `artifacts/org/mcp-servers` do not exist — only `community` carries that type — so the change alters no observable behaviour at this commit and a test would assert against empty directories. Residual cost, worth one line somewhere: `report_unresolved` now lists two never-populated roots as "examined".

**Finding 3 — `class: tech` (non-blocking).** `ensure_overlay_tiers_fresh` prunes `dist/<tier>/<cli-root>` for all three tiers *before* invoking the build (`:486`), so a build refused by the new R13 pre-pass leaves the operator with a pruned staging tree and a build that keeps refusing until the collision is fixed. The message "nothing has been installed" (`:500`) is true but silent about the prune. Self-inflicted and self-healing once the collision is resolved (`ensure_tier_built` rebuilds), so this is a wording nit, not a behaviour change I would ask for.

### On the other two deviations and the plan correction

- **Deviation 2** (unnamed-request rebuild gated on types that resolve from compiled output) — sound. A never-compiled type has no staging tree to refresh, so the rebuild would be a no-op with a cost. Case 11 depends on the gating. No objection.
- **Deviation 3** (`artifacts/FORMAT.md` edited outside the stated ownership list) — endorsed. It is R14's stated documentation home, so the requirement could not land anywhere else. The Bash-floor correction is independently verified: `ci/bash32-forbidden.txt` is the guard's list, `scripts/check-bash32-portability.sh` exits 0 here, and the previous "**Bash 4+** for associative arrays" line did contradict a guard that fails builds on exactly that basis.
- **PLAN v4 step 11 correction** (four types survive truncation, not five) — recorded in the PR body and on #789; plan comments are immutable, so nothing further is actionable.

### What could have made this not an approval

Recorded so the approval is falsifiable rather than a shrug:

1. **A decorative suite.** Four reintroduced defects, four red runs. Had M1 stayed red-proof-free, coverage of the *reported* defect would have been unproven.
2. **A false green on the runner-absent configuration** — the failure mode this ticket already hit once. Gated on three probes plus a toolchain sanity check, aborting rather than running if any probe found `task` reachable.
3. **A collision guard keyed on the wrong thing.** If `installed_targets` had keyed on path while the build keys on frontmatter `name`, R13 would guard something the build never produces. Both sides checked and identical.
4. **A doc correction that was itself wrong.** Row 4's new Antigravity cell was checked against `build_agents` rather than accepted.
5. **Invented numbers.** 172 / 0 / 9 recomputed from the library against this repository; all three exact.
6. **A silent regression from the `*.json` → `*` widening.** Every overlay artifact directory inspected; all hold only `.gitkeep`.
7. **A dirtied checkout** — R3's whole subject. `git status --porcelain` empty after all runs.

hcross:
## Correction to the review above — one claim was mine, not the PR's

The verdict is unchanged: **APPROVE**, still at `c6da611`. One statement in the *Linter findings* section was wrong, and it was wrong in the reviewer's favour, so it is corrected here rather than quietly.

**What I claimed.** That `lint-shell.sh` "could not be used as the skill documents it", treating a list of paths as a single argument and exiting 1 on `not a regular file`.

**What was actually true.** `lint-shell.sh` handles multiple positional arguments correctly (`for file in "$@"`, `scripts/lint-shell.sh:26`). The tool shell here is **zsh 5.9**, not bash, and zsh does not word-split an unquoted parameter expansion. I had collected the eight paths into one variable and passed it unquoted, so the script received exactly one argument — the whole list as a single path — and reported it accurately. Re-run with the eight paths as separate arguments, the script works as documented.

This is precisely the error class the review spends its length checking for elsewhere: a signal read as evidence about the thing under review when it was evidence about my own harness. It is also the second time this session that a green-or-red result nearly became a finding about someone else's code — the first was the inert M1 mutation, which is disclosed in the review above.

**What the correct run adds.** Four notes not in the original review, none of which change the verdict:

- `component-resolve.sh: not executable in git (mode 100644; expected 100755)` — the linter's heuristic assumes every `*.sh` is executed. This one is sourced, never executed, and `100644` matches its five siblings in `scripts/lib/`. The PR body pre-empts this exact point, correctly.
- `component-resolve.sh: missing 'set -e'` — correct for a sourced library, which must not impose `errexit` on its callers. The library instead documents its own exit-status contract at `:15-32`, which is the stronger answer.
- **SC2086** at `component-resolve.sh:534` and `:573` — `ensure_overlay_tiers_fresh "$refresh_cli" $COMPONENT_OVERLAY_TIERS`, unquoted. **Deliberate and load-bearing**: the word-splitting is the mechanism that passes three tier names as three arguments. Taking shellcheck's suggested fix would pass `library community org` as a single argument, and `:486` would then evaluate `rm -rf "$repo/dist/library community org/$cli_root"`. The value is a fixed literal with no glob characters, so the split is safe. An `# shellcheck disable=SC2086` with that one-line reason would be worth more than the fix.
- **SC2016** in the test at `:792-795` — single-quoted `sed` expressions, intentional.

All remain non-blocking, and shellcheck is still not wired into this repository's CI.

hcross:
## Verdict: APPROVE

Iteration 2, reviewed at `373fe66`. Three findings, all `class: tech`, none blocking — the code under test is correct and R20 is satisfied. Per the retroactive loop this pass surfaces findings, so it does not terminate the lifecycle.

Both new commits were audited as new code, not as answers. The direct question put to this pass — *is the rest of the suite in the same position anywhere else?* — has an answer, and it is yes.

### CI status

13/13 required checks **pass** at `373fe66`; `mergeStateStatus: CLEAN`. Six guards re-run locally with exit codes taken from the command itself — `check-no-machine-paths`, `check-bash32-portability`, `check-test-wiring`, `check-ci-parity`, `check-core-paths`, `check-skill-versions` — **all exit 0**. Worktree clean afterwards.

### The three configurations, re-established

23/23, exit 0, in each: `/bin/bash` 3.2.57, bash 5.3.15, and `task` genuinely absent (2858-entry shadow `PATH`, gated on three probes — `command -v`, `type`, `-x` — plus a toolchain sanity check, aborting rather than running if any probe found the runner).

**The tester's falsification of configuration 3 is confirmed independently.** Corrupting only the fallback driver's `NAME` substitution (`{{\.NAME}}` → `{{\.NAMEZZ}}`, test:1076) leaves configuration 1 at **23/23 green** and turns configuration 3 **red at 22/23**, case 20 failing. Configuration 3 therefore exercises the extracted-`cmd:` driver rather than silently repeating configuration 1. That is the check whose absence produced the earlier false green, and it now holds.

### F3 — fixed, and verified live rather than from the diff

Reproduced independently: seeded a colliding `dup-name` in `library` and `org`, pre-populated `dist/library/.claude/skills/sentinel`, then drove a named-miss install. The root existed before, did not after, exit 1, and the message named all three pruned roots by absolute path. The claim describes something that happens.

**The moved `core` validation is right and is the better half of the commit.** Rejecting `core` mid-loop would have pruned the earlier tiers and returned before reaching the reporting branch — the same silent half-done state the message exists to prevent, reachable only by internal misuse but reachable. Hoisting it into its own loop ahead of the prune is correct, and it is not scope creep: it is the identical defect the commit is about, at a smaller scale.

### F1 — closed, and the fix is genuinely live

The staging-root assertion now reads the argument of `component_set_staging_roots` rather than any incidental mention. Confirmed by four declaration-targeted mutations, none of which my first pass probed:

| Mutation | `R20/R1` | `R20/R2+R5` |
|---|---|---|
| Gemini **skills** root → `.gemini/sub-skills` | PASS | **FAIL** |
| Gemini **agents** root → `.gemini/sub-agents` | PASS | **FAIL** |
| Copilot skills root → `.github/sub-skills` | PASS | **FAIL** |
| Gemini landing zone → `$GEMINI_HOME/x-$TYPE` | **FAIL** | PASS |

Each fails the case that binds it and only that case. The first row is verbatim the mutation that reported PASS before this commit.

**The stated cost is real, correctly described, and the right trade.** An indirect declaration (`SR=".gemini/skills"; component_set_staging_roots "$SR"`) fails the case while every behavioural case stays green, and the diagnostic prints `it declares: .gemini/agents $SR` — naming what it found so the next author can restore the literal. Given the substring form was provably unable to fail, a false-fail that prints its own remedy is strictly better than a false-pass that prints nothing.

**The `(CLI, type)` row set is exhaustive, and for a stronger reason than the enumeration given.** The enumeration checks out — `manage-claude-component.sh` and `manage-antigravity-component.sh` declare no `agents` arm (their dispatch types are `{claude-skills, policies, mcp-servers}` and `{antigravity-skills, policies, mcp-servers}`), and `manage-copilot-component.sh` refuses `agents` by design. But the load-bearing reason is that the property case 19 asserts — a resolution basis diverging such that one route reads compiled output while the other reads authoring sources — can only arise for a type the build compiles. Only `skills`, `commands` and `agents` compile, and no assisted setup installs `commands`. So `skills` across four CLIs plus `agents` on Gemini is the complete set, not a sample. No fifth pair.

---

## Findings

**Finding 4 — `class: tech` (non-blocking). The identical vacuity survives three lines below the fix, in the same case.**

The tier-set loop at `scripts/tests/test-component-tier-resolution.sh:978` still asserts by bare substring over the same scan set:

```sh
for t in library community org; do
  printf '%s\n' "$scan" | xargs grep -qw -- "$t" 2>/dev/null || { ok="false"; … }
done
```

Measured: mutating `COMPONENT_OVERLAY_TIERS="library community org"` → `"library community"` — verbatim the R5/R7/R11 divergence that makes `org` unreachable on every command, on both request shapes — turns **8 behavioural cases red** and leaves `R20/R2+R5` reporting **PASS**, despite that case's own title claiming "over all three overlay tiers".

Root cause pinned exactly: with `org` removed from the served set, `grep -qw org` still matches `scripts/lib/component-resolve.sh` at **`:9`** (`a component in \`org\` became`) and **`:300`** (`a library-versus-org command-name collision`) — two comments. Reproducing the assertion's own probe verbatim against `manage-copilot-component.sh` plus the library succeeds. This is the same failure shape the commit just fixed: the scan set deliberately includes the shared resolver, and the resolver mentions every tier name for unrelated reasons.

Severity is genuinely lower than the one that was fixed, and I want to be precise about why rather than overstate it: **R20 remains satisfied**, because the eight behavioural cases do fail. This is a misleading assertion, not a coverage gap. It nonetheless reads as coverage that does not exist, which is the whole objection that produced this iteration.

The fix already exists in the same file, three lines up: read the declaration rather than the mention. `COMPONENT_OVERLAY_TIERS` is a single assignment in a known file, so asserting on its resolved value — as `resolved_by_name` already does for landing zones — would bind the tier set the way the staging root is now bound.

**Finding 5 — `class: tech` (non-blocking). Two negative guards cannot fire against the text the code emits.**

Measured by capturing each reporter's real output and running the test's own pattern against it:

- **`:309`** — `grep -qiE 'collision|colliding|ambiguous'`, guarding "a phantom collision was reported against build residue". `report_collision` emits `Refusing 'acme-review': one installed name is claimed by more than one component.` / `Every source presenting it, in no significant order:`. None of the three alternatives occurs. **Cannot fire.**
- **`:675`** — `grep -qiE 'not found|unresolved|searched'`, guarding "a refused rebuild was reported as an unresolved name". `report_unresolved` emits `Error: no component named 'zz-absent' … resolved in any served tier.` / `Locations examined, in resolution order:`. `unresolved` does not match `resolved`; the report says *examined*, not *searched*. **Cannot fire.**

(`:796`'s `build-components\.sh` guard **does** match the emitted build hint and is live — the pattern-matching method is sound, these two patterns are not.)

The two differ in consequence, and only one matters:

- `:675` is **redundant**. Case 13's `grep -Fq 'acme-review'` already catches the misreport: were the refused rebuild reported as an unresolved name, the colliding name would be absent. Dead weight only.
- `:309` is **not redundant, and leaves a real hole**. Case 2's other two assertions are "`doomed-skill` is not installed" and "`keeper-skill` is installed". A phantom collision raised against `doomed-skill` satisfies **both** — the collision is precisely what would stop it installing. `:309` is the only assertion that could catch it, and it cannot. The regression named in that case's own header is undetectable by that case.

Cheapest fix: match what is emitted — `Refusing|claimed by more than one` and `resolved in any served tier|Locations examined`.

Worth one line on the side: **R17 says "name every tier that was searched"; the implementation says "Locations examined"; the test greps for "searched".** That drift is the root cause of the second half of this finding, and aligning the vocabulary would close it at the source.

**Finding 6 — `class: tech` (non-blocking). Commit `f071215`'s entire diff is unasserted.**

Grepping the suite for `emptied|repopulat|pruned|staging roots were` returns **nothing**, and `ensure_overlay_tiers_fresh` appears **0 times** in the test file, so neither half of the commit is bound:

- the new pruned-roots wording — which now carries a positive claim, *"nothing outside them was touched"* — has no assertion;
- the hoisted `core` validation is reachable only by internal misuse, and no case exercises it.

I verified both live myself, so they work today. But this ticket's own thesis is that the expensive failures are the silent operator-facing ones, and this message exists solely to stop a destructive step from being silent. A case asserting that a refused rebuild names every root it emptied would cost a few lines on top of case 13's existing fixture, which already produces exactly this state — the earlier `iter:1` review's "no test covers it either way" reasoning applied to a message rather than a tier set.

### Not re-raised

**F2 upheld with no action** — agreed, nothing further. **Iteration 1's finding 3 is closed** by `f071215`. The four `component-resolve.sh` shellcheck notes are unchanged and remain intentional (SC2089/SC2090 are literal quotes in report text; the two SC2086 word-splits at `:534`/`:573` are the mechanism passing three tier names as three arguments — quoting them would break `ensure_overlay_tiers_fresh`). shellcheck is still not wired into CI.

### What could have made this not an approval

1. **The fix not being live.** Four declaration-targeted mutations, each red on exactly the case that binds it — including verbatim the one that used to pass.
2. **Configuration 3 still being a false green.** Falsified independently: corrupting only the fallback substitution splits configuration 1 from configuration 3.
3. **A fifth `(CLI, type)` pair.** Checked every command's dispatch arms and every setup; none exists, and the compiled-basis argument shows why none can.
4. **The F3 message describing something that does not happen.** Reproduced live — root present before, absent after, all three named.
5. **My own instrumentation lying to me again.** It did, twice more. My first probe of Finding 5 reported both guards LIVE — because my capture labels contained the literal words `report_collision` and `report_unresolved`, and I was grepping for `collision` and `unresolved`. Re-run with unlabelled capture, both are inert. Last iteration it was an inert mutation; this time a contaminated sample. Both would have produced a confidently wrong finding, in opposite directions.

hcross:
## Verdict: APPROVE

Iteration 3, reviewed at `9c9b3f7`. **Zero findings.** CI 13/13 green, `mergeStateStatus: CLEAN`. All three of iteration 2's findings are closed, each verified by a mutation that turns the new assertion red. I could not find a third inert assertion, and I could not falsify the rule.

### First: my `:675` finding was wrong, and the falsification is correct

I claimed `:675` was redundant because case 13's `grep -Fq 'acme-review'` would already catch the misreport — reasoning that a refused rebuild reported as an unresolved name would lack the colliding name. Reproduced the tester's mutation (`ensure_overlay_tiers_fresh … || return $?` → `|| true`, so the refusal falls through to `report_unresolved`):

**25 passed, 1 failed** — a single case, `step 3 exit contract`, and the detail line is `a refused rebuild was reported as an unresolved name`. That is `:675`'s message. The `acme-review` arm did **not** fire.

The tester's explanation is exactly right: `ensure_overlay_tiers_fresh` prints the build's own report — which names `acme-review` — to stderr *before* returning, so the two reports coexist rather than substitute. The colliding name is present either way. `:675` is the sole detector for the regression its case names, identical in status to `:309`, not lesser.

My error was reasoning about print order instead of tracing it. I had read that function twice in iteration 1 and still asserted a control-flow property from memory rather than measurement — the same failure mode as the inert mutation in iteration 1 and the contaminated sample in iteration 2, arriving a third time in a third disguise.

### F4 — closed, and the design call was better than my suggestion

Same mutation that exposed it (`COMPONENT_OVERLAY_TIERS` cut to `"library community"`): previously `R20/R2+R5` reported PASS; now **FAIL**, with `claude/skills: the declared served tier set does not include the org tier; it declares: library community`.

**The tester was right to decline my `resolved_by_name` suggestion and I was wrong to make it.** That helper is per-script, and `COMPONENT_OVERLAY_TIERS` lives in `component-resolve.sh` — inside the scan set, not in the command. Reading the anchored assignment across the scan set is the only form that reaches the declaration where it actually lives, and it is symmetric with the fix three lines up.

Trap 5's bounding claim is verified rather than accepted: `$scan` is consumed at exactly two sites, `:1199` (`declared`) and `:1224` (`served`), and both now read a declaration.

### F5 — closed, and the control arm genuinely closes the trap that bit me

Case 13b holds both patterns to the code. Two mutations:

- **Rewording `report_collision`** (`claimed by more than one component` → `taken by several components`, and the second line likewise) → case 13b **red**. The coupling is real.
- **Contaminating the harness** — injecting `echo "harness says: collision unresolved" >&2` into the probe scaffolding, which is precisely my iteration-2 mistake, and deliberately containing both trigger words → the **control arm fired**: `the control capture is not empty, so the harness contributes text to its own samples and neither result below can be trusted`.

That second one is the answer to the question I actually cared about. My contaminated sample matched *because* my labels carried the words; this control fails the case on any harness contribution whatsoever, before the pattern arms are consulted. It is a structural fix, not a carefulness fix.

### F6 — closed by asserting the claim, and live in both directions

- **Prune removed** → `R17` red (plus the residue case).
- **Prune widened** to reach `dist/library/.gemini`, outside the named roots → `R17` red: `the prune reached outside the roots it named: a sentinel under dist/library/.gemini was removed, while the refusal claims nothing outside them was touched`.

Asserting the claim rather than the wording is the right choice — the sentence `nothing outside them was touched` is a positive claim about the filesystem, and it is now bound as one. Passing the `core` arm last, after two valid tiers, is also correct: passed first it would prove only that nothing had been pruned yet.

### The rule — attacked, not falsified

> A positive assertion over real output cannot go inert; a negative guard can.

**I could not break it, and I tried from two directions.**

*Direction 1 — a class the rule does not name: negative **filesystem** assertions* (`[ -e PATH ] && ok=false`), which are inert if `PATH` can never be produced. Nineteen of these exist. The sharpest is case 15's copilot project-tree arm, whose comment calls it decisive. Two realistic regressions:

- dropping the `--tier` filter from the rebuild, so the unfiltered build reaches `core` → **`R6` red**, on that exact arm: `copilot: resolved the core tier and wrote it into the project tree at .github/skills/core-only`;
- routing `core` into the resolution roots → **`R6` red** on the same arm, plus the new `internal contract` case.

That arm is live, and it catches the precise regression `cli-matrix` row 15 says the `--tier` flag exists to prevent.

*Direction 2 — the three `--- Tier:` negative guards (`:531`, `:576`, `:594`)*, which are inert unless a real build emits that string. Verified at the source (`build-components.sh:831`) and live: a real `--target claude --tier library` build emits it once.

**Where the rule is genuinely incomplete — and why that is not a finding.** Inertness has two mechanisms, not one. The rule describes *drift*: the code's wording changes and the pattern stops matching, which is silent for a negative guard and loud for a positive one. It does not describe *an over-broad pattern from birth* — and that is the mechanism behind F4, where `grep -qw org` was a **positive** assertion (it required a match and failed on absence) and was inert anyway, because comments supplied the match.

So the rule alone would not have predicted F4. What makes the file complete is the **conjunction**: trap 6 covers assertions over real output, trap 5 covers assertions over source text and bounds itself to the scan set's two consumers. Both statements of the rule in the file carry the `over real output` qualifier explicitly (`:81`, `:748`), so the artifact is not overclaiming — I checked that specifically before deciding not to raise it. The one durable caveat, worth knowing rather than fixing: **trap 5's completeness rests on that "exactly two consumers" bound**, so a third scan-set consumer added later would land outside both defences. It is exact today; it is an invariant, not a theorem.

### Three configurations, re-established at this head

26/26, exit 0, in each: `/bin/bash` 3.2.57, bash 5.3.15, and `task` genuinely absent.

Configuration 3 falsified again at this head: corrupting only the fallback driver's `NAME` substitution leaves configuration 1 at **26/26 green** and turns configuration 3 **red at 25/26**.

On the tester's self-caught empty-shadow bug — my harness builds its shadow with an explicit `IFS` loop under `set -f` rather than `read -ra`, and now aborts if the shadow holds under 100 entries. It reported **2858 entries**, three non-zero probes, and a toolchain sanity check, so its greens are not the empty-shadow artefact. (My own harness bit me once this pass in a smaller way: the first configuration-3 invocation exited **127** because I had deleted the script during iteration-2 cleanup. An empty grep plus a non-zero exit is easy to skim as "no failures"; it was neither.)

### Everything else

Six guards exit 0 locally (`check-no-machine-paths`, `check-bash32-portability`, `check-test-wiring`, `check-ci-parity`, `check-core-paths`, `check-skill-versions`). Commit scope is the test file alone (+269/−7), so no `cli-matrix` co-maintenance and no version bump are triggered; the suite remains wired into both CI engines. `shellcheck` on the test file reports only the pre-existing `SC2034` on the `pass` loop variable. Worktree clean; all throwaway clones removed.

### What could have made this not an approval

1. **A third inert assertion.** Hunted through a class the rule does not name (nineteen negative filesystem assertions) and through the three `--- Tier:` guards. All live, with mutation evidence I produced.
2. **The F4 fix not reaching the declaration.** Re-ran the exact mutation that exposed it; now red with a precise diagnostic. Trap 5's two-consumer bound verified by enumeration.
3. **The control arm being decorative.** Injected contamination carrying the very trigger words; it fired.
4. **`R17` asserting wording rather than the claim.** Widened the prune past the named roots; it fired.
5. **My own finding standing when it was wrong.** It was wrong; the falsification reproduces exactly, and it is recorded above rather than quietly dropped.

Issue #798, opened 2026-08-10
👷 No CI job runs Bash 3.2, so half the portability regime has no mechanical enforcement


hcross:
## Summary

Spec 0111 established a portability regime for the Bash 3.2 that ships with
macOS. Half of that regime is enforced mechanically; the other half is enforced
by three conventions and a maintainer's memory. **No continuous-integration job
anywhere in this repository runs a Bash 3.2 interpreter**, so a regression in the
unenforced half is green in CI and broken for every macOS contributor.

Surfaced while planning issue #755, which proposes adding two accumulator arrays
to `scripts/check-core-paths.sh` — a change that sits squarely in the unenforced
half. Deliberately kept out of that ticket's scope.

## Evidence

**No 3.2 interpreter exists in CI.** Every runner is Linux:

```sh
$ grep -rh 'runs-on:' .github/workflows/ | sed 's/.*runs-on: *//' | sort | uniq -c
  19 ubuntu-latest

$ grep -rn 'macos' .github/workflows/ .gitlab-ci.yml ci/
(no output)
```

Nineteen jobs, one runner image, zero occurrences of `macos` in any workflow, in
`.gitlab-ci.yml`, or under `ci/`. `ubuntu-latest` ships Bash 5.x.

**The unenforced half is documented as unenforceable by the existing guard.**
`docs/scripting-conventions.md:228-233`:

> A fourth trap is not grep-detectable, so it is deliberately absent from the
> declared set — but it breaks a script just as thoroughly. Under `set -u`, Bash
> 3.2 treats an **empty array** as an unset variable, so `\"${arr[@]}\"` aborts
> when `arr` is empty. Write `${arr[@]+\"${arr[@]}\"}` instead […] This bites
> accumulator arrays hardest, because an accumulator is empty on precisely the
> success path.

So `scripts/check-bash32-portability.sh` cannot substitute: its declared set is
`mapfile`, `readarray` and `declare -A`, and the empty-array trap is excluded by
design because grep cannot see it.

**The asymmetry is measurable in seconds.** A guard iterating an empty
accumulator:

```sh
$ /bin/bash -c 'set -u; A=(); for d in "${A[@]}"; do :; done; echo ok'
A[@]: unbound variable          # 3.2.57, exit 1

$ bash -c 'set -u; A=(); for d in "${A[@]}"; do :; done; echo ok'
ok                               # 5.x, exit 0
```

[...]

PR #799, opened 2026-08-10
🧹 Make the forbidden-construct set have one reader

hcross:
Makes `scripts/check-bash32-portability.sh` the repository's single reader of `ci/bash32-forbidden.txt`: it gains a `--list-constructs` invocation that reports what its own parse loop accepted, and the sync check in `case f` consumes that report instead of forming a second reading. Implements spec 0120 for issue #721.

## How to read this PR?

Two files. Read the guard first.

**`scripts/check-bash32-portability.sh` (+60).** Three additions and one of them is the interesting one. Mode selection sits *before* the repository root is resolved, so an unrecognised argument can never scan the tree. The parse loop gains one accumulator (a newline-joined string, never an array — Bash 3.2 treats an empty array as unset under `set -u`, which is what the file's lines 61-68 have always been about). And the query returns from **a single point, placed after the vacuous-set refusal at 124-127 and before the pattern composition at 129**. That one placement is the whole design: every refusal predicated on the declared set is upstream of it and therefore shared code, and everything downstream is the tree scan the query must not perform — so the two invocations refuse exactly the same declared sets with no duplicated condition.

The header gains a contract block. Read it: it states what is promised (the row set, and therefore the count) and, deliberately, **what is not** — row order, stderr wording, any third field. Naming the non-promises is what keeps a published surface affordable at this size.

**`scripts/tests/test-check-bash32-portability.sh` (+334/-31).** `case f` stops reading the declared set and becomes a `sync_status <repo>` helper that *echoes* one of four tokens — `ok:<n>`, `unaskable`, `vacuous`, `missing:<list>` — called from three sites. Echoing rather than asserting is what lets a negative call site assert a *failure* outcome without incrementing the suite's fail counter. Cases `j` and `k` are new.

The ordering inside `sync_status` is load-bearing rather than stylistic: **exit status is branched on before the row count.** A guard that never ran produces the same empty stdout as a guard that ran and found nothing, so testing rows first would report every "could not be asked" as "reported no construct" — the exact confusion spec 0120 R6 exists to remove, and a failure mode that reading a file does not have.

## How to test this PR?

The suite must run on the shell this ticket is about:

```sh
/bin/bash scripts/tests/test-check-bash32-portability.sh   # 31/31 on 3.2.57
task check-bash32-portability
```

**Read that first command precisely, because it certifies less than it appears to.** `run_check` invokes the script under test as bare `bash` resolved from `$PATH` (line 125), so launching the harness with `/bin/bash` runs the **harness** on 3.2.57 and the **guard** on whatever `bash` comes first — 5.x on a typical machine. The 31/31 is a real result about the suite; it is not on its own evidence that the guard's new code runs on 3.2. That idiom is repo-wide (19 of 68 suites) and is recorded as issue #798; this PR deliberately does not fix it.

The guard's own 3.2 behaviour was therefore established **separately, by invoking it directly**, and all three modes are clean:

```sh
$ /bin/bash scripts/check-bash32-portability.sh
OK: no forbidden Bash 4+ construct in scripts hooks (3 declared construct(s), 151 file(s) scanned).   # exit 0

$ /bin/bash scripts/check-bash32-portability.sh --list-constructs
command<TAB>mapfile
command<TAB>readarray
declare-flag<TAB>A                                                                                    # exit 0

$ /bin/bash scripts/check-bash32-portability.sh --bogus
Error: unrecognised argument(s): --bogus                                                              # exit 2
```

Reproduce those three if you reproduce nothing else — they are the only evidence in this PR that the added code runs on the shell the ticket is about.

Cold review went one better and ran the **whole suite with `$PATH` bash shimmed to 3.2.57 → 31/31**, which certifies both halves at once and is a stronger result than the three commands above. Recorded here rather than left in a review comment, since it is the best evidence this change has. The added code uses no array, so the empty-array trap that motivates #798 does not apply to it; the accumulator is a newline-joined string precisely for that reason.

**The mutation table is the part worth re-running**, because this ticket's whole subject is a test that could pass without checking. Each is a throwaway working-tree edit, reverted immediately:

| # | Mutation | Expected | Observed |
|---|---|---|---|
| 1 | Delete `sync_status`'s exit-status branch | RED | RED 30/31 — `expected 'unaskable' … got 'vacuous'` |
| 2 | Delete both discrimination branches | RED | RED 30/31 — `got 'ok:0'` |
| 3 | Swap the two branch messages | RED | RED 30/31 — `got 'vacuous'` |
| 4 | Evaluate the row count before the exit status | RED | RED 30/31 — `got 'vacuous'` |
| 5 | **Growth** — append a construct plus its Rule 5 row | **GREEN** | **GREEN 31/31** |
| 6 | Emit each row *inside* the parse loop | RED | **RED 30/31** — the leaked rows are named in the failure message |
| 6b | Mutation 6, with the sixth fixture removed | **GREEN** | **GREEN 31/31** |

**Mutations 6 and 6b are a pair and must be read together — they are the record of a finding this PR shipped with and then fixed.** Cold review found that spec 0120's malformed-row scenario has two clauses — *"refuses with the same non-zero outcome **And it reports no construct**"* — and that the suite asserted only the first: `case k` never examined `CHECK_STDOUT`. That clause is the reason the guard accumulates and emits after the loop rather than per row, and this body quoted it as the design's justification while nothing asserted it.

Mutation 6 demonstrates the gap: move the emission inside the parse loop and partial rows leak, with the suite fully green. **6b is why the fix is two parts and not one.** With the `CHECK_STDOUT` assertion present but the sixth fixture removed, the same mutation goes green again — every other fixture is single-row or comment-only, so nothing can leak and the new assertion passes vacuously on all five. The sixth fixture, a valid row *preceding* the malformed one, is the only shape the assertion can fail on. Adding the assertion alone would have been a false green about a false green.

Fixed in `fc385f2`, one file, the suite only — the guard needed no change.

**The assertion count does not move (31/31, delta 0), and that is not a shortfall.** The new fixture joins `case k`'s existing aggregate loop and the clause joins an existing predicate, so neither adds an `ok` call. The visible difference is one relabelled line:

```diff
- PASS  case-k: all 5 malformed declared sets are refused with exit 2 by both requests (R5)
+ PASS  case-k: all 6 malformed declared sets are refused with exit 2 by both requests, reporting no construct (R5)
```

Splitting it into a separate `ok` was considered and declined: the pair 6/6b establishes the discriminating power, and a second `ok` line would add visibility without adding a case the suite can fail on.

Mutation 5 is a different kind and it is not optional. Mutating only toward red proves assertions fail when the contract breaks; it cannot detect **over-coupling**, which is the failure that ships green — an assertion pinned to today's contents passes at merge and breaks on the first legitimate growth of the declared set, for a reason nobody would connect back to this change. Under mutation 5 the two count-bearing assertions *track* the growth (`all 4 declared construct(s)`, `row count (4) … verdict reports (4)`) instead of resisting it.

For mutation 5, use a token disjoint from every fixture's — `growthprobe`, not `coproc`. `coproc` is what `case f`'s third call site uses as the construct deliberately absent from Rule 5; growing the shipped set with it would flip that site to `ok:5`, a red with nothing wrong in the code.

R4 (the verdict path is unchanged) was verified by two independent byte-diffs of stdout, stderr and exit code: pristine-guard-on-pristine-tree versus modified-on-modified, and pristine versus modified guard on the *same* tree, isolating the guard change from the tree change. Both identical. If you reproduce it, pin `CREWRIG_REPO_DIR` on both sides — the guard resolves its root from `dirname "$0"`, so a pristine copy in `/tmp` looks for `/tmp/ci/bash32-forbidden.txt` and exits 2.

Other gates, all green: `check-core-paths.sh`, `test-check-core-paths.sh` (8/8), `check-test-wiring.sh`, `task lint-markdown`, `task spec:lint`. `shellcheck -x` finding set unchanged from the pristine files.

**Added to `scripts/check-bash32-portability.sh`:** a `MODE` selector accepting zero arguments (verdict, unchanged) or exactly `--list-constructs` (query), and rejecting anything else with exit 2 naming the offending argument. Silent tolerance was rejected in plan review: a typo would fall through to a verdict run, which scans the tree and reaches `case f` as empty stdout, where it is misdiagnosed as "reported no construct" rather than "wrong argument". A `LISTING` accumulator is filled inside the existing parse loop and emitted only after it completes — emitting per row would leave partial output before a late malformed row, contradicting spec 0120's scenario "a malformed row is refused on both requests → it reports no construct". A header contract block documents the surface and its non-promises.

**Rewritten in the test suite:** `case f`'s `while read` over `ci/bash32-forbidden.txt` is gone. `run_check` forwards `"$@"` (safe with zero remaining positionals under `set -u` on 3.2.57, so all pre-existing callers are unaffected). Rows are read with a here-document-fed `while IFS= read -r` — **not** `printf … | while read`, which is a pipeline and therefore a subshell, where `checked` and `missing` die at the `done` and the case reports zero rows on a healthy set, a false failure that reads exactly like the anti-vacuity guard firing. Fields are split with `cut -f1`/`cut -f2`, matching `case i`'s existing idiom.

**Row assertions are shape-based, never content-based.** `case j`(a) and (b) and `case k`'s converse assert that every emitted row matches `^(command|declare-flag)<TAB>[A-Za-z0-9_-]+$` with at least one row present — order-free, silent about *which* constructs are declared, therefore growth-proof. `case f`'s first call site additionally cross-checks the query's row count against the verdict's own `(N declared construct(s)` figure; both numbers come from the enforcement, so the declared set is still read exactly once and the test interprets neither its row format nor its contents.

**Requirement coverage.** R1 one interpretation, with the contract published in the guard header. R2 kind and token. R3 verified empirically against a `ci/`-only tree, where the verdict exits 2 at 151-154 while the query exits 0 with rows — direct proof the query returns before target resolution. R4 byte-identical. R5 verified in both directions on all five malformed shapes plus the missing-file case. R6 and R7 now **asserted** rather than preserved by code alone; `case f`'s third call site is the first executable discharge anywhere in the suite of spec 0120's scenario "a declared construct is absent from the written conventions". R8 no declared-set or Rule 5 entry gained or lost. R9 all 22 pre-change assertions present and passing, established by `comm -23` against a pristine copy (left-only set empty; the pristine suite also passes 22/22 against the modified guard). R10 the suite runs to completion on 3.2.57.

**Deliberately not changed.** `bare_expansions_in` and cases `h`/`i`; the guard's own reading rules, anchor, scan scope and verdict exit codes (settled across the seven REVIEW iterations of issue #697); the contents of `ci/bash32-forbidden.txt` and of Rule 5 — the prose restatement there is the duplication `case f` exists to police and is deliberately not derived.

**Blast radius.** No build outputs (nothing under `artifacts/`), no version bump (neither file carries a provenance block), **no CLI-matrix trigger** — re-derived against `docs/cli-matrix-maintenance.md`, whose trigger list names `scripts/build-components.sh` and `scripts/{build,install,setup,import,manage}-*.sh`; `check-*.sh` and `tests/test-*.sh` match neither. All seven downstream invocation sites (`scripting-conventions.yml:64`, `build.yml:277`, `.gitlab-ci.yml:114` and `:227`, `ci-capabilities.yml:130` and `:303`, `Taskfile.yml:355`) pass zero arguments and therefore keep the verdict path R4 freezes; no CI file is edited.

**Process.** Spec 0120 merged as `780b860` (PR #794). The plan needed **five revisions and two REQUEST CHANGES rounds**, both findings produced by prototyping and mutating rather than reading: the R6 discrimination was prescribed as load-bearing and asserted by nothing (three mutations left the suite fully green), and the row assertions were pinned to the declared set's contents. Issue #721 is the logbook per `AGENTS.md` → *Logbook Issues → Rule A* and is closed manually after merge per Rule C — this body carries no closing keyword.

### DIFF
--- a/scripts/check-bash32-portability.sh
+++ b/scripts/check-bash32-portability.sh
@@ -40,15 +40,53 @@
 # included, on the same terms as the scripts it ships.
 #
 # Usage:
-#   bash scripts/check-bash32-portability.sh
+#   bash scripts/check-bash32-portability.sh                  # render a verdict
+#   bash scripts/check-bash32-portability.sh --list-constructs # report what it read
 #
 # Override the repository root with CREWRIG_REPO_DIR (used by the self-test
 # against temporary fixtures), mirroring the sibling check-*.sh guards.
 #
 # Exits 0 when no governed script uses a declared construct, non-zero (naming
 # every offending file and line) otherwise.
+#
+# Contract of the second invocation (spec 0120, requirements 1-3 and 5). It
+# exists so that this guard is the repository's single interpretation of the
+# declared set's row format: a consumer that needs to know what the declared set
+# contains asks the enforcement instead of parsing ci/bash32-forbidden.txt a
+# second time. What is promised:
+#   - Rows on stdout, one per declared construct, in the shape
+#     <kind><TAB><token> — the two fields requirement 2 names, drawn from the
+#     alphabets this script already validates (kind is `command` or
+#     `declare-flag`, token is letters, digits, underscore and hyphen).
+#   - The row *set*, and therefore the row count, is exactly what the parse loop
+#     below accepted. Nothing is scanned and no verdict is rendered.
+#   - Exit 0 with rows; exit 2 for any refusal — every declared set this script
+#     refuses when asked for a verdict is refused here too, and the converse.
+#     Never exit 1: exit 1 is a verdict, and this invocation renders none.
+# What is deliberately NOT promised, so that the surface stays cheap to keep:
+#   - Row order. A consumer that depends on it is depending on file order, which
+#     is not part of this contract.
+#   - The wording of anything written to stderr.
+#   - Any third field. The reason column is free text that may itself hold tabs,
+#     which is the field-boundary ambiguity this invocation exists to remove; it
+#     is not reported.
 
 set -euo pipefail
+
+# Mode selection happens before the repository root is resolved and before
+# anything is read, so an unrecognised argument cannot scan the tree (R3).
+MODE=verdict
+if [ "$#" -gt 0 ]; then
+  # `[ "$#" -gt 0 ]` before touching `$1`: this script runs under `set -u`,
+  # where a bare `$1` with no arguments aborts on `$1: unbound variable`.
+  if [ "$#" -eq 1 ] && [ "$1" = "--list-constructs" ]; then
+    MODE=list
+  else
+    echo "Error: unrecognised argument(s): $*" >&2
+    echo "Usage: bash ${0##*/} [--list-constructs]" >&2
+    exit 2
+  fi
+fi
 
 REPO_DIR="${CREWRIG_REPO_DIR:-"$(cd "$(dirname "$0")/.." && pwd)"}"
 FORBIDDEN="$REPO_DIR/ci/bash32-forbidden.txt"
@@ -66,10 +104,14 @@ fi
 # the failure class this script exists to catch, and it would be invisible to
 # both the grep rule below (not grep-detectable) and CI (which runs Bash 5).
 # `while read` rather than the Bash 4 line-reading builtin, for the same reason.
+# LISTING is the report the second invocation returns: the same rows, joined by
+# newlines rather than held in an array, for the reason above.
 CMD_ALT=""
 FLAG_ALT=""
 declared=0
 TAB="$(printf '\t')"
+NL=$'\n'
+LISTING=""
 
 while IFS= read -r line || [ -n "$line" ]; do
   line="${line%$'\r'}"                       # tolerate CRLF
@@ -119,11 +161,27 @@ while IFS= read -r line || [ -n "$line" ]; do
       ;;
   esac
   declared=$((declared + 1))
+  # Accumulated, never emitted here: a malformed row later in the file must be
+  # refused with no partial report already on stdout, so the report is written
+  # once, below, after the whole file has been accepted.
+  LISTING="${LISTING:+$LISTING$NL}$kind$TAB$token"
 done < "$FORBIDDEN"
 
 if [ "$declared" -eq 0 ]; then
   echo "Error: $FORBIDDEN declares no construct — the guard would pass vacuously." >&2
   exit 2
+fi
+
+# --- Report what was read, when that is what was asked ----------------------
+# Placed here on purpose, and the placement is the whole design: everything
+# upstream of this point is shared with the verdict path — the missing
+# declared-set file, every refusal inside the parse loop, and the vacuous-set
+# refusal above — so both invocations refuse exactly the same declared sets
+# (R5) without a duplicated condition. Everything downstream is the tree scan
+# the report must not perform (R3).
+if [ "$MODE" = list ]; then
+  printf '%s\n' "$LISTING"
+  exit 0
 fi
 
 # --- Compose the match pattern ----------------------------------------------
--- a/scripts/tests/test-check-bash32-portability.sh
+++ b/scripts/tests/test-check-bash32-portability.sh
@@ -19,16 +19,29 @@
 #      (scenario 4, R10).
 #   e. A forbidden construct in a *test* script is rejected on the same terms as
 #      in a shipped script (scenario 6, R5).
-#   f. Every token in ci/bash32-forbidden.txt is named in Rule 5 of
+#   f. Every construct in the declared set is named in Rule 5 of
 #      docs/scripting-conventions.md (R2's "one place both the enforcement and
 #      the documentation refer to"). This is the doc-to-data link that no other
-#      gate covers.
+#      gate covers. Per spec 0120 it no longer parses ci/bash32-forbidden.txt
+#      itself: it consumes the report the enforcement returns for the set it
+#      actually read (`--list-constructs`), through the `sync_status` helper,
+#      which echoes one of `ok:<n>` / `missing:<list>` / `vacuous` / `unaskable`.
+#      Three call sites assert that discrimination — the repository (`ok:<n>`
+#      with n > 0, cross-checked against the verdict run's own count), a
+#      comment-only declared set (`unaskable`), and a set holding a construct
+#      absent from Rule 5 (`missing:<list>`).
 #   g. A declared set with no usable entry fails closed rather than passing
 #      vacuously (docs/scripting-conventions.md Rule 4).
 #   h. No array expansion in the six suites corrected under R6 is unguarded —
 #      the standing half of scenario 5 (R8). See `bare_expansions_in`.
 #   i. That detector is itself probed on 13 fixtures in both directions, so a
 #      future narrowing of it fails here rather than silently going quiet.
+#   j. Asking the guard what it read neither scans the governed tree nor renders
+#      a verdict, and an unrecognised argument is refused rather than falling
+#      through to a verdict run (spec 0120 R3).
+#   k. Both requests refuse the same declared sets and accept the same ones —
+#      case g's five malformed shapes replayed against each, plus the converse on
+#      a well-formed set (spec 0120 R5).
 #
 # Scenario 5 — a corrected suite reports the same verdicts on both shells — is
 # only partly scriptable here: comparing two shells needs two Bash binaries, and
@@ -60,6 +73,17 @@ REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
 DECLARED_SET="$REPO_ROOT/ci/bash32-forbidden.txt"
 CONVENTIONS_DOC="$REPO_ROOT/docs/scripting-conventions.md"
 
+TAB="$(printf '\t')"
+
+# The row shape the guard publishes for `--list-constructs` in its header
+# contract: <kind><TAB><token>, both fields drawn from alphabets the guard
+# itself validates. Cases j and k assert this shape rather than the rows'
+# contents or their order — the contract promises the shape and the row set, and
+# explicitly declines to promise order, so an assertion pinned to a literal
+# block would consume a guarantee that was never given and would turn red on the
+# next legitimate growth of ci/bash32-forbidden.txt (spec 0111 R2's floor).
+ROW_SHAPE='^(command|declare-flag)'"$TAB"'[A-Za-z0-9_-]+$'
+
 if [ ! -f "$SCRIPT_UNDER_TEST" ]; then
   echo "FATAL: cannot find $SCRIPT_UNDER_TEST" >&2
   exit 2
@@ -86,14 +110,19 @@ mk_fixture() {
   cp "$DECLARED_SET" "$1/ci/bash32-forbidden.txt"
 }
 
-# run_check <dir> — run the guard with CREWRIG_REPO_DIR set, capturing stdout,
-# stderr and exit code into CHECK_EXIT / CHECK_STDOUT / CHECK_STDERR.
+# run_check <dir> [arg…] — run the guard with CREWRIG_REPO_DIR set, capturing
+# stdout, stderr and exit code into CHECK_EXIT / CHECK_STDOUT / CHECK_STDERR.
+# Any argument after <dir> is forwarded to the guard, which is how the cases
+# below reach its `--list-constructs` invocation. Verified on 3.2.57: `"$@"`
+# with no remaining positional parameters expands to nothing under `set -u`
+# rather than aborting, so every existing zero-argument caller is unaffected.
 run_check() {
   local repo="$1" out_file err_file
+  shift
   out_file="$(mktemp "$TMP_ROOT/out.XXXXXX")"
   err_file="$(mktemp "$TMP_ROOT/err.XXXXXX")"
   CHECK_EXIT=0
-  ( CREWRIG_REPO_DIR="$repo" bash "$SCRIPT_UNDER_TEST" >"$out_file" 2>"$err_file" ) || CHECK_EXIT=$?
+  ( CREWRIG_REPO_DIR="$repo" bash "$SCRIPT_UNDER_TEST" "$@" >"$out_file" 2>"$err_file" ) || CHECK_EXIT=$?
   CHECK_STDOUT="$(cat "$out_file")"
   CHECK_STDERR="$(cat "$err_file")"
   rm -f "$out_file" "$err_file"
@@ -301,7 +330,80 @@ bare_expansions_in() {
 }
 
 # ---------------------------------------------------------------------------
-# Case f — Declared set and Rule 5 stay in sync (R2).
+# sync_status <repo> — check a repository's declared set against Rule 5, and
+# ECHO which of four outcomes it reached rather than calling ok/bad itself.
+#
+# Echoing a token instead of recording a verdict is what lets a call site assert
+# a *failure* outcome — `unaskable`, `missing:` — without incrementing the
+# suite's fail counter, so "this must fail" is expressible other than by turning
+# the suite red.
+#
+# It obtains the declared set's contents from the enforcement (spec 0120 R1):
+# this suite no longer interprets ci/bash32-forbidden.txt, it reads the rows the
+# guard reports for the set it actually parsed. Reads the caller's `rule5`, which
+# case f assigns before its first call site.
+#
+# The four outcomes, in this evaluation order:
+sync_status() {
+  local repo="$1" missing checked row kind token needle
+  run_check "$repo" --list-constructs
+
+  # `unaskable` FIRST, before anything reads the rows. The order is load-bearing
+  # rather than stylistic: a guard that never ran leaves the same empty stdout as
+  # a guard that ran and reported nothing, so testing the rows first would report
+  # every "could not be asked" as "reported no construct" — the confusion R6
+  # exists to remove, and the failure mode consuming a *program* has that reading
+  # a *file* does not.
+  if [ "$CHECK_EXIT" -ne 0 ]; then
+    echo "unaskable"
+    return 0
+  fi
+
+  missing=""
+  checked=0
+  # Fed by an unquoted here-document rather than `printf … | while read`: the
+  # pipeline is a subshell, so `checked` and `missing` would die at the `done`
+  # and this helper would report `vacuous` on a healthy declared set. And a
+  # `while read` loop rather than the Bash 4 line-reading builtin, because this
+  # file is governed by the very guard it tests (R10) and CI runs Bash 5, where
+  # that mistake is invisible. Same idiom as cases h and i below.
+  while IFS= read -r row || [ -n "$row" ]; do
+    [ -n "$row" ] || continue
+    kind=$(printf '%s' "$row" | cut -f1)
+    token=$(printf '%s' "$row" | cut -f2)
+    [ -n "$token" ] || continue
+    # A bare flag letter would match almost any prose, so a declare-flag entry
+    # is required to appear in its usable form.
+    needle="$token"
+    if [ "$kind" = "declare-flag" ]; then
+      needle="declare -$token"
+    fi
+    checked=$((checked + 1))
+    if ! printf '%s\n' "$rule5" | grep -qF -- "$needle"; then
+      missing="${missing:+$missing, }$needle"
+    fi
+  done <<SYNC_ROWS
+$CHECK_STDOUT
+SYNC_ROWS
+
+  if [ "$checked" -eq 0 ]; then
+    # `vacuous` — the anti-vacuity floor case f has carried since spec 0111,
+    # preserved verbatim in meaning. It is unreachable through the shipped guard:
+    # a declared set holding no construct is already refused with exit 2, which
+    # arrives here as `unaskable`. Kept as defence against a future guard that
+    # stops refusing it, which is why no fixture below reaches it.
+    echo "vacuous"
+  elif [ -n "$missing" ]; then
+    echo "missing:$missing"
+  else
+    echo "ok:$checked"
+  fi
+}
+
+# ---------------------------------------------------------------------------
+# Case f — Declared set and Rule 5 stay in sync (R2), with the declared set's
+#          contents obtained from the enforcement (spec 0120 R1) and the
+#          discrimination R6 demands asserted at three call sites.
 # ---------------------------------------------------------------------------
 {
   # The Rule 5 section only, so a token named under some other rule does not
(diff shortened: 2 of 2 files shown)


hcross:
## Correction to this PR's own body — the 3.2 claim was weaker than it read

I wrote the *How to test this PR?* section citing `/bin/bash scripts/tests/test-check-bash32-portability.sh` → 31/31 as evidence that the change runs on the shell this ticket is about. **That inference does not hold**, and I have rewritten the section rather than leave it.

`run_check` invokes the script under test as a bare `bash` resolved from `$PATH`:

```sh
# scripts/tests/test-check-bash32-portability.sh:125
( CREWRIG_REPO_DIR="$repo" bash "$SCRIPT_UNDER_TEST" "$@" >"$out_file" 2>"$err_file" ) || CHECK_EXIT=$?
```

So launching the harness with `/bin/bash` runs the **harness** on 3.2.57 and the **guard** on whatever `bash` comes first — 5.x here. The 31/31 is a true and useful result about the suite. It is not, on its own, evidence about the guard.

The idiom is repo-wide — 19 of 68 suites, none passing the interpreter through — and is recorded as **#798**. This PR deliberately does not fix it; that would be scope creep into a ticket about a duplicated parser.

**What does establish the claim**, added to the body: invoking the guard directly under 3.2.57. All three modes are clean — verdict `OK: …` exit 0, `--list-constructs` emitting its three rows exit 0, an unrecognised argument exit 2 naming itself. The added code uses no array, so the empty-array trap behind #798 does not reach it; the accumulator is a newline-joined string for exactly that reason.

### Why I am posting this rather than quietly editing

The failure mode is this ticket's own subject, committed in its own pull request description. An assertion whose output is identical whether or not the thing it certifies works carries no information — and "the suite passed on 3.2" reads as coverage of the guard while certifying only the harness.

Two useful distinctions came out of it, both from the sibling ticket #755 where this was first found:

- **Verifying the hazard is not the same as verifying the path to it.** The 3.2 trap was verified; that the prescribed run would reach the guard was not.
- The sentence *reads* as coverage. That is why #798 needs a mechanical check rather than a convention — the next person to write `/bin/bash scripts/tests/test-X.sh` in a plan will be wrong the same way, and nothing will tell them.

Nothing in the diff changes. The correction is to what this PR claims about itself.

hcross:
## Verdict: REQUEST CHANGES

Cold review of `56192a0`, no authoring context. Everything below was executed against a sandbox export of the head tree; the worktree was not modified. Reviewed against the PR body as of `updatedAt` 2026-08-10T22:48:10Z, i.e. **after** the author's self-correction of the 3.2 claim.

The change is well built and the design argument in the body holds up under test: the single-return-point placement really does give both invocations the same refusals, R4 really is byte-identical, and the growth mutation really does track. One property the diff itself names as load-bearing is asserted by nothing, and on a ticket whose subject is "a test that can pass without checking" that is blocking.

---

### CI status

All **13** required checks **pass** on `56192a0` (`build`, `check-agents-size`, `check-ci-parity`, `check-components`, `check-extension-version-bump`, `check-feedback-routing`, `check-skill-versions`, `check-spec-id-reserved`, `gitlab-ci-check`, `grep-anti-patterns`, `lint-markdown`, `lint-specs`, `test-harness-curate`). `check-components` was pending at first poll and has since completed green. No failing or pending job.

---

### Mutation table — re-run, observed

Each mutation applied to a fresh copy of the head tree, suite run with `/bin/bash` (3.2.57).

| # | Mutation | Body claims | **Observed here** |
|---|---|---|---|
| 1 | Delete `sync_status`'s exit-status branch | RED 30/31, `got 'vacuous'` | **RED 30/31** — `expected 'unaskable' … got 'vacuous' (R6)` ✅ |
| 2 | Delete both discrimination branches | RED 30/31, `got 'ok:0'` | **RED 30/31** — `got 'ok:0' (R6)` ✅ |
| 3 | Swap the two branch messages | RED 30/31, `got 'vacuous'` | **RED 30/31** — `got 'vacuous' (R6)` ✅ |
| 4 | Evaluate row count before exit status | RED 30/31, `got 'vacuous'` | **RED 30/31** — `got 'vacuous' (R6)` ✅ |
| 5 | **Growth** — append construct + Rule 5 row | GREEN 31/31 | **GREEN 31/31** ✅ |

All five reproduce exactly as written, including the specific failure text. Mutation 5 used `command<TAB>growthprobe` plus a Rule 5 row, per the body's warning against `coproc`; both count-bearing assertions tracked the growth to 4 rather than resisting it.

I extended the sweep with five more mutations. Three behaved correctly, two stayed green:

| # | Mutation | Result |
|---|---|---|
| 6 | Guard emits each row **inside** the parse loop instead of accumulating | **GREEN 31/31** ← finding below |
| 7 | Delete only the `vacuous` branch (keep `unaskable`) | GREEN 31/31 — disclosed, see *Not findings* |
| 8 | Guard emits every row twice | RED 29/31 — `row count '6' and verdict count '3' disagree` ✅ |
| 9 | Guard omits `declare-flag` rows from the listing | RED 30/31 — `row count '2' and verdict count '3' disagree` ✅ |
| 10 | Guard reverses row order | GREEN — correct, the contract declines to promise order ✅ |

Mutations 8 and 9 are the interesting confirmations: the row-count cross-check at `scripts/tests/test-check-bash32-portability.sh:315-324` is doing real work, and it does it without pinning to the declared set's current size.

---

### Findings

#### 1. The "reports no construct" clause of spec 0120's malformed-row scenario is asserted by nothing

**class: `tech`**

`scripts/check-bash32-portability.sh:167` accumulates into `LISTING` and `:183` emits it once, after the loop. The diff states why, naming the spec clause it discharges:

> A `LISTING` accumulator is filled inside the existing parse loop and emitted only after it completes — emitting per row would leave partial output before a late malformed row, contradicting spec 0120's scenario "a malformed row is refused on both requests → it reports no construct".

Spec 0120's scenario has two `Then`/`And` clauses. Case k asserts the first (same non-zero outcome) and adds a stderr-non-empty check. **It never asserts the second.** `scripts/tests/test-check-bash32-portability.sh:826` reads:

```sh
if [ "$k_verdict_exit" -ne 2 ] || [ "$CHECK_EXIT" -ne 2 ] || [ -z "$CHECK_STDERR" ]; then
```

`CHECK_STDOUT` is never examined on the refusal path. `grep -nE '\-z "\$CHECK_STDOUT"' scripts/tests/test-check-bash32-portability.sh` → no match anywhere in the file.

Demonstrated (mutation 6): move the emission inside the loop, so rows stream as they are accepted. Against a declared set holding two valid rows followed by one with an empty token field:

```
shipped guard:   exit 2, stdout []
mutated guard:   exit 2, stdout [command<TAB>mapfile
                                command<TAB>readarray]
suite:           31/31 passed
```

The guard now violates the scenario it is written to satisfy, and the suite is fully green.

**Two changes are needed, not one.** Adding `[ -z "$CHECK_STDOUT" ]` to `:826` alone would be green-for-the-wrong-reason: all five of case k's fixtures (`:808-816`) and all of case g's (`:461-567`) are single-row or comment-only, so **no existing fixture can produce partial output** — the new assertion would pass vacuously on every one of them. The fix needs a sixth fixture whose malformed row is preceded by at least one valid row, *and* the stdout-emptiness assertion. Verify by re-running mutation 6 and confirming it goes red.

This is the same shape as the two findings the plan review already caught on this ticket — a property prescribed as load-bearing, implemented correctly, asserted by nothing.

---

### Correctness

Verified rather than read:

- **R3 — the query neither scans nor renders a verdict.** Against a `ci/`-only tree the verdict exits 2 while the query exits 0 with rows. Reproduced directly: `/bin/bash scripts/check-bash32-portability.sh --list-constructs` → 3 rows, exit 0.
- **R4 — verdict path byte-identical.** Reproduced on two axes, `CREWRIG_REPO_DIR` pinned on both sides. Pristine guard (`ea97bc4`) vs modified guard on the **same** tree: stdout, stderr and exit code all identical (exit 0, `OK: … (3 declared construct(s), 151 file(s) scanned).`). Repeated on a tree holding a real `mapfile` + `declare -A` violation: identical, exit 1 both sides.
- **R5 — refusal parity.** The single-return-point placement at `:182` sits after the vacuous-set refusal and before pattern composition, so every upstream refusal is shared. Confirmed empirically by case k's five shapes and independently against a hand-built malformed set.
- **R9 — no pre-change assertion lost.** `comm -23` on sorted PASS labels, pristine vs head: base 22, head 31, **left-only set empty**. Matches the body's claim.
- **R10 / issue #798.** On this machine `/bin/bash` is 3.2.57 and `$PATH` bash is 5.3.15, so the trap is live: `run_check` at `:125` invokes the guard as bare `bash`, meaning `/bin/bash <suite>` runs the *harness* on 3.2 and the *guard* on 5.3. I reached the same conclusion independently before reading the author's correction comment of 2026-08-10T22:48:10Z, which discloses it and rewrites *How to test this PR?* accordingly. Reviewing against the corrected body: the claim is now correctly scoped and **the three commands it offers as the actual 3.2 evidence reproduce exactly** — verdict `OK: … (3 declared construct(s), 151 file(s) scanned).` exit 0, `--list-constructs` emitting its three rows exit 0, `--bogus` exit 2 naming itself. I also re-ran the entire suite with `$PATH` bash shimmed to 3.2.57, which the body does not claim: **31/31**, so the guard is clean on 3.2 by the stronger test too. No over-claim remains, and #798 is correctly left out of scope.
- **Bash 3.2 traps.** No `mapfile`/`readarray`/`declare -A` in added lines except inside quoted `printf` fixture strings (`:244`, `:309`, `:317`, `:323`), which is the file's documented idiom and which `task check-bash32-portability` accepts across 151 files. No bare `"${arr[@]}"` added anywhere — the new code uses no arrays. The one array-shaped expansion, `"$@"` at `:125` after `shift`, I verified on 3.2.57 under `set -uo pipefail` with zero remaining positionals: expands to nothing, no abort. `NL=$'\n'` is fine on 3.2.
- **Fixture hazard.** The new `<<SYNC_ROWS` heredoc at `:385-387` has a body of exactly `$CHECK_STDOUT` — one variable, no declared token in command position. That is the same construction as the existing `<<DETECTOR_PROBE` at `:709-711` (body `$i_rows`). Inside existing practice; case b stays green.

---

### Convention compliance

| Rule | Result |
|---|---|
| *Naming Convention* — Gitmoji, no conventional-commit prefix | ✅ `🧹 Make the forbidden-construct set have one reader` |
| *Pull Request Format* — required sections | ✅ *How to read this PR?* / *How to test this PR?* / *Detailed description (for agents)* all present |
| No closing keyword for #721 (Rule C + spec-PR independence) | ✅ `gh pr view 799 --json closingIssuesReferences` → `[]` |
| *Version Bump Convention* | ✅ N/A — neither file carries a provenance block; `check-skill-versions.sh` → `OK: no existing skill/agent sources modified` |
| *CLI Matrix Maintenance* | ✅ Not triggered — no path matches the trigger list; `check-*.sh` and `tests/test-*.sh` are outside `scripts/{build,install,setup,import,manage}-*.sh`. Re-derived independently |
| *Logbook Issues* Rule A/B | ✅ #721 is the logbook, 10 incremental comments, still open |
| Built components | ✅ N/A — nothing under `artifacts/` |
| Downstream call sites unaffected | ✅ All 7 verified passing zero arguments (`scripting-conventions.yml:64`, `build.yml:277`, `.gitlab-ci.yml:114`/`:227`, `ci-capabilities.yml:130`/`:303`, `Taskfile.yml:355`) |

---

### Test coverage

31 assertions, up from 22, none lost. Cases j and k are genuinely shape-based: `ROW_SHAPE` at `:155` is a predicate, not a literal block, and every row assertion pairs it with `rows > 0` rather than a cardinality. I specifically looked for the three failure shapes named in the brief:

- **Literal expected block** — none. The only exact-match assertion is `missing:coproc` at `:353`, and `coproc` is injected by the test's own fixture at `:349`, not read from the declared set.
- **Hardcoded cardinality** — none load-bearing. `k_total -ne 5` at `:832` is a loop-ran-completely self-check, not a pin on the declared set. The count assertion at `:315-324` compares the query against the verdict, both from the enforcement.
- **Row-order dependence** — none; mutation 10 confirms reversing order stays green, correctly, since the contract at `:34-35` declines to promise order.

The gap is the one in *Findings*: the refusal path is asserted on exit code and stderr, never on stdout.

---

### Linter findings

| Gate | Result |
|---|---|
| `task check-bash32-portability` | `OK: no forbidden Bash 4+ construct in scripts hooks (3 declared construct(s), 151 file(s) scanned).` |
| `bash scripts/check-core-paths.sh` | `OK: all 50 strict/adopt-on-edit core-paths entries resolve at HEAD.` (exit 0) |
| `bash scripts/tests/test-check-core-paths.sh` | `Results: 8/8 passed` |
| `bash scripts/check-test-wiring.sh` | `OK: all test scripts are wired (67) or exempted with a reason (1)` (exit 0) |
| `task lint-markdown` | clean |
| `task spec:lint` | `Linting passed!` (168 files) |
| `bash scripts/check-skill-versions.sh` | `OK` (exit 0) |
| `shellcheck -x` (both changed files) | Finding set **identical** to pristine: `16 × SC2016`, `3 × SC2086`, all `info`, all pre-existing. Confirms the body's claim |
| Suite, `/bin/bash` 3.2.57 | `Results: 31/31 passed` |
| Suite, bash 5.3.15 | `Results: 31/31 passed` |
| Suite, `$PATH` bash shimmed to 3.2.57 | `Results: 31/31 passed` |

---

### Not findings (checked, no action)

- **Mutation 7 — the `vacuous` branch is unreachable and unasserted.** True, and the comment at `:390-394` says so plainly. Consequence-free in both directions: delete it and `checked == 0` yields `ok:0`, which call site 1's `ok:[1-9]*` pattern at `:296` still routes to `bad`. Fails closed either way; only message precision is at stake. Disclosed and correctly reasoned — leave it.
- **`missing:coproc` growth coupling.** If `coproc` were ever added to both `ci/bash32-forbidden.txt` and Rule 5, call site 3 would need updating. The body already warns mutation-5 testers about exactly this. Self-consistent fixture choice, failure message would name the cause. No change wanted.
- **`--list-constructs` with a trailing extra argument** is refused by the `[ "$#" -eq 1 ]` condition at `:49` but only `--bogus-argument` is exercised at `:422`. Behaviour is correct; not worth an assertion.

---

**Routing:** one `tech` finding → DEV. The guard needs no change; the fix is a sixth fixture plus a stdout-emptiness assertion in case k of `scripts/tests/test-check-bash32-portability.sh`. Everything else in this PR verified clean, and the R4/R9/mutation-table claims in the body all reproduced exactly as written.

hcross:
## Verdict: APPROVE

Targeted re-review of **`fc385f2`**, the fix for the single `tech` finding raised against `56192a0`. Zero findings of any class. CI green. Both judgement calls put to me are, in my reading, decided correctly — reasoning below rather than assertion.

Scope: I re-derived what the new commit could disturb and spot-checked what it could not. `git diff 56192a0 fc385f2 --name-only` returns `scripts/tests/test-check-bash32-portability.sh` alone, and `diff` on the two guard revisions confirms **`scripts/check-bash32-portability.sh` is byte-identical across the fix** — so R4's byte-identity and the R3/R5 guard-side reasoning carry over untouched rather than needing re-derivation.

---

### CI status

All **13** required checks **pass** on `fc385f2`. `check-components` was pending when the re-review was requested and is now green; I confirmed before verdicting, as asked.

---

### The finding is fixed, and the fix discriminates

Both halves I specified landed, and the vacuity trap I warned about was avoided. I re-ran the pair myself rather than taking the reported result:

| Mutation | Observed here |
|---|---|
| **M6** — guard emits each row inside the parse loop | **RED 30/31**, failing on `valid-rows-then-malformed`, `bad` message carrying the two leaked rows and all four observables ✅ |
| **M6b-full** — same guard mutation, sixth fixture dropped **and** `k_total` lowered to 5 | **GREEN 31/31** ✅ — the fixture is genuinely load-bearing |

That pair is exactly as reported. Three further probes I added:

| Mutation | Observed |
|---|---|
| **M6b-strict** — sixth fixture dropped from the loop list **only**, `k_total` guard left at 6 | **RED 30/31** — `case-k: expected 6 malformed declared sets, ran 5` |
| **M6d** — guard leaks a *single* row early (subtler than wholesale streaming) | **RED 29/31** — caught twice: case k's new clause *and* case f's row-count cross-check |
| **M6e** — guard appends a junk row on the *success* path | **RED 25/31** — shape predicate and count cross-check both fire |

**M6b-strict is a stronger result than the commit claims for itself.** Defeating the new assertion is not one edit but **two coordinated** ones: the `k_total -ne 6` guard at the bottom of case k means deleting the fixture alone turns the suite red and names the reason. Worth recording, because it is the structural protection that makes the "assertion inside a conjunction" question below much less sharp than it first looks.

M6d matters too: the clause does not only catch the wholesale streaming mutation it was written against. A guard leaking one row out of three is caught, and caught by two independent assertions.

---

### Judgement call 1 — scoping the clause to the query run: **correct as decided**

The developer's reasoning holds, and the spec text is the reason.

Spec 0120's scenario reads:

```text
When  the enforcement is asked for the declared set's contents
Then  it refuses with the same non-zero outcome as when asked for a portability verdict
And   it reports no construct
```

The `When` names the **contents** request. Both `Then`/`And` clauses hang off that `When`, so "reports no construct" is scoped to the query by the spec's own grammar. R5, the other candidate source, governs parity of *refusal outcome* — "refused, with the same non-zero outcome" — and says nothing about stdout.

I checked the tempting symmetry empirically: on a refused set the **verdict run's stdout is also empty today** (exit 2, stdout `[]`, one stderr line), so asserting it there would currently pass. That is precisely what makes declining it the right call — it would be an assertion with no requirement behind it, true today by coincidence of implementation, and it would redden case k for any future verdict-side stdout diagnostic that no requirement forbids. That is the over-coupling failure mode this ticket has been correcting throughout; adding it here would reintroduce the pattern in the commit that fixes it. Leave it scoped to the query.

---

### Judgement call 2 — no separate `ok`, count unchanged at 31: **correct as decided, not a finding**

I was invited to call this one and I decline, on a concrete ground rather than deference.

The concern is that a clause hidden inside a conjunction is easy for a future maintainer to delete unnoticed. That is true — I confirmed it: **M6c**, deleting `|| [ -n "$CHECK_STDOUT" ]` while keeping the fixture, gives GREEN 31/31. But a separate `ok` call does not fix it. That second `ok` would be derived from the same conjunction, so it could not fail independently, and it would be deletable by exactly the same one-line edit with exactly the same silent-green result. It relocates the exposure rather than removing it — visibility, not coverage.

What actually protects the clause is what is already there:

- `k_total -ne 6` makes the fixture undeletable in silence (M6b-strict, RED).
- The `ok` label was relabelled to name the clause — `all 6 malformed declared sets are refused with exit 2 by both requests, **reporting no construct**` — so the passing output does surface it, and the label is accurate to what the conjunction asserts.
- The `bad` message prints all four observables including stdout, so a failure says which one broke rather than just that the case failed.

A delta of 0 in the assertion count is the honest number here: the new clause strengthens an existing case rather than adding one the suite can independently fail. Inflating the count would have made the change look larger in the summary line without making it more detectable. Declining was right.

---

### Spot-checks on what the new commit could have disturbed

Re-derived, not taken on trust:

- **Mutation table M1–M5, re-run on `fc385f2`** — all four still RED 30/31 with their original messages (`got 'vacuous'` ×3, `got 'ok:0'`), and M5 growth still **GREEN 31/31**. The sixth fixture is written by the test itself, so it neither depends on nor resists declared-set growth, as claimed.
- **R9** — `comm -23` of sorted PASS labels against pristine `ea97bc4`: base 22, head 31, **left-only set empty**. Nothing lost.
- **Case b** — green, all three assertions. The new fixture is `printf`-written, so the file's command-position discipline is intact and the guard scanning its own test file is unaffected.
- **Sixth fixture shape** — verified rendered: `command<TAB>mapfile`, `command<TAB>readarray`, `command<TAB><TAB>…`. Two accepted rows then the empty-token row; the only one of the six that can leak.
- **Suite** — 31/31 on `/bin/bash` 3.2.57, 31/31 on bash 5.3.15, and 31/31 with `$PATH` bash shimmed to 3.2.57 so the guard runs on 3.2 too (the `#798`-proof run, now in the body).
- **Gates** — `check-bash32-portability` OK (151 files) · `check-core-paths.sh` OK (50) · `test-check-core-paths.sh` 8/8 · `check-test-wiring.sh` OK (67+1) · `check-skill-versions.sh` OK · `lint-markdown` clean · `spec:lint` passed.
- **`shellcheck -x`** — finding set still **identical to pristine**: 16×SC2016, 3×SC2086, all `info`.
- **Conventions** — commit `✅ Assert that a refused declared set reports no construct` is Gitmoji, no conventional-commit prefix · `closingIssuesReferences` still `[]`, so #721 stays open for manual closure under Rule C · three required body sections · no provenance carrier, no version bump owed · CLI matrix still untriggered · PR remains two files.

---

### Note on the adopted evidence

Moving the `$PATH`-shimmed 3.2.57 full-suite run into the body was the right home for it. It is the strongest evidence this change has that the added code runs on the shell the ticket is about, and a review comment is a worse place for it than the PR's own test section — reviewers of the merge commit will read the body, not this thread.

---

**Termination:** APPROVE, **zero findings of any class**, CI green on `fc385f2`. Per `AGENTS.md` → *Retroactive review loop*, the lifecycle's termination criterion is met and this ticket is clear to merge, subject to the user's merge gate.

PR #800, opened 2026-08-10
🐛 Guarantee upstream sync for Antigravity built outputs

hcross:
Adds `.agents/skills` and `.agents/agents` to the core-paths manifest, and — the substance of the change — teaches `scripts/check-core-paths.sh` the direction it was structurally blind to, so the omission cannot recur for the next output root anyone teaches the build to write. Implements spec 0121 for issue #755.

## ⚠️ This is a live behaviour change for existing forks, not a pure addition

**A fork that has hand-edited anything under `.agents/skills` or `.agents/agents` will now have its next `scripts/sync-from-upstream.sh` run halt**, naming those paths, where it previously proceeded without a word. Those two subdirectories, and not the `.agents` root: an edit confined to `.agents/ANTIGRAVITY.md` or `.agents/settings.local.json.example` still does **not** halt, which is the R4 design this body defends further down. Sixty-five of the sixty-seven tracked files under `.agents/` are covered. That is spec 0121 R3 working as specified — a local modification to a built Antigravity output must halt a sync exactly as a local modification to a built Claude Code output already does — but it will look like a regression to whoever meets it first, and it is stated here rather than discovered in the field.

Who it actually reaches: a fork that hand-edited a *built output* was already living dangerously, since `scripts/build-components.sh` regenerates those files from `artifacts/` and `--check` already reports the drift. The halt surfaces a problem that existed silently.

## How to read this PR?

**Start with why CI was green while the bug stood**, because it is the whole point of the change. `scripts/check-core-paths.sh` has run in CI on both forges since June, checking spec 0031 R5: every manifest entry resolves to tracked content at HEAD. That is **manifest → tree**. This issue is **tree → manifest** — **65** tracked files (43 under `.agents/skills`, 22 under `.agents/agents`) that the build writes and the manifest never names. *(An earlier draft said 67. That is the whole `.agents/` root, and it therefore counts the two files this very body argues must **not** be swept in — a figure inherited from the issue text without re-deriving it against the argument it was supporting.)* The guard was certifying the neighbouring property, so its output was identical whether or not the property this issue is about held.

Read in this order:

1. **`.crewrig/core-paths.txt`** — two entries plus a comment block recording why the `.agents` root is not claimed whole. No policy column, matching every sibling and relying on the documented `strict` default; `strict` is not merely conventional here, since R3 requires a local edit to *halt*, which `adopt-on-edit` (preserve, never abort) and `excluded` (never guarded) both fail to do.
2. **`docs/layers.md`** rows 169-170 — R7 requires the human-readable and machine-readable records to agree.
3. **`scripts/check-core-paths.sh`** — the reverse direction. It **derives** the built-output set by reading `scripts/build-components.sh` rather than carrying a copy that can drift. All sixteen output writes funnel through two helpers, `check_or_write` and `propagate_skill_resources`, and every one passes a target literally beginning `"$out_root/`; extracting `$out_root/<a>/<b>` yields exactly nine directories. The bug is *computed*, not asserted, and the granularity falls out with no special-casing.
4. **`scripts/tests/test-check-core-paths.sh`** — cases e through **l**, and one line in `run_check`.

Two details worth pausing on. **A write-helper call site whose *target* yields no directory fails the run**, naming file and line — that fail-closed rule converts *target* indirection from a silent under-report into a loud failure. It does not cover *call-site position*: the rule only sees lines that already matched the helper pattern, so until `1997b99` a call preceded by anything (`if ! …`, `for … do …`, `[ … ] && …`) was discarded before reaching it — three shapes in which the build wrote into an ungoverned directory and the guard reported the repository clean. Found in review, fixed in `1997b99`, pinned by case j. And **`run_check` now passes the harness's interpreter through** (`"${BASH:-bash}"`): without it, `/bin/bash scripts/tests/test-check-core-paths.sh` runs the harness on 3.2.57 and the guard on `$PATH`'s bash, so the 3.2 gate certifies nothing about the guard. That idiom is repo-wide and is issue #798; this PR fixes only the suite it touches.

## How to test this PR?

```sh
bash scripts/check-core-paths.sh
/bin/bash scripts/tests/test-check-core-paths.sh    # 24/24
bash      scripts/tests/test-check-core-paths.sh    # 24/24
```

On the real tree the guard now reports both directions:

```
OK: all 52 strict/adopt-on-edit core-paths entries resolve at HEAD.
OK: all 9 built-output directories derived from scripts/build-components.sh are governed.
```

Revert the two manifest entries and it reproduces the original bug exactly:

```
FAILED: 2 of 9 built-output director(y/ies) derived from
scripts/build-components.sh carry no upstream-sync guarantee:
  - .agents/skills
  - .agents/agents
```

**The mutation table is the evidence, and every row below is observed at `1997b99` rather than predicted.** Each was a throwaway working-tree edit, reverted immediately, with the 24/24 baseline re-confirmed between.

| Mutation | Observed (5.3.15 / 3.2.57) |
|---|---|
| (a) reverse block deleted | **10/24** both shells |
| (b) matcher + comment skip on the raw line instead of the trimmed line | **22/24** both shells — **case l** red |
| (c) 17th write site added to the real build script | guard exits 1 naming `.newcli/skills`; **23/24** both |
| (d1) array guard dropped, **governance** loop | **24/24 on 5.x**, **22/24 on 3.2** |
| (d2) array guard dropped, **dedupe** loop | **24/24 on 5.x**, **15/24 on 3.2** |
| (d3) array guard dropped, ungoverned-report loop | **24/24 on both** — that guard sits behind a `-gt 0` check and is dead code |
| (e) matcher re-anchored to the start of the line | **22/24** both — **case j** red |
| (f) nested-child restriction dropped from the coverage predicate | **23/24** both — **case k** red; the real tree stays green |
| (g) the `#`-comment skip at `check-core-paths.sh:144` removed, so a commented line reaches the helper patterns | **18/24** both |

Rows **(d1)** and **(d2)** justify the `run_check` change: red on 3.2, green on 5.x. Without the interpreter passthrough both are green on both shells — which is what the plan shipped with until cold review caught it. Row **(d3)** is recorded because a green run there would otherwise read as a completed mutation.

Row **(e)** pins the fix this PR gained in review. Until `1997b99` the extractor matched a helper call only as the **first token** of its line, so `if ! check_or_write "$out_root/.newcli/skills/…"`, the same call in a `for` body, and the same call after `[ … ] &&` were each discarded before reaching the fail-closed rule — three shapes where the build wrote into an ungoverned directory and the guard printed `OK: all 9 … are governed`, exit 0, suite fully green. Both stated preconditions held in all three. The fix widens the pattern by two characters; case j is its only pin.

Row **(b)** is where this PR's own history is worth reading, because **widening the matcher had a second-order consequence nobody had asked about**. It made the pattern indentation-insensitive, which silently retired the mutation that had pinned the trim: (b) was 15/19 red before the fix and **green afterwards**, leaving the trim certified by nothing. Case **l** — an indented comment naming a helper must stay a comment rather than parse as a call site — re-pins the one place the trim still decides an outcome, and (b) kills again at 22/24. *Closing a check can retire a neighbouring one, and only re-running the whole table afterwards shows it.*

Rows **(b)** and **(g)** are a pair, and each names the exact line it removes so a reader can reproduce them without this description: (b) drops the trim before `case` matching, (g) drops the `#`-comment skip at `check-core-paths.sh:144`. Between them they cover the two things the trim-and-skip block still decides.

Row **(f)** covers the coverage predicate's excluded-**ancestor** half, which no fixture reached before. Note the direction: the unpinned variant fails **loud** (a false red), and the dangerous too-loose direction is already covered by case h — so this is cheap insurance rather than a closed hole.

**Cases j, k and l are new in `1997b99`**, added with the review findings they pin. Case **i** remains the R6 discriminator for the *derivation*: it copies the **real** build script into a fixture paired with a manifest naming no built output, driving the guard down the **failure** path where it names each directory — the success path prints only a count, so only the failure path pins identities.

**One divergence from the plan's wording, recorded rather than smoothed over.** Step 13(d) predicts "cases a-d MUST abort" under the array-guard mutation; observed is c and d only. Cases a and b assert exit 1 for a *forward*-direction phantom, and the forward direction exits before reaching the reverse block, so they never reach the empty accumulator — they pass for the reason they were written to pass. This follows from placing the reverse *verdict* after the forward loop, which the plan pins for the *enumeration* (step 3) but never for the verdict. The placement is required by the plan's own step 12: case i gives its fixture one trivially-resolvable entry "so the forward direction passes and the non-zero exit is unambiguously the reverse direction's" — a sentence that only means anything if the forward direction runs first. The load-bearing property is unchanged and confirmed by (d1)/(d2). The plan's prediction was imprecise; its requirement is met.

**Requirement coverage.** R1 by construction plus the existing forward guard. R2 the guard is written over the derived class, not over `.agents`. R3 `strict` with no policy column. R4 discharged by omission — `.agents/ANTIGRAVITY.md`, `.agents/settings.local.json.example` and `.agents/settings.local.json` appear only inside comments, never as entries, and `.gitignore:47` keeps the last untracked. R5 the reverse verdict names each ungoverned directory. R6 case e with its discriminating twin case f for the *rejection*, and case **i** for the anti-blinding half — the clause requiring a guard that has stopped detecting the condition to fail the build. Case i is exercised by rows **(a)** and **(g)**, whose red sets are `e g h i j l` and `i l` respectively. **Not** by row (b): (b) carried that role until the matcher was widened, and no longer does — see the paragraph on it above. R7 manifest lines 102-103 against `docs/layers.md` rows 169-170.

**Case design notes from DEV.** Case f asserts exit 0 **only** — an earlier draft also asserted the reverse-direction success line, which would have turned the *control* case red under mutation (a) and destroyed the e-red/f-green contrast the pair exists to create. Case h uses the **nested** carve-out form (`.newcli strict` + `.newcli/skills excluded`), mirroring the real `.github/copilot/settings.json` shape; a top-level `excluded` entry never enters the carve-out branch and would exercise none of the logic mirrored from `path_is_governed()`. Case g asserts non-zero **and** that stderr names `build-components.sh:2`, since "non-zero" alone would be satisfied by an unrelated crash. The 16 call sites were verified by instrumenting the guard, not assumed: lines 447, 448, 517, 518, 549, 550, 582, 583, 620, 630, 660, 679, 737, 752, 770, 785, zero unmatched; instrumentation reverted.

**Plan history.** Three cold-review rounds, four findings. The manifest name set was keyed correctly from the start; what needed three rounds was the *verification*: R6 was not discharged by the original case set (the guard could go blind with the suite at 14/14), the coverage predicate diverged from `path_is_governed()` on an `excluded`-ancestor shape, and the 3.2 gate meant to protect the new accumulators was itself inert. The third is the one worth carrying forward — the plan verified the hazard and never verified the path to it.

**Blast radius.** No build outputs, no `artifacts/` source, no version bump. **No CLI-matrix trigger** — `check-*.sh` and `tests/test-*.sh` are on neither list in `docs/cli-matrix-maintenance.md`, and `check-core-paths.sh` is not named individually in the manifest (only `sync-from-upstream.sh` and `build-docs-index.sh` are named individually — at `:135` and `:138` **at this PR's head**, which is `:125`/`:128` on `main`, the ten-line shift being this PR's own manifest addition; the guard falls under the blanket `scripts` entry). No CI file changes: guard and suite are already wired at `build.yml:107,110`, `.gitlab-ci.yml:59-60`, `ci-capabilities.yml:68-69`, so R6's "every CI run" is satisfied by inheritance. `check-ci-parity.sh` and `check-test-wiring.sh` stay green untouched.

**Known and deliberately not fixed here.** Every other test suite that spawns its subject without propagating the interpreter is issue #798. Measured with Python rather than a `grep` on a `${…}` pattern, because this repository's `grep` is `ugrep` and silently returned 0 on such patterns twice this session: of the **68** suites under `scripts/tests/`, **0 propagate the interpreter on `main`**, and **1 does at this PR's head** — the one suite this PR touches, at `test-check-core-paths.sh:121`. Sweeping the rest belongs to #798; each needs its own two-shell re-run.

Issue #755 is the logbook per `AGENTS.md` → *Logbook Issues → Rule A*, and is closed manually after merge per Rule C — this body carries no closing keyword.

### DIFF
--- a/.crewrig/core-paths.txt
+++ b/.crewrig/core-paths.txt
@@ -91,6 +91,16 @@ artifacts/library
 .github/skills
 .github/agents
 .github/workflows
+# The `.agents` root is deliberately NOT claimed whole (spec 0121 → Out of
+# scope): no CLI output root is, and claiming this one would sweep in
+# `.agents/ANTIGRAVITY.md` and `.agents/settings.local.json.example` — the
+# latter exactly the class of settings template that already needed the
+# `.github/copilot/settings.json` carve-out above. Only the two
+# subdirectories the component build writes are claimed.
+# `.agents/settings.local.json` is untracked (.gitignore) and stays unlisted,
+# so it is never guarded and never restored (spec 0121 R4).
+.agents/skills
+.agents/agents
 
 # Extension distribution channel
 extension-skeleton
--- a/docs/layers.md
+++ b/docs/layers.md
@@ -166,6 +166,8 @@ exist, how org artifacts integrate) is defined in spec 0012 sub-spec E2.
 | `.github/workflows/` | CI/CD pipeline definitions. |
 | `.github/copilot/` | GitHub Copilot workspace configuration. |
 | `.github/copilot/settings.json` | Committed workspace settings, `strict` by default as a member of `.github/copilot/` above — except its `hooks` key, which the transcript-hooks opt-in in `setup-copilot-interactive.sh` deliberately rewrites locally with an absolute path (ADR-0001 Discovery finding #8). Reclassified `excluded`, nested under the strict `.github/copilot/` parent (spec 0097 / issue #605), so that designed-in local mutation no longer aborts `scripts/sync-from-upstream.sh`; sibling members such as `extension.json` remain `strict` and still abort on a local diff. |
+| `.agents/skills/` | Compiled Antigravity CLI skill definitions. |
+| `.agents/agents/` | Compiled Antigravity CLI agent definitions. |
 
 ### Extension distribution channel
 
--- a/scripts/check-core-paths.sh
+++ b/scripts/check-core-paths.sh
@@ -1,13 +1,22 @@
 #!/bin/bash
-# check-core-paths.sh — Reject phantom entries in the core-paths manifest.
+# check-core-paths.sh — Keep the core-paths manifest and the tree in agreement,
+# in BOTH directions.
 #
-# Per spec 0031 (Requirement 5), continuous integration MUST fail a pull
-# request when any entry in .crewrig/core-paths.txt does not resolve to
-# tracked content at the repository HEAD. A "phantom" entry — a manifest
-# line naming a path with no tracked content — makes an adopter's
+# Forward direction — manifest to tree (spec 0031 R5). Continuous integration
+# MUST fail a pull request when any entry in .crewrig/core-paths.txt does not
+# resolve to tracked content at the repository HEAD. A "phantom" entry — a
+# manifest line naming a path with no tracked content — makes an adopter's
 # sync-from-upstream.sh run skip-with-warning at best, and historically
 # aborted the whole sync. This guard catches such an entry before it can
 # reach the canonical branch.
+#
+# Reverse direction — tree to manifest (spec 0121 R5). Continuous integration
+# MUST fail when a directory the component build writes component outputs into
+# carries no upstream-synchronisation guarantee, and MUST name that directory.
+# The forward direction is structurally blind to this: it walks the manifest,
+# so a built-output directory the manifest never mentions is invisible to it.
+# That blindness is why `.agents/skills` and `.agents/agents` sat unguaranteed
+# through a fully green CI (issue #755).
 #
 # Policy semantics (spec 0020) mirror sync-from-upstream.sh:
 #   strict / adopt-on-edit  Upstream-owned content — MUST resolve at HEAD.
@@ -18,11 +27,41 @@
 # second token = policy, default `strict`; blank/`#` lines skipped; CRLF
 # tolerated).
 #
+# Two deliberate scoping decisions in the reverse direction:
+#
+#   1. Depth-2 extraction. Only a `$out_root/<a>/<b>` target yields a
+#      directory, because `<a>/<b>` is the granularity the manifest itself
+#      claims (`.claude/skills`, not `.claude`). A future top-level output —
+#      `"$out_root/AGENTS.md"` — therefore yields nothing and trips the
+#      fail-closed rule below. That is loud by design, not a false positive:
+#      such an output needs a manifest classification decision, and this guard
+#      is where the decision gets forced rather than skipped.
+#   2. Bounded identity with sync-from-upstream.sh's path_is_governed().
+#      dir_is_governed() below mirrors it EXCEPT for the
+#      `.crewrig/.synced-markers` short-circuit (sync-from-upstream.sh:369),
+#      which is the R8 marker-bookkeeping carve-out and can never name a build
+#      output. Everything else — the strict/adopt-on-edit policy filter, the
+#      nested-`excluded`-child carve-out, and continuing the scan rather than
+#      concluding when such a child matches — is the identical rule.
+#
+# Known limitation, named here so it is not rediscovered the hard way: a write
+# site reachable only for a non-`core` tier would demand a manifest entry for a
+# path with no tracked content at HEAD, which the forward direction would then
+# reject as a phantom — an unsatisfiable pair, since output_root_for_tier() in
+# build-components.sh sends non-core tiers to `dist/<tier>`. The escape is to
+# teach build-components.sh to declare its own output directories (e.g. a
+# `--list-output-dirs` query mode) and consume that here instead of parsing
+# call sites; a declaration can distinguish tiers, this extraction cannot.
+#
 # Usage:
 #   bash scripts/check-core-paths.sh
 #
-# Exits 0 if every strict/adopt-on-edit entry resolves at HEAD, non-zero
-# (with a per-entry failure list) otherwise.
+# Exit codes:
+#   0  Both directions agree.
+#   1  A manifest entry does not resolve at HEAD (forward), or a built-output
+#      directory carries no guarantee (reverse). Failing items are listed.
+#   2  A required input is missing or unclassifiable: the manifest, the build
+#      script, or a write-helper call site whose target yields no directory.
 
 set -euo pipefail
 
@@ -52,6 +91,125 @@ while IFS= read -r line || [ -n "$line" ]; do
   PATHS+=("$path")
   POLICIES+=("$policy")
 done < "$MANIFEST"
+
+# ---------------------------------------------------------------------------
+# Reverse direction, part 1 — derive the built-output directory set from the
+# build script itself (spec 0121 R5).
+#
+# Deriving rather than declaring is the whole point: a list held here would
+# drift from the build exactly as the manifest just did, which is the failure
+# this guard exists to end. Every output write in build-components.sh funnels
+# through one of two helpers, and every call site passes a target beginning
+# `"$out_root/`, so the call sites ARE the declaration.
+# ---------------------------------------------------------------------------
+BUILD_SCRIPT="$REPO_DIR/scripts/build-components.sh"
+
+if [ ! -f "$BUILD_SCRIPT" ]; then
+  echo "Error: build script not found: $BUILD_SCRIPT" >&2
+  echo "The reverse direction (spec 0121 R5) derives the built-output directory" >&2
+  echo "set from this file. A rename must be followed here — it must never" >&2
+  echo "silently retire the guard." >&2
+  exit 2
+fi
+
+DERIVED=()
+bs_lineno=0
+while IFS= read -r bs_line || [ -n "$bs_line" ]; do
+  bs_lineno=$((bs_lineno + 1))
+  bs_line="${bs_line%$'\r'}"
+
+  # Match on the TRIMMED line, and this is load-bearing: the real call sites
+  # are indented inside `for`/`if` bodies, so a matcher tested against the raw
+  # line matches only column-0 stubs and derives nothing from the real script.
+  trimmed="${bs_line#"${bs_line%%[![:space:]]*}"}"
+
+  case "$trimmed" in
+    # Comment skip first — it is what excludes the prose mentions of both
+    # helper names. The definition lines (`check_or_write() {`) are excluded
+    # naturally instead: they carry `(`, not a space, after the name.
+    \#*) continue ;;
+    "check_or_write "*|"propagate_skill_resources "*) ;;
+    *) continue ;;
+  esac
+
+  yielded=0
+  rest="$trimmed"
+  while :; do
+    case "$rest" in
+      *'$out_root/'*) ;;
+      *) break ;;
+    esac
+    rest="${rest#*'$out_root/'}"
+
+    seg_a="${rest%%/*}"
+    seg_b="${rest#*/}"
+    seg_b="${seg_b%%/*}"
+
+    # Depth-2 only: both segments must be literal directory names.
+    case "$seg_a" in ''|*'"'*|*'$'*|*/*) continue ;; esac
+    case "$seg_b" in ''|*'"'*|*'$'*|*/*) continue ;; esac
+
+    yielded=1
+    dir="$seg_a/$seg_b"
+    seen=0
+    for d in ${DERIVED[@]+"${DERIVED[@]}"}; do
+      [ "$d" = "$dir" ] && { seen=1; break; }
+    done
+    [ "$seen" -eq 0 ] && DERIVED+=("$dir")
+  done
+
+  # Fail closed. A write-helper call whose target this guard cannot classify is
+  # an output directory it cannot check — indistinguishable, from the outside,
+  # from a repository with nothing to report. Refusing to run is the only
+  # answer that cannot be mistaken for a clean bill of health.
+  if [ "$yielded" -eq 0 ]; then
+    echo "Error: unclassifiable write-helper call site" >&2
+    echo "  $BUILD_SCRIPT:$bs_lineno" >&2
+    echo "  $trimmed" >&2
+    echo "" >&2
+    echo "This guard extracts an output directory from a \$out_root/<a>/<b>" >&2
+    echo "target and found none on this line. Depth-2 is deliberate — see the" >&2
+    echo "header comment of this script. A top-level or tier-conditional output" >&2
+    echo "needs a manifest classification decision before it can be built." >&2
+    exit 2
+  fi
+done < "$BUILD_SCRIPT"
+
+# ---------------------------------------------------------------------------
+# dir_is_governed <dir>
+# Mirror of path_is_governed() in sync-from-upstream.sh (line 366), minus its
+# `.crewrig/.synced-markers` short-circuit — no build output can live there.
+# Governed iff some strict/adopt-on-edit entry covers <dir>, with no `excluded`
+# entry nested strictly under THAT entry covering it. Consulting excluded
+# ancestors, or concluding on the first excluded match instead of continuing
+# the scan, would both be stricter than the sync actually is — and reachable:
+# `.agents excluded` + `.agents/skills strict` is governed for the sync.
+# ---------------------------------------------------------------------------
+dir_is_governed() {
+  local dir="$1" i j gov skip
+  for i in "${!PATHS[@]}"; do
+    case "${POLICIES[$i]}" in
+      strict|adopt-on-edit) ;;
+      *) continue ;;
+    esac
+    gov="${PATHS[$i]}"
+    case "$dir" in
+      "$gov"|"$gov"/*)
+        skip=0
+        for j in "${!PATHS[@]}"; do
+          [ "${POLICIES[$j]}" = "excluded" ] || continue
+          case "${PATHS[$j]}" in "$gov"/*) ;; *) continue ;; esac
+          case "$dir" in
+            "${PATHS[$j]}"|"${PATHS[$j]}"/*) skip=1; break ;;
+          esac
+        done
+        [ "$skip" -eq 1 ] && continue
+        return 0
+        ;;
+    esac
+  done
+  return 1
+}
 
 failures=()
 checked=0
@@ -84,3 +242,30 @@ if [ "${#failures[@]}" -gt 0 ]; then
 fi
 
 echo "OK: all $checked strict/adopt-on-edit core-paths entries resolve at HEAD."
+
+# ---------------------------------------------------------------------------
+# Reverse direction, part 2 — every derived directory must carry a guarantee
+# (spec 0121 R2/R5). No exemption allowlist: R2 admits none, and an allowlist
+# is how this class of guard usually decays.
+# ---------------------------------------------------------------------------
+UNGOVERNED=()
+for d in ${DERIVED[@]+"${DERIVED[@]}"}; do
+  dir_is_governed "$d" || UNGOVERNED+=("$d")
+done
+
+if [ "${#UNGOVERNED[@]}" -gt 0 ]; then
+  echo "" >&2
+  echo "FAILED: ${#UNGOVERNED[@]} of ${#DERIVED[@]} built-output director(y/ies) derived from" >&2
+  echo "scripts/build-components.sh carry no upstream-sync guarantee:" >&2
+  for d in ${UNGOVERNED[@]+"${UNGOVERNED[@]}"}; do
+    echo "  - $d" >&2
+  done
+  echo "" >&2
+  echo "Every directory the component build writes component outputs into must be" >&2
+  echo "governed by .crewrig/core-paths.txt with a strict or adopt-on-edit policy" >&2
+  echo "(spec 0121 R2). Add the entry, and its docs/layers.md row in the same diff" >&2
+  echo "per the co-maintenance rule in AGENTS.md." >&2
+  exit 1
+fi
+
+echo "OK: all ${#DERIVED[@]} built-output directories derived from scripts/build-components.sh are governed."
--- a/scripts/check-core-paths.sh
+++ b/scripts/check-core-paths.sh
@@ -27,7 +27,8 @@
 # second token = policy, default `strict`; blank/`#` lines skipped; CRLF
 # tolerated).
 #
-# Two deliberate scoping decisions in the reverse direction:
+# Three limits on what the reverse direction can see. The first two are
+# deliberate scoping decisions; the third is a precondition it cannot check:
 #
 #   1. Depth-2 extraction. Only a `$out_root/<a>/<b>` target yields a
 #      directory, because `<a>/<b>` is the granularity the manifest itself
@@ -43,6 +44,17 @@
 #      output. Everything else — the strict/adopt-on-edit policy filter, the
 #      nested-`excluded`-child carve-out, and continuing the scan rather than
 #      concluding when such a child matches — is the identical rule.
+#   3. PRECONDITION, not a fact: every output write goes through one of the two
+#      helpers. True at the time of writing and verified then, but nothing here
+#      enforces it, and no matcher can — a call-site parser cannot see a write
+#      that is not a call. A direct redirection,
+#          printf '%s' "$content" > "$out_root/.newcli/skills/$name/SKILL.md"
+#      writes an ungoverned directory and this guard reports the repository
+#      clean. Teaching the build a third write path therefore retires the
+#      reverse direction for whatever that path writes, silently. If one is
+#      ever added, either route it through a helper or move to the
+#      `--list-output-dirs` escape below — which closes this class too, since a
+#      declaration covers every write regardless of how it is spelled.
 #
 # Known limitation, named here so it is not rediscovered the hard way: a write
 # site reachable only for a non-`core` tier would demand a manifest entry for a
@@ -98,9 +110,11 @@ done < "$MANIFEST"
 #
 # Deriving rather than declaring is the whole point: a list held here would
 # drift from the build exactly as the manifest just did, which is the failure
-# this guard exists to end. Every output write in build-components.sh funnels
-# through one of two helpers, and every call site passes a target beginning
-# `"$out_root/`, so the call sites ARE the declaration.
+# this guard exists to end.
+#
+# This works only while every output write goes through one of the two helpers
+# — a precondition, not a property this guard can verify (header, limit 3). So
+# long as it holds, the call sites are the declaration.
 # ---------------------------------------------------------------------------
 BUILD_SCRIPT="$REPO_DIR/scripts/build-components.sh"
 
@@ -128,7 +142,14 @@ while IFS= read -r bs_line || [ -n "$bs_line" ]; do
     # helper names. The definition lines (`check_or_write() {`) are excluded
     # naturally instead: they carry `(`, not a space, after the name.
     \#*) continue ;;
-    "check_or_write "*|"propagate_skill_resources "*) ;;
+    # Leading `*` deliberately: a helper call need not be the first token of
+    # its line. `if ! check_or_write …`, `for … do check_or_write …` and
+    # `[ -n "$x" ] && check_or_write …` all write output, and a prefix-anchored
+    # pattern discarded them here — silently, because the fail-closed rule
+    # below only ever sees lines that already matched. All sixteen call sites
+    # happen to be bare today, which is exactly why measuring the matcher
+    # against the real build script could not surface the gap.
+    *"check_or_write "*|*"propagate_skill_resources "*) ;;
     *) continue ;;
   esac
 
--- a/scripts/tests/test-check-core-paths.sh
+++ b/scripts/tests/test-check-core-paths.sh
@@ -6,14 +6,46 @@
 # not resolve to tracked content at HEAD. This is the parity sibling mandated by
 # the repo convention "every check-*.sh has a test-*.sh".
 #
+# Since spec 0121 the same script also guards the reverse direction — every
+# directory the component build writes component outputs into must carry an
+# upstream-sync guarantee — so cases e-i exercise that half.
+#
 # Cases:
+#   Forward direction (manifest → tree, spec 0031 R5)
 #   a. Phantom strict entry → exit 1, stderr names the failing entry.
 #   b. Phantom adopt-on-edit entry → exit 1, stderr names the failing entry.
 #   c. Fully resolvable manifest → exit 0 with the OK line on stdout.
 #   d. Phantom excluded entry → exit 0 (excluded is org-owned, NOT checked).
 #
+#   Reverse direction (tree → manifest, spec 0121 R5)
+#   e. Built output absent from the manifest → exit 1, stderr names it.
+#   f. Same tree, manifest lists it → exit 0. The control that makes e
+#      informative: it isolates the manifest omission as e's cause.
+#   g. Write-helper call site with no $out_root/ target → non-zero, names the
+#      line (fail-closed: unclassifiable is not the same as nothing to report).
+#   h. Built output covered only by an `excluded` entry → exit 1. `excluded` is
+#      the absence of a guarantee, and R2 admits no directory without one.
+#   i. The repository's REAL build script + a manifest naming no built output →
+#      exit 1 naming `.agents/skills` and `.agents/agents`, 9 derived. The only
+#      case run against the real script's real shape, so the only one a
+#      matcher-blinding change cannot pass.
+#
+# Cases a-d commit a stub `scripts/build-components.sh` with no call sites, so
+# the derived set is empty in half the suite — the accumulator-empty path the
+# array guards in check-core-paths.sh must survive under bash 3.2 `set -u`. The
+# stub is committed, not merely written, so the fixtures hold up if the guard
+# ever reads the build script from HEAD instead of from disk.
+#
+# Case i reads `scripts/build-components.sh`. That coupling is deliberate: a
+# legitimate tenth output directory turns case i red on its count until the
+# literal is updated. That is the forcing function working, not flakiness.
+#
 # Usage:
 #   bash scripts/tests/test-check-core-paths.sh
+#   /bin/bash scripts/tests/test-check-core-paths.sh   # the bash 3.2 gate;
+#       real only because run_check spawns via "${BASH:-bash}". No CI job runs
+#       a 3.2 interpreter, so this local run is the only place the empty-array
+#       trap is caught.
 
 # -e intentionally omitted: pass/fail counters control the harness; adding -e
 # would abort on expected non-zero exits from the script under test.
@@ -74,11 +106,40 @@ run_check() {
   out_file="$(mktemp "$TMP_ROOT/out.XXXXXX")"
   err_file="$(mktemp "$TMP_ROOT/err.XXXXXX")"
   CHECK_EXIT=0
-  ( CREWRIG_REPO_DIR="$repo" bash "$SCRIPT_UNDER_TEST" >"$out_file" 2>"$err_file" ) || CHECK_EXIT=$?
+  # `${BASH:-bash}`, not bare `bash`: bare `bash` resolves from $PATH, so
+  # launching this harness as `/bin/bash test-check-core-paths.sh` would run the
+  # harness on 3.2.57 and the script under test on whatever modern bash $PATH
+  # supplies — making the 3.2 gate below inert. The `:-` default keeps it safe
+  # under `set -u`. (Repo-wide, 19 of 68 suites still carry the bare form; #798.)
+  ( CREWRIG_REPO_DIR="$repo" "${BASH:-bash}" "$SCRIPT_UNDER_TEST" >"$out_file" 2>"$err_file" ) || CHECK_EXIT=$?
   CHECK_STDOUT="$(cat "$out_file")"
   CHECK_STDERR="$(cat "$err_file")"
   rm -f "$out_file" "$err_file"
 }
+
+# ---------------------------------------------------------------------------
+# Stub build scripts. Single-quoted so `$out_root` stays literal.
+#
+# Deliberately written at column 0, unlike the real script's six-space-indented
+# call sites: a matcher that tested the raw line instead of the trimmed one
+# would still match these stubs and only case i would catch it.
+# ---------------------------------------------------------------------------
+
+# No write-helper call site → the derived set is empty.
+BUILD_STUB_NONE='#!/bin/bash
+# Stub build script: no write-helper call sites.
+echo "stub"
+'
+
+# One call site writing into .newcli/skills.
+BUILD_STUB_NEWCLI='#!/bin/bash
+check_or_write "$out_root/.newcli/skills/$name/SKILL.md" "$content" "$source"
+'
+
+# One call site whose target this guard cannot classify (line 2).
+BUILD_STUB_UNCLASSIFIABLE='#!/bin/bash
+check_or_write "$some_other_root/x.md" "$content"
+'
 
 # ---------------------------------------------------------------------------
 # Case a — Phantom strict entry → exit 1, stderr names the failing entry.
@@ -86,7 +147,9 @@ run_check() {
 {
   repo="$(mktemp -d "$TMP_ROOT/repo.XXXXXX")"
   init_git_repo "$repo"
-  make_initial_commit "$repo" "real.txt" "tracked content"
+  make_initial_commit "$repo" \
+    "real.txt" "tracked content" \
+    "scripts/build-components.sh" "$BUILD_STUB_NONE"
   write_manifest "$repo" $'real.txt\tstrict\nphantom.txt\tstrict\n'
 
   run_check "$repo"
@@ -115,7 +178,9 @@ run_check() {
 {
   repo="$(mktemp -d "$TMP_ROOT/repo.XXXXXX")"
   init_git_repo "$repo"
-  make_initial_commit "$repo" "real.txt" "tracked content"
+  make_initial_commit "$repo" \
+    "real.txt" "tracked content" \
+    "scripts/build-components.sh" "$BUILD_STUB_NONE"
   write_manifest "$repo" $'real.txt\tstrict\nphantom.txt\tadopt-on-edit\n'
 
   run_check "$repo"
@@ -146,7 +211,8 @@ run_check() {
   init_git_repo "$repo"
   make_initial_commit "$repo" \
     "real.txt"  "tracked content" \
-    "other.txt" "other tracked content"
+    "other.txt" "other tracked content" \
+    "scripts/build-components.sh" "$BUILD_STUB_NONE"
   write_manifest "$repo" $'real.txt\tstrict\nother.txt\tadopt-on-edit\n'
 
   run_check "$repo"
@@ -177,7 +243,9 @@ run_check() {
 {
   repo="$(mktemp -d "$TMP_ROOT/repo.XXXXXX")"
   init_git_repo "$repo"
-  make_initial_commit "$repo" "real.txt" "tracked content"
+  make_initial_commit "$repo" \
+    "real.txt" "tracked content" \
+    "scripts/build-components.sh" "$BUILD_STUB_NONE"
   # phantom-excluded.txt resolves nowhere at HEAD, but `excluded` is skipped.
   write_manifest "$repo" $'real.txt\tstrict\nphantom-excluded.txt\texcluded\n'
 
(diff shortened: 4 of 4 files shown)


hcross:
## Verdict: REQUEST CHANGES

Cold review, no authoring context. Observed against head `8909b34` (`updatedAt=2026-08-10T23:10:00Z`, unchanged across this review) in a throwaway clone outside the repository. Shells: `/bin/bash` 3.2.57(1) arm64-apple-darwin25, `bash` 5.3.15.

One blocking finding. It is not in the manifest entries, the docs row, or the mutation table — all of which I re-derived and reproduced exactly. It is in the extractor's **call-site recognition**, one layer beneath the target parsing the fail-closed rule protects.

---

### CI status

All 13 required checks **pass** at `8909b34`. `check-components` was `pending` when I began and resolved to `pass` before this verdict; no check is failing or pending now.

`build`, `check-agents-size`, `check-ci-parity`, `check-components`, `check-extension-version-bump`, `check-feedback-routing`, `check-skill-versions`, `check-spec-id-reserved`, `gitlab-ci-check`, `grep-anti-patterns`, `lint-markdown`, `lint-specs`, `test-harness-curate`.

---

### Correctness

**The derived set re-derived independently — the PR's arithmetic is exact.** Every `$out_root` occurrence in `scripts/build-components.sh` is an assignment (400-401, 593-594, 689-690), a comment (556), or one of exactly 16 helper call sites: 447, 448, 517, 518, 549, 550, 582, 583, 620, 630, 660, 679, 737, 752, 770, 785 — the PR's list, line for line. Depth-2 extraction over those yields exactly 9 unique directories (`.gemini/skills`, `.gemini/commands`, `.gemini/agents`, `.claude/skills`, `.claude/agents`, `.github/skills`, `.github/agents`, `.agents/skills`, `.agents/agents`). Grepping for every tree-write primitive in the build script confirms the only writes are `check_or_write`'s `mkdir -p`/`echo >` at `build-components.sh:288-289` and `propagate_skill_resources`'s `mkdir -p`/`cp` at `321-322`. **No write bypasses the two helpers today.** Baseline on the real tree: `OK: all 52 …` / `OK: all 9 …`, exit 0; the 52 is confirmed independently (59 non-comment entries, 7 `excluded`).

**The coverage predicate matches the sync.** `dir_is_governed()` (`check-core-paths.sh:188-212`) is line-for-line the semantics of `path_is_governed()` (`sync-from-upstream.sh:366`) composed with `excluded_children_of()` (`:140`): the strict/adopt-on-edit filter, the `"$parent"/*` restriction that makes only excluded **children of the matched entry** disqualify, the `"$excl"|"$excl"/*` inner match, and `continue` rather than `return 1` on a carve-out hit. The cited line numbers are exact — `path_is_governed()` is at `:366` and the `.crewrig/.synced-markers` short-circuit it omits is at `:369`. On the shape a prior review round got wrong (`.agents excluded` + `.agents/skills strict`), the shipped code returns **governed**, agreeing with the sync. Correct.

**The tier caveat is recorded, and the record is accurate.** `output_root_for_tier()` sends `core` to `$REPO_DIR` but other tiers to `$CHECK_STAGING_ROOT/$tier` or `$REPO_DIR/dist/$tier`; the extraction is sound only because the same 16 lines serve every tier. The header's "Known limitation" paragraph states this and correctly derives the unsatisfiable pair a tier-conditional write site would create against the forward direction. Nothing to add.

**`run_check` passes the interpreter through exactly once, and the gate is load-bearing.** `test-check-core-paths.sh:114` is the only spawn of `$SCRIPT_UNDER_TEST`; no second path bypasses `"${BASH:-bash}"`. Verified by counterfactual: reverting `:114` to bare `bash` turns mutations (d1) **and** (d2) into **19/19 on `/bin/bash`** — the 3.2 gate would certify nothing. The change is justified.

**Empty-array expansions.** All three expansions of the two new accumulators carry the `${arr[@]+"${arr[@]}"}` guard (`:155`, `:252`, `:260`). `${#DERIVED[@]}` / `${#UNGOVERNED[@]}` are counts and safe on 3.2 under `set -u` — demonstrated by cases a-d passing on 3.2 with an empty derived set.

**R4 discharged by omission — verified.** `.crewrig/core-paths.txt` names `.agents/ANTIGRAVITY.md` only at line 96 and `.agents/settings.local.json` only at line 100, both inside the comment block; the only non-comment `.agents` lines are `102` and `103`. `git check-ignore -v` confirms `.gitignore:47` keeps `.agents/settings.local.json` untracked, and `git ls-tree HEAD` confirms it absent while the other two are tracked. Exact.

**R7 — the two records agree.** Manifest `102-103` (`.agents/skills`, `.agents/agents`) against `docs/layers.md` `169-170` (`` `.agents/skills/` ``, `` `.agents/agents/` ``). Both updated in this diff, satisfying *Core-paths manifest co-maintenance*.

**The plan divergence disclosed in the body is accurate.** Under (d1) I observe red on **c and d only**, exactly as disclosed: cases a and b assert exit 1 for a forward-direction phantom and the forward loop exits before line 252 is reached. The verdict's placement after the forward loop is correct — the enumeration is before it (`:95-212` precedes `failures=()` at `:214`), which is what the plan pins, and the verdict must follow it for case i's "the non-zero exit is unambiguously the reverse direction's" to hold. Its only cost is that a PR failing both directions needs two round trips to see both; acceptable for a CI guard, and worth no change.

**Case f asserting exit 0 only is the right call.** Its job is to differ from case e in exactly one variable — the manifest entry — so that "e red, f green" identifies the omission as e's cause. Adding the reverse success line would make f a second detector of the reverse block rather than a control. Sound.

**(d3) is dead code, and shipping it is right.** Confirmed: `:260` sits behind `[ "${#UNGOVERNED[@]}" -gt 0 ]` at `:256` and can never expand empty, so dropping its guard is 19/19 on both shells. Keep it — it costs nothing, it is uniform with the two live guards, and the day someone hoists that loop out of the `-gt 0` branch the guard is already correct. Recording the green row rather than quietly omitting it is the right disclosure.

---

#### Finding 1 — BLOCKING — the extractor only sees a helper call that is the *first token* of its line, so a call to a new ungoverned output root under `if` / `for` / `&&` is silently skipped

`class: tech`

`scripts/check-core-paths.sh:131-132` matches the trimmed line with a **prefix** pattern:

```sh
    "check_or_write "*|"propagate_skill_resources "*) ;;
    *) continue ;;
```

and the fail-closed rule at `:165-175` only fires for lines that already matched. A line where anything precedes the helper name never reaches it — it is discarded by `*) continue ;;` at `:132` with no diagnostic.

That is not the documented depth-2 case. Both stated preconditions hold for these shapes: the write goes **through** `check_or_write`, and the target begins literally `"$out_root/`. Injecting a 17th write site into the real build script, each writing to an ungoverned `.newcli/skills`:

| Injected 17th write site | Guard | Suite |
|---|---|---|
| `check_or_write "$target_path" …` (target via variable) | **exit 2**, names `build-components.sh:583` | 15/19 |
| `if ! check_or_write "$out_root/.newcli/skills/$name/SKILL.md" …; then :; fi` | **exit 0** — `OK: all 9 built-output directories … are governed` | **19/19** |
| `for x in a; do check_or_write "$out_root/.newcli/skills/$name/SKILL.md" …; done` | **exit 0** — `OK: all 9 …` | **19/19** |
| `[ -n "$c" ] && check_or_write "$out_root/.newcli/skills/$name/SKILL.md" …` | **exit 0** — `OK: all 9 …` | **19/19** |

Three shapes where the build writes into a directory carrying no guarantee and CI reports the repository clean. Spec 0121 R5 requires the build to **fail** whenever that holds; R2 admits no such directory. This is #755's failure mode one level down — the guard's output is identical whether or not the property holds — and it contradicts the script's own rationale at `:161-164`: *"Refusing to run is the only answer that cannot be mistaken for a clean bill of health."* The PR body's claim that the fail-closed rule *"converts indirection blindness from a silent under-report into a loud failure"* is true for **target** indirection (row 1) and false for call-site position (rows 2-4).

The fix is two characters — anchor the match to the helper name as a word rather than to the start of the line:

```sh
    *"check_or_write "*|*"propagate_skill_resources "*) ;;
```

I prototyped it. It closes all three shapes (each becomes `FAILED: 1 of 10 … - .newcli/skills`, exit 1), preserves the target-indirection fail-closed (still exit 2 naming file and line), and regresses nothing: the real tree still derives 9 and exits 0, and the suite is still **19/19 on both `/bin/bash` 3.2.57 and `bash` 5.3.15**. No false positive is reachable in the build script either — the only lines mentioning a helper are the 16 call sites plus the two definitions, and definitions carry `(` rather than a space after the name while comments are already skipped first, both verified by grep.

Please also pin it with a case: a stub whose call site is indented behind `if !`, asserting exit 1 and that stderr names the directory. Without it the matcher can narrow again silently — no existing case exercises a non-leading call, since case i runs the real script where all 16 are bare calls.

---

#### Finding 2 — non-blocking — the "all writes funnel through two helpers" premise is stated as fact, not as a precondition whose violation silently retires the guard

`class: tech`

`check-core-paths.sh:101-103` states *"Every output write in build-components.sh funnels through one of two helpers … so the call sites ARE the declaration."* I verified that is true at `8909b34`. But it is load-bearing, not incidental: a direct write added later is invisible.

```
printf "%s" "$antigravity_content" > "$out_root/.newcli/skills/$name/SKILL.md"
  → guard exit 0, "OK: all 9 built-output directories … are governed", suite 19/19
```

Unlike Finding 1 this is not fixable by a better matcher — a call-site parser cannot see a write that is not a call — so the remedy is disclosure, not code. The header already documents the depth-2 rule and the tier limitation in exactly this register; add a third bullet naming this one, so the next maintainer knows that introducing a third write path retires the reverse direction for it. The `--list-output-dirs` escape already sketched at `:49-54` closes this class too, which is worth saying at the same spot.

---

#### Finding 3 — non-blocking — the excluded-*ancestor* half of `dir_is_governed()` is pinned by no test, so the divergence a prior round caught can return silently

`class: tech`

Deleting the nested-child restriction at `check-core-paths.sh:201` — `case "${PATHS[$j]}" in "$gov"/*) ;; *) continue ;; esac` — makes excluded **ancestors** disqualify a governed child, which is precisely the divergence from `path_is_governed()` an earlier plan revision shipped. The suite stays **19/19 on both shells** and the real tree stays green.

Severity is limited by an asymmetry worth stating plainly: the *dangerous* direction is covered. Making the predicate too **loose** (dropping the carve-out so an `excluded` child no longer disqualifies) is caught by case h, and that is the direction producing a false green. The unpinned direction is too **strict**, whose failure mode is a loud false red rather than a silent pass. Hence non-blocking — but the ticket spent a whole review round on this exact predicate, and one case would keep it: `.newcli excluded` + `.newcli/skills strict` → exit 0, the inverse of case h's nesting.

---

#### Finding 4 — non-blocking — "67 tracked files" in the PR body is the count of a root the same body says is not claimed

`class: tech`

The body's *How to read this PR?* says *"67 tracked files under two directories the build writes and the manifest never names."* At `8909b34`:

```
.agents/skills   43
.agents/agents   22
              -> 65 under the two directories
.agents (whole)  67   = 65 + ANTIGRAVITY.md + settings.local.json.example
```

67 is the whole `.agents/` root — including exactly the two files this PR argues at length must **not** be swept in. The correct figure for the claim is 65. No code impact; it is the one number in an otherwise scrupulously-grounded body that does not trace to what it describes.

---

### Convention compliance

| Rule | Result |
|---|---|
| *Naming Convention* — Gitmoji, no conventional-commit prefix | ✅ `🐛 Guarantee upstream sync for Antigravity built outputs` |
| Branch naming `<prefix>/<NNNN>-<slug>` | ✅ `fix/0755-antigravity-outputs-in-core-paths` |
| *Pull Request Format* — required sections | ✅ *How to read this PR?*, *How to test this PR?*, *Detailed description (for agents)* all present |
| *Logbook Issues* Rule A — the feature issue IS the logbook | ✅ #755, `bug` label, no separate logbook issue |
| Rule B — incremental | ✅ 9 comments interleaved with the work: two logbook entries, PLAN v1/v2/v3, three cold PLAN reviews |
| **No closing keyword for #755** | ✅ `gh pr view 800 --json closingIssuesReferences` → `[]`; Rule C close stays manual |
| *Version Bump Convention* | ✅ N/A — no skill/agent source under `artifacts/` touched |
| *CLI Matrix Maintenance* | ✅ Not triggered. `scripts/check-*.sh` and `scripts/tests/test-*.sh` are on no trigger list; the list names `scripts/build-components.sh` and `scripts/{build,install,setup,import,manage}-*.sh`, and `check-` matches none |
| *Core-paths manifest co-maintenance* | ✅ `docs/layers.md` and `.crewrig/core-paths.txt` in the same diff |
| Built components staged | ✅ N/A — `artifacts/` untouched |
| Executable bits | ✅ both scripts `100755` |
| Policy column | ✅ omitted, so `strict` by default — and the body's argument is right: R3 requires a local edit to *halt*, which neither `adopt-on-edit` nor `excluded` does |

The behaviour-change warning at the top of the body is the correct call. R3 mandates the halt, and a fork that hand-edited a built output will read it as a regression; stating it in the PR beats discovering it in the field.

---

### Test coverage

**Every row of the PR's mutation table reproduces exactly.** Each mutation was a working-tree edit in an isolated clone, reverted and byte-compared against a pristine copy, with the 19/19 baseline re-confirmed between.

| Mutation | PR claim | Observed |
|---|---|---|
| (a) reverse block deleted | 9/19 both; e, g, h, i red | ✅ **9/19** both shells, red `e g h i` |
| (b) matcher on the raw line | 15/19 both; case i red; repo root `OK: all 0 …` | ✅ **15/19** both shells, only case i red (its 4 assertions); repo root printed `OK: all 0 built-output directories …`, **exit 0** — the silent false green |
| (c) 17th write site | exits 1 naming `.newcli/skills`; 18/19 | ✅ `FAILED: 1 of 10 … - .newcli/skills`, exit 1; **18/19** both shells, only case i's count assertion red |
| (d1) array guard dropped, governance loop | 17/19 on 3.2, 19/19 on 5.x | ✅ **17/19** on 3.2 (red `c d`), **19/19** on 5.3 |
| (d2) array guard dropped, dedupe loop | 13/19 on 3.2, 19/19 on 5.x | ✅ **13/19** on 3.2 (red `e f h i`), **19/19** on 5.3 |
| (d3) array guard dropped, report loop | 19/19 on both — dead code | ✅ **19/19** both shells; dead behind `-gt 0` at `:256` |

Row (b) is the discriminating one for R6 and it holds: the matcher-blinding change drives the repo-root run to a vacuous `all 0` pass, and **case i is the only thing that goes red**. R6 is genuinely discharged against that mutation. The e/f pair, case g's `build-components.sh:2` assertion, case h's nested carve-out, and case i's failure-path identity pinning are all well-chosen — case i in particular is right to assert identities on the failure path rather than a count on the success path.

**Beyond the table**, two shapes are unpinned. Finding 1's non-leading call sites (19/19 under three separate injections) is the blocking one. Finding 3's excluded-ancestor predicate is the other. One more, too minor to be a finding: deleting the reverse-direction success line at `:271` leaves **19/19 on both shells** — no case asserts it. Harmless, since exit codes are what R5 rides on and (b) is caught by case i regardless, but worth knowing it is unpinned.

`bash scripts/check-test-wiring.sh` → `OK: all test scripts are wired (67) or exempted with a reason (1)`.

---

### Linter findings

`bash scripts/check-ci-parity.sh`

```
OK: reference, GitHub Actions, GitLab agree on the portable capability set; every pipeline job is traceable.
```

`bash scripts/check-test-wiring.sh`

```
OK: all test scripts are wired (67) or exempted with a reason (1); the exemption allowlist is honest.
```

`bash scripts/check-bash32-portability.sh`

```
OK: no forbidden Bash 4+ construct in scripts hooks (3 declared construct(s), 151 file(s) scanned).
```

As the body notes, this guard is grep-based and cannot see an unguarded empty-array expansion — which is why the (d1)/(d2) two-shell runs, not this script, are the evidence for the accumulators.

`markdownlint docs/layers.md` → exit 0, no output.

`shellcheck scripts/check-core-paths.sh scripts/tests/test-check-core-paths.sh` → exit 1, four **SC2016 (info)** hits only, at `check-core-paths.sh:139,142` and `test-check-core-paths.sh:135,140`. All four are deliberate: the literal `$out_root/` the extractor matches on, and the single-quoted stub heredocs that must keep `$out_root` unexpanded. Nothing to change; no CI job gates on shellcheck.

---

### Summary

The substance of this change is right, and the evidence behind it is unusually honest — the mutation table reproduces row for row, the two-shell counterfactual confirms the `run_check` fix is load-bearing rather than cosmetic, the coverage predicate matches `path_is_governed()` including the shape a prior round got wrong, the tier limitation is documented before anyone had to hit it, and the plan divergence is disclosed accurately rather than smoothed over.

The blocking finding is narrow and its fix is two characters plus one test case. It matters because it is the same shape as the bug being fixed: the guard fails closed on the indirection it recognises and silently open on the indirection it does not, so `OK: all 9 … are governed` reads identically whether or not R2 holds. Close Finding 1 and this is an APPROVE; Findings 2-4 are yours to take or leave, though 3 costs one case for a predicate that already consumed a review round.

hcross:
## Gate provenance — every check re-run against merged content

This branch is 2 commits behind `main` (`51e2b94`). One of those commits, [#799](https://github.com/crewrig/crewrig/pull/799) (`9cba467`), **rewrote `scripts/check-bash32-portability.sh`** — a gate this PR's DEV stage ran and reported green, twice, before that rewrite landed.

That green certified the *old* reader's opinion of these files, not the reader CI will run after merge. Nothing in the guard's output said so. Recording the re-verification here rather than leaving an APPROVE resting on a gate run against a superseded guard.

**A green is scoped to the guard's revision as much as to the tree's.** The disciplines applied throughout this PR — mutate the code under test, verify the path to the hazard, assert the property rather than a proxy — all silently assume the guard you ran is the guard that will run.

### Method

`1997b99` merged with `crewrig/main` at `51e2b94` in a throwaway worktree state (clean merge, no conflicts), every gate re-run against the merged tree, then `git reset --hard 1997b99`. **The pushed head is unchanged and byte-identical to what is under review**; local and remote both at `1997b99`. Re-derivable with `git merge crewrig/main` from `1997b99`.

### Results against merged content

| Gate | Result |
|---|---|
| `scripts/check-bash32-portability.sh` — **main's new version** | exit 0 — `no forbidden Bash 4+ construct in scripts hooks (3 declared construct(s), 151 file(s) scanned)` |
| `scripts/tests/test-check-bash32-portability.sh` — main's new suite | exit 0 |
| `scripts/check-core-paths.sh` | exit 0 — 52 forward entries, 9 built-output directories governed |
| `scripts/tests/test-check-core-paths.sh` on `bash` 5.3.15 | **24/24** |
| `scripts/tests/test-check-core-paths.sh` on `/bin/bash` 3.2.57 | **24/24** |
| `scripts/check-test-wiring.sh` | exit 0 |
| `scripts/check-ci-parity.sh` | exit 0 |
| `scripts/build-components.sh --check` | exit 0 — no drift |
| `task lint-markdown` | exit 0 |
| `task spec:lint` | exit 0 |

Exit codes captured directly, not inferred from an `OK:` line — one gate was first measured through a pipe, where `$?` reports `tail` rather than the script, and was re-run plainly before being quoted here.

The forbidden-construct set itself is unchanged by #799 (still `mapfile`, `readarray`, `declare -A`); only the reader was rewritten. So the risk was that the new reader judged these files differently, not that the rules moved. It does not.

### Merge sequencing

Branch protection rejects a merge from behind `main` — it surfaces as *base branch policy prohibits the merge* with every check green, which reads like a failure and is not one. `gh pr update-branch` is the answer, never `--admin`. The update is deliberately **not** pushed: moving the head mid-review would retarget the reviewer at a SHA they have not read. Order is re-review at `1997b99` → `gh pr update-branch` → merge.

hcross:
## Verdict: REQUEST CHANGES

Re-review of `1997b99` (`updatedAt=2026-08-10T23:36:09Z`), targeted per the brief: fresh clone at the new head, and I did not re-spend on what `8909b34` already settled and the fix commit does not touch — `git diff --name-only 8909b34 1997b99` is exactly `scripts/check-core-paths.sh` and `scripts/tests/test-check-core-paths.sh`, so the manifest, `docs/layers.md`, R4, R7, the 52 forward entries and the tier caveat carry forward untouched.

**The code is correct and complete.** All three findings are closed, the second-order consequence was caught correctly and its remedy holds, and I found no new defect in the guard or the suite. The one blocking finding is in the **PR body**, which still documents `8909b34` — including a mutation row that is now false about the shipped code.

---

### CI status

All 13 required checks **pass** at `1997b99`. `check-components` was `pending` at my preflight on the new head and resolved to `pass`; nothing is failing or pending now.

---

### Correctness

**Finding 1 — closed. All five shapes re-observed at `1997b99`, injected as a 17th write site into the real build script:**

| Injected shape | Guard |
|---|---|
| `if ! check_or_write "$out_root/.newcli/skills/…"; then` | **exit 1**, `FAILED: 1 of 10`, names `- .newcli/skills` |
| `for x in a; do check_or_write …; done` | **exit 1**, same |
| `[ -n "$c" ] && check_or_write …` | **exit 1**, same |
| target via intermediate variable (control) | **exit 2**, unclassifiable — preserved |
| direct redirection (F2's class) | **exit 0** — still blind, by construction |

The fix is the two-character widening at `check-core-paths.sh:152`. Mutation (e) reverting it gives **22/24 on both shells, red on case j alone** — the pin is real and it is the only pin, exactly as reported.

**The widened matcher introduces no false positive against the real build script.** I instrumented the guard to print its derived set: still exactly the 9 directories, byte-identical to `8909b34`, guard exit 0. The 16 call sites are unchanged, and the claim about what else could match holds — the only helper mentions in `build-components.sh` are those 16 lines, the two definitions (which carry `(`, not a space, so the widened pattern misses them), and comments, skipped first at `:144`.

#### The case-l question — answered: yes, case l genuinely re-pins the trim

This was the right thing to attack, and it survives. I enumerated every use of `$trimmed` at the new head: `:140` (the `case`), `:157` (`rest="$trimmed"`), `:189` (the diagnostic echo). Taking them in turn:

- **`:157` is whitespace-immaterial.** Leading blanks cannot affect `*'$out_root/'*` matching or the `%%/*` / `#*/` segment extraction. No outcome depends on the trim here.
- **`:189` is cosmetic** — it only formats the offending line in the fail-closed message.
- **`:140` is the whole remaining surface, and within it only the comment skip at `:144`.** The helper patterns at `:152` are now indentation-insensitive by construction, which is precisely why step 13(b) stopped discriminating.

So the trim has exactly **one** load-bearing role left, and case l is aimed at it. Confirmed by observation: mutation (b) gives **22/24 on both shells, red on case l alone**, and case l's failure is specifically `expected exit 1, got 2` with stderr `Error: unclassifiable write-helper call site … build-components.sh:3` — the comment being parsed as a call site, which is the exact mechanism. Nothing else in the suite notices; case i does not, because the real script's indented comment at `build-components.sh:625` ends *at* the helper name with no trailing space and therefore never matches the widened pattern. The test's own comment says this, and it is accurate — I verified there is no indented comment in the real build script with a helper name followed by a space.

One sub-mode of trim removal that case l does not cover, and does not need to. Case l's fixture comment yields no directory, so trim removal there produces exit 2. A commented-out call site that *does* carry `$out_root/` — realistically, someone disabling a write site during debugging — behaves differently:

```
# indented `# check_or_write "$out_root/.newcli/skills/$name/SKILL.md"` above a live call site
WITH trim (shipped):  OK: all 9 built-output directories … are governed   (correctly ignored)
trim REMOVED:         exit 1, FAILED naming a phantom .newcli/skills      (false positive)
```

Both trim-removal modes therefore fail **loud** — exit 2 in one, a false red demanding a manifest entry in the other. Neither can produce a false green, because adding entries to `DERIVED` can only add failures. Case l pinning the exit-2 mode is sufficient; a second case would buy nothing that matters. **No finding.**

#### Finding 2's register — honest

`:30` reframes the block as *"Three limits … the first two are deliberate scoping decisions; the third is a precondition it cannot check"*, which is the distinction that was missing: a scoping decision and an unverifiable assumption are not the same kind of thing and should not sit in one list unlabelled. Limit 3 at `:47-57` then does the four things that make disclosure useful rather than decorative — names the precondition as a precondition, writes the defeating redirection out literally so a reader recognises the shape, says the failure is **silent**, and gives two remedies. `:115-117` no longer asserts the call sites *are* the declaration; it says *"So long as it holds, the call sites are the declaration."* The cross-reference is accurate too: the `--list-output-dirs` escape really is below, at `:59-65`.

That is as honest as prose can be about a hole prose is the only remedy for. **No finding.**

#### Finding 3 — closed

Case k pins it. Mutation (f) dropping the nested-child restriction at `:222` gives **23/24 on both shells, red on case k**, and I confirmed the real tree stays **green** under that mutation — so a fixture was indeed the only thing that could catch it. Adding the case despite my softening was the better call: the cost was one assertion, and the predicate had already consumed a review round.

---

#### Finding 1 (this round) — BLOCKING — the PR body still documents `8909b34`, and its mutation table now asserts an observation that is false of the shipped code

`class: tech`

The body's *How to read this PR?* prose was updated for the fix — my previous over-claim finding is corrected, and case j is named. The **evidence sections were not**. Against `1997b99`:

| Body location | Says | Actual at `1997b99` |
|---|---|---|
| *How to test this PR?*, lines 26-27 | `# 19/19` (twice) | **24/24** on both shells |
| line 46 | *"the **19/19** baseline re-confirmed between"* | baseline is **24/24** |
| **line 52, row (b)** | *"**15/19** both shells — **case i red**"* | **22/24** both shells, **case l red**; case i does **not** notice |
| line 57 | *"both are 19/19 on both shells"* | 24/24 |
| line 18, item 4 | *"cases e through i"* | cases e through **l** |
| mutation table | 6 rows | 10 mutations; **(e), (f), (g) absent** |
| anywhere | — | **case k and case l are never mentioned** |

Row (b) is the one that makes this blocking rather than cosmetic. It sits under the sentence *"**The mutation table is the evidence, and every row below is observed rather than predicted.**"* — so the body guarantees as observed a result that is not reproducible at the head it ships. Worse, (b) is the row the body leans on for R6: *"Row (b) is the discriminating one for R6."* At `1997b99` it no longer discriminates for R6 at all — widening made the matcher indentation-insensitive, which is the exact second-order consequence the fix commit's own test comments document. The body is the only artifact on this PR that contradicts that.

The `# 19/19` in *How to test this PR?* compounds it: a reader who follows the body's own instructions sees 24/24 and cannot tell whether the suite grew or something regressed.

This is a record describing a superseded state with nothing mechanical noticing — the failure class spec 0121 exists to end, in the PR that ends it. It is also the same class as the "67 vs 65" figure I raised last round, one scale up, so waving it through would be inconsistent.

**The fix is transcription, not work.** The correct 10-mutation `/24` table already exists — it was relayed to me in full and every number in it reproduces (below). Update lines 18, 26-27, 46, 52, 57 and the table; name cases j, k, l; and record the second-order consequence and case l's role, which is the most interesting thing in this commit and currently appears only in the source comments. One incidental consequence: **I could not verify mutation (g)**, because no artifact on the PR defines it — which is itself the finding.

---

### Observations — deliberately not findings

- **Widening enlarges the fail-closed surface, correctly.** A *non-comment* line naming a helper — say `[ -n "$source" ] || echo "check_or_write requires a source" >&2` — now trips **exit 2** where it was previously skipped. That is the design working (a line-based parser genuinely cannot tell that mention from a call, and `:161-164`'s rationale says refusing to run is the right answer), it names file, line and the offending text, and none exists today. Not a defect. If a clause ever lands in the `:141-151` comment saying so, a future maintainer saves a minute; I would not gate merge on it.
- **(d3) remains dead code and remains right to ship** — `24/24` on both shells, still behind `-gt 0`.
- The reverse-direction success line at `:292` is still asserted by no case (deleting it stays green). Unchanged from last round, still too minor to matter.

### The `--list-output-dirs` question — I agree it should not gate this PR

You asked for an `arch` position if I had one. I do not, and here is why, since a bare "agreed" is not worth much:

1. **R5 is discharged against the build as it exists.** I re-verified that every tree write funnels through the two helpers — `check_or_write` at `build-components.sh:288-289` and `propagate_skill_resources` at `321-322` are the only writes, and all 16 call sites are covered. There is no directory the build writes today that the guard does not check.
2. **The residual is a property of static parsing, not a defect in this change.** No call-site parser can see a write that is not a call. R5 cannot reasonably be read as requiring robustness against arbitrary future refactors of the build script, and spec 0121's *Out of scope* uses the same register when it defers classification decisions to separate tickets.
3. **The blast radius is disproportionate and lands in the wrong stage.** The refactor touches `scripts/build-components.sh`, which is named explicitly in the *CLI Matrix Maintenance* trigger list, so it drags `docs/cli-matrix.md` into a bug-fix PR. It also adds a query mode to the build's public surface — an `arch` decision that belongs in SPECS/PLAN, not appended to a REVIEW iteration.

One thing I would add: **no open issue tracks it.** I searched and found nothing (`#798` is the interpreter idiom, a different thing). Disclosure in a source comment is a good carrier for a maintainer already reading that file, and a poor one for planning. Filing the ticket would close the loop — a recommendation, not a finding.

---

### Convention compliance

Re-checked only what the new commit could disturb.

| Rule | Result |
|---|---|
| *Naming Convention* — Gitmoji on the new commit | ✅ `🐛 Extract a write-helper call that is not the first token of its line` |
| `8909b34` left unamended | ✅ both commits present, history append-only |
| **No closing keyword for #755** | ✅ `closingIssuesReferences` → `[]` |
| *Version Bump Convention* | ✅ N/A — no `artifacts/` source touched |
| *CLI Matrix Maintenance* | ✅ not triggered — only `scripts/check-*.sh` and `scripts/tests/test-*.sh`, on no trigger list |
| *Core-paths manifest co-maintenance* | ✅ N/A — neither the manifest nor `docs/layers.md` is in this commit |
| Executable bits | ✅ both `100755` |
| *Pull Request Format* | ❌ see Finding 1 — sections present, but *How to test this PR?* states the wrong expected result |

---

### Test coverage

**24 assertions across 12 cases** (a-d 2 each, e 2, f 1, g 2, h 2, i 4, j 2, k 1, l 2), **24/24 on `/bin/bash` 3.2.57 and 24/24 on `bash` 5.3.15**. Real tree: `OK: all 52 …` / `OK: all 9 …`, exit 0.

Every number relayed to me reproduces. Each mutation was applied to a pristine copy in an isolated clone, run on both shells, reverted and `diff -q`-confirmed, with the 24/24 baseline re-established between:

| Mutation | Relayed | Observed |
|---|---|---|
| (a) reverse block deleted | 10/24 | ✅ **10/24** both shells, red `e g h i j l` |
| **(b) trim removed (matcher on raw line)** | 22/24 | ✅ **22/24** both shells, **red `l` alone**; case l fails with `expected exit 1, got 2` |
| (c) 17th write site | 23/24 | ✅ **23/24** both shells |
| (d1) array guard off, governance loop | 22 on 3.2 / 24 on 5.x | ✅ **22/24** on 3.2 (red `c d`), **24/24** on 5.3 |
| (d2) array guard off, dedupe loop | 15 on 3.2 / 24 on 5.x | ✅ **15/24** on 3.2 (red `e f h i j k l`), **24/24** on 5.3 |
| (d3) array guard off, report loop | 24/24 both | ✅ **24/24** both — dead behind `-gt 0` |
| (e) matcher re-anchored to prefix | 22/24 | ✅ **22/24** both shells, **red `j` alone** |
| (f) nested-child restriction dropped | 23/24 | ✅ **23/24** both shells, **red `k` alone**; real tree green under it |
| (g) | 18/24 | ⚠️ **not verifiable** — no artifact on the PR defines this mutation (Finding 1) |

Cases j, k and l each die to exactly one mutation and nothing else, which is what a pin should look like: (e)→j, (f)→k, (b)→l. That the three new cases are mutually orthogonal, and orthogonal to the existing nine, is the strongest signal in this commit.

---

### Linter findings

`bash scripts/check-bash32-portability.sh`

```
OK: no forbidden Bash 4+ construct in scripts hooks (3 declared construct(s), 151 file(s) scanned).
```

`bash scripts/check-ci-parity.sh`

```
OK: reference, GitHub Actions, GitLab agree on the portable capability set; every pipeline job is traceable.
```

`bash scripts/check-test-wiring.sh`

```
OK: all test scripts are wired (67) or exempted with a reason (1); the exemption allowlist is honest.
```

`shellcheck scripts/check-core-paths.sh scripts/tests/test-check-core-paths.sh` → exit 1, **SC2016 (info) only**, now six hits (was four): `check-core-paths.sh:160,163` and `test-check-core-paths.sh:142,147,155,169`. The two new ones are the `BUILD_STUB_NESTED_CALL` and `BUILD_STUB_INDENTED_COMMENT` heredocs, which must keep `$out_root` unexpanded — deliberate, same as the existing four. No CI job gates on shellcheck. Nothing to change.

`markdownlint` → N/A, no Markdown in this commit.

---

### Summary

The engineering here is done. Finding 1 is closed with the minimal fix and pinned by a case no other mutation kills; Finding 3 is closed the same way; Finding 2's remedy is disclosure and the disclosure is honest, correctly relabelled as a precondition rather than a scoping choice. The second-order consequence — that widening the matcher silently retired the trim's only pin — was caught by the developer, diagnosed exactly right, and remedied at the one place the trim still decides an outcome. I attacked that specifically and it holds: the trim has one load-bearing role left, case l is on it, and the sub-mode case l does not cover fails loud in both directions.

What is left is the record. The body's evidence table still describes `8909b34` and now asserts, as observed, a row that no longer reproduces — in the PR whose entire purpose is to stop a record drifting without anything noticing. Transcribe the `/24` table, fix the four `19/19` figures and item 4, name cases k and l, and write down the second-order consequence where a reader will find it. Do that and this is an APPROVE with nothing outstanding.

hcross:
### Addendum to the verdict above — gate provenance, and a correction to my own evidence

Two notes prompted by the *Gate provenance* comment, which landed one minute before my verdict and which I had therefore not read when I wrote it.

**My verdict stands unchanged.** The PR body is byte-identical to the copy I read before writing (`gh pr view --json body` diffed against my captured copy — no difference; the `updatedAt` bump from `23:36:09` to `23:44:55` was that comment, not a body edit). So the blocking finding — the body still documenting `8909b34`, with row (b) asserting as observed a result that no longer reproduces — holds exactly as written.

**Correcting my own linter section.** I quoted `check-bash32-portability.sh` as green. That run was made in a clone at `1997b99`, which predates #799 (`9cba467`) — so it exercised the **superseded** reader, precisely the scoping error the provenance comment names. I verified the divergence rather than assuming it: `git diff 1997b99 51e2b94 -- scripts/check-bash32-portability.sh` is `1 file changed, 59 insertions(+), 1 deletion(-)`. My green was real but narrower than I presented it.

So I re-ran it against merged content myself rather than relying on the table above. `1997b99` merged with `51e2b94` in the sandbox — **clean, 0 conflicts** — and confirmed the merged tree carries main's version of the guard (`git diff --quiet HEAD 51e2b94 -- scripts/check-bash32-portability.sh` → identical). Exit codes captured plainly, one command per line, never through a pipe:

| Gate on merged content | Result |
|---|---|
| `check-bash32-portability.sh` — **main's new reader** | exit 0 — `no forbidden Bash 4+ construct in scripts hooks (3 declared construct(s), 151 file(s) scanned)` |
| `tests/test-check-bash32-portability.sh` — main's new suite | exit 0 — **31/31** |
| `check-core-paths.sh` | exit 0 — 52 forward entries, 9 built-output directories governed |
| `tests/test-check-core-paths.sh` on `bash` 5.3.15 | exit 0 — **24/24** |
| `tests/test-check-core-paths.sh` on `/bin/bash` 3.2.57 | exit 0 — **24/24** |
| `check-test-wiring.sh` | exit 0 |
| `check-ci-parity.sh` | exit 0 |

Independently reproduced, then `git reset --hard 1997b99`. The new reader judges these files the same way the old one did, so nothing in my correctness or test-coverage sections moves — but the corroboration for that row is this run, not the one I first quoted.

**The re-verification was the right call and the sequencing is right.** A green is scoped to the guard's revision as much as to the tree's, and my own quoted green is a live instance of getting that wrong. Order — re-review at `1997b99`, then `gh pr update-branch`, then merge — is correct, and not pushing the update mid-review is the courtesy that let me diff the body against what I had actually read.

hcross:
## Verdict: REQUEST CHANGES

Body-only re-review, head unchanged at `1997b99` (`updatedAt=2026-08-10T23:50:02Z`), CI **13/13 pass**, `closingIssuesReferences` → `[]`. No code re-verified this round beyond what the body's own claims required.

**Row (b)'s new text is true, not merely no longer false — I reproduced all three of its parts, including the counterfactual.** Every stale figure you fixed is gone. What remains is three **cross-references** rather than figures, one of which repeats last round's defect in a different paragraph: the requirement-coverage line still credits R6 to row (b).

---

### Row (b)'s positive claim — reproduced in full

The claim is a before-and-after, so I built both states rather than reasoning about them.

| Step | Constructed | Observed |
|---|---|---|
| (b) **before** the fix | `8909b34` guard + `8909b34` suite | **15/19** both shells — matches the body's historical figure exactly |
| (b) **after**, counterfactual | `1997b99` guard + **pre-fix suite** (baseline 19/19) | **19/19 GREEN** both shells — the mutation really is retired |
| (b) **after**, as shipped | `1997b99` guard + `1997b99` suite | **22/24**, case l red |
| is case l the **sole** re-pin? | case l excised (baseline 22/22 green), then (b) applied | **22/22 GREEN** both shells |

So each clause holds: widening made the pattern indentation-insensitive, (b) went from 15/19-red to fully green, the trim was left certified by nothing, and case l is what restores it — and nothing else in the suite does. The generalisation you added — *closing a check can retire a neighbouring one, and only re-running the whole table afterwards shows it* — is the right lesson and is now earned by evidence rather than asserted.

One methodological note on my own run: my first pass at the "excise case l" step reported 22/22 while the guard still carried mutation (b) from the previous step, so the number was right for the wrong reason. I re-ran it from a pristine copy, confirmed the 22/22 baseline **with the trim intact** first, and only then applied (b). The figure above is from the clean run.

### The other rows

Row **(g)** is now defined precisely enough to verify, and it does: removing the `#`-comment skip at `check-core-paths.sh:144` gives **18/24 on both shells**. With that, **all ten rows reproduce** — (a) 10/24, (b) 22/24, (c) 23/24, (d1) 24/5.x ÷ 22/3.2, (d2) 24/5.x ÷ 15/3.2, (d3) 24/24, (e) 22/24, (f) 23/24, (g) 18/24. Rows (b) and (g) each naming the exact line they remove is what made (g) checkable without asking you.

### Stale-figure sweep — clean

I swept the whole body for every `N/M` figure rather than grepping the two strings you named. Result: `24/24` ×6, `22/24` ×4, `23/24` ×2, `18/24`, `15/24`, `10/24` — all current — plus exactly two survivors from the `/19` family, both legitimate:

- **`15/19`** (line 64) is deliberately historical, scoped *"before the fix"*, and I confirmed it is the correct figure for `8909b34`.
- **`14/14`** (line 80) is scoped to the plan's original case set, a statement about plan history rather than about the shipped suite.

Neither is stale. `19/19`, `17/19`, `13/19`, `9/19`, `18/19` and `cases e through i`: zero occurrences.

### Testable claims in *How to test this PR?* — both reproduce

- `# 24/24` on both shells ✅ and the two real-tree `OK:` lines ✅.
- The revert demonstration ✅ — removing the two entries yields exactly `FAILED: 2 of 9 built-output director(y/ies) …` naming `- .agents/skills` and `- .agents/agents`.

### CI-wiring citations — correct

`build.yml:107,110`, `.gitlab-ci.yml:59-60` and `ci-capabilities.yml:68-69` all land on the right lines. My first check reported the third missing because I guessed `.github/workflows/`; it is `ci/ci-capabilities.yml`, where `:68-69` are indeed the guard and its suite. The body's bare-basename style is consistent across the list, so nothing to change.

---

#### Finding 1 — BLOCKING — the requirement-coverage line still credits R6 to row (b), contradicting the body's own lines 64 and 70

`class: tech`

Line 76: *"R6 case e with its discriminating twin case f, **plus row (b) above**."*

That clause is last round's defect in a different paragraph. R6 has two halves: a case the guard must reject, run on every CI run — cases e/f/h/j/l — and *"so that a guard which has stopped detecting the condition fails the build instead of passing it"*, the anti-blinding half. Row (b) used to carry the second half. At `1997b99` it does not, which line 64 explains at length and line 70 states outright: *"Case **i** remains the R6 discriminator for the derivation."*

So an auditor asking how R6 is discharged is sent to row (b), reads that row (b) went green after the fix and had to be re-pinned by case l, and comes away believing R6's anti-blinding guarantee rests on a mutation this PR says was retired. Requirement-coverage claims are the ones the lifecycle's termination test leans on, which is why I am treating a single clause as blocking rather than as a nit — and why waving it through after blocking the table for the same lean would be inconsistent.

**R6 is substantively discharged; only the citation is wrong.** I measured which cases each blinding mutation reddens so the replacement is grounded rather than plausible:

| Mutation | Cases reddened |
|---|---|
| (a) reverse block deleted | `e g h i j l` — **case i included** |
| (g) comment skip removed | `i l` — **case i included** |
| (e) matcher re-anchored | `j` alone |

Case i is still exercised as the anti-blinding discriminator, by (a) and (g). Something like *"R6 case e with its discriminating twin case f; the anti-blinding half by case i, which rows (a) and (g) redden, and by case j for call-site position via row (e)"* would be true. The exact wording is yours — the point is that the clause must stop pointing at row (b).

#### Finding 2 — non-blocking — the manifest line citations at line 82 are stale by exactly the ten lines this PR added

`class: tech`

Line 82: *"only `sync-from-upstream.sh` at 125 and `build-docs-index.sh` at 128 are"*.

Those were the line numbers **before** this PR. Measured:

```
base ea97bc4 :125  scripts/sync-from-upstream.sh   strict
base ea97bc4 :128  scripts/build-docs-index.sh     strict
head 1997b99 :135  scripts/sync-from-upstream.sh   strict
head 1997b99 :138  scripts/build-docs-index.sh     strict
```

A ten-line shift, which is exactly the size of this PR's own manifest addition — the eight-line comment block plus two entries. The body cites pre-diff coordinates for the file the diff modifies. The surrounding argument is unaffected and correct: I confirmed no `check-*.sh` is named individually in the manifest, so the guard does fall under the blanket `scripts` entry and the no-CLI-matrix-trigger conclusion stands. `102-103` and `docs/layers.md` `169-170`, cited elsewhere, are both still right.

#### Finding 3 — non-blocking — "0 of 68 propagate it" is falsified by this PR, and contradicts "the other 37" in the same sentence

`class: tech`

Line 84: *"The other 37 test suites that spawn their subject without propagating the interpreter — **0 of 68 propagate it**, 38 spawn a subprocess — are issue #798."*

Measured at `1997b99`: **68 suites under `scripts/tests/`, of which exactly 1 contains `${BASH:-bash}`** — `test-check-core-paths.sh`, at `:121`, this PR's own change. So it is 1 of 68, not 0. The sentence also disagrees with itself: *"the other 37"* presupposes one suite is fixed, while *"0 of 68"* presupposes none is. A survey figure taken before the fix and not re-derived after it.

I am not restating the "38 spawn" figure either way — my detection heuristic for that is not reliable enough to assert a count, and it is not load-bearing for the deferral to #798.

---

### Convention compliance

| Rule | Result |
|---|---|
| *Pull Request Format* — all sections present, *How to test* reproduces | ✅ |
| **No closing keyword for #755** | ✅ `closingIssuesReferences` → `[]` |
| Language — English only | ✅ |
| Head unchanged, no code motion | ✅ still `1997b99`, body-only edit |
| Naming, version bump, CLI matrix, core-paths co-maintenance | ✅ unchanged from last round |

### Summary

The body now tells the truth about the code, and row (b) — the passage I blocked on — is the strongest thing in it: a documented second-order consequence, with the retirement, the re-pin and the general lesson all reproducible. I verified the counterfactual specifically because a before-and-after claim is the kind that reads plausible and turns out unmeasured, and it held.

Three cross-references outlived the figures. Finding 1 is the one that matters: the R6 coverage clause is the last place in the document still describing the pre-fix arrangement, and it contradicts two other lines of the same body. Findings 2 and 3 are a shifted line pair and a survey figure the change itself invalidated — same family, lower stakes. Fix Finding 1 and I will approve on sight; take 2 and 3 in the same edit and there is nothing left outstanding.

I have no `arch` position on `--list-output-dirs`, unchanged from last round: defer it, and the one gap worth closing is that **no open issue tracks it yet**.

hcross:
## Verdict: APPROVE

Fourth-round, body-only review at `1997b99`, cold — no authoring context, fresh clone, mutations run in a scratchpad clone rather than the ticket worktree. Observed body: `updatedAt=2026-08-10T23:58:10Z`, `sha256(body)=e92c3f9b…50686`, re-read immediately before posting and unchanged since my first read.

**All three findings are closed, and the corrected text is true of the code rather than merely no longer false.** I reproduced the two mutation rows the corrected R6 credit now rests on, plus its negative half, and re-derived every figure in the body from the artifact. One non-blocking precision nit, stated below; it does not gate the merge.

---

### CI status

All 13 required checks **pass** at `1997b99`. Nothing failing, nothing pending: `build`, `check-agents-size`, `check-ci-parity`, `check-components`, `check-extension-version-bump`, `check-feedback-routing`, `check-skill-versions`, `check-spec-id-reserved`, `gitlab-ci-check`, `grep-anti-patterns`, `lint-markdown`, `lint-specs`, `test-harness-curate`. `mergeStateStatus: CLEAN`.

---

### Correctness

#### Finding 1 (was BLOCKING) — closed. Both red sets reproduced exactly, and the negative claim with them

The corrected clause credits R6's anti-blinding half to case **i**, exercised by rows **(a)** and **(g)**. I did not take the red sets on trust in either direction — I applied all three mutations to a clean `1997b99` clone and diffed the `FAIL` labels:

| Mutation applied | Observed | Red set observed | Body claims |
|---|---|---|---|
| (a) reverse block deleted (derivation 107-234 + verdict 267-292) | **10/24** both shells | `e g h i j l` | `e g h i j l` ✅ |
| (g) `\#*) continue ;;` at `check-core-paths.sh:144` deleted | **18/24** both shells | `i l` | `i l` ✅ |
| (b) `case "$trimmed"` → `case "$bs_line"` | **22/24** both shells | `l` **only — case i green** | "**Not** by row (b)" ✅ |

Both red sets land character-for-character, on both 5.3.15 and 3.2.57. The negative half holds too: under (b) case i stays green, so the body is right that (b) no longer carries the anti-blinding role.

The sets are also *internally* consistent with the table's pass counts, which is an independent check the body passes: with the live assertion census (a2 b2 c2 d2 e2 f1 g2 h2 **i4** j2 k1 l2 = 24), `e g h i j l` → 14 red → 10/24, and `i l` → 6 red → 18/24. Both match the figures printed in the same table.

**And case i is genuinely the discriminator, not merely a case that happens to redden.** Verified by construction at `scripts/tests/test-check-core-paths.sh:448-514`: it `cp`s the **real** `build-components.sh` into the fixture (`:469`), pairs it with a manifest naming no built output (`real.txt strict` alone), and asserts the **failure** path — exit 1 plus `- .agents/skills` and `- .agents/agents` by name, as literals never re-derived from the guard's own extraction. The success path prints only a count (`check-core-paths.sh:292`), so the body's reason for driving the failure path is exactly right. Both blinding mutations (a) and (g) make it red, which is precisely R6's clause: *a guard which has stopped detecting the condition fails the build instead of passing it*.

The other half of the clause checks out as stated: case e is a case the guard is expected to reject (ungoverned directory → exit 1) and case f its discriminating twin (same tree, manifest lists it → exit 0), per the assertion labels.

#### Finding 2 — closed. Both coordinates correct, and the shift is correctly attributed

```
head 1997b99  :135  scripts/sync-from-upstream.sh   strict
head 1997b99  :138  scripts/build-docs-index.sh     strict
main 51e2b94  :125  scripts/sync-from-upstream.sh   strict
main 51e2b94  :128  scripts/build-docs-index.sh     strict
```

Head and `main` coordinates both verified, each named with its tree. The attribution is verified rather than plausible: `git diff crewrig/main 1997b99 -- .crewrig/core-paths.txt` is exactly `+10 / -0`, in one hunk at `@@ -91,6 +91,16 @@` — above line 125 — so the ten-line shift *is* this PR's own manifest addition (8 comment lines + 2 entries). The surrounding conclusion also holds: no `check-*.sh` is named individually, so the guard falls under the blanket `scripts` entry.

#### Finding 3 — closed. Both figures re-derived with a parser, not a grep

Measured in Python over both trees (`git ls-tree` → `scripts/tests/test-*.sh`), counting non-comment interpreter propagation:

| Tree | Suites | Propagate the interpreter |
|---|---|---|
| `main` 51e2b94 | **68** | **0** |
| head `1997b99` | **68** | **1** — `test-check-core-paths.sh:121` |

Both figures exact, and `:121` is indeed the `( CREWRIG_REPO_DIR="$repo" "${BASH:-bash}" … )` line. Dropping the "38 spawn" figure was the right call rather than a concession: my own bare-form heuristic returns 38 on `main` / 37 at head, which is close enough to look authoritative and is exactly the kind of heuristic count that should not be asserted. The body is also consistent with the in-file comment at `:120` ("19 of 68 carry the bare form") — different measures, compatible, no contradiction.

#### Sweep for the three failure families named in the brief

**Figures whose provenance is unreachable from the artifact — none found.** Every mutation row is now reproducible from the body alone; I reproduced (a), (g) and (b) working only from the row text, and (a) and (g) name the block and the exact line they remove. The remaining rows name their target unambiguously against labelled structure in the guard: (c) a 17th write site, (d1/d2/d3) which of the three loops loses its array guard, (e) the matcher's anchoring, (f) the coverage predicate `dir_is_governed()`.

**Line numbers or counts without their tree — none found.** `:135`/`:138` vs `:125`/`:128` are explicitly attributed; `test-check-core-paths.sh:121` is scoped *"at this PR's head"* in its own sentence; manifest `102-103` and `layers.md` `169-170` are lines this PR creates, so they exist only at head; `check-core-paths.sh:144` sits inside the table whose preamble scopes every row to `1997b99`. The citations the PR does **not** touch are tree-invariant and verified at head anyway — `.gitignore:47`, `build.yml:107,110`, `.gitlab-ci.yml:59-60`, `ci/ci-capabilities.yml:68-69`, and all 16 `build-components.sh` call sites.

**"Observed rather than predicted" — the guarantee holds where I tested it.** I reproduced 3 of the 10 rows exactly, plus both real-tree `OK:` lines verbatim (`all 52 …` / `all 9 …`) and the revert demonstration, which yields precisely the quoted block:

```
FAILED: 2 of 9 built-output director(y/ies) derived from
scripts/build-components.sh carry no upstream-sync guarantee:
  - .agents/skills
  - .agents/agents
```

The rows I did not run are arithmetically consistent with the assertion census above ((b)→l→22, (e)→j→22, (f)→k→23, (c)→case i's count assertion→23), and row (c)'s mechanism is corroborated by the suite's own header note that a tenth output directory reddens case i on its count.

#### Other body figures, all re-derived

- **65 tracked files (43 + 22)** — `git ls-tree -r` gives 43 under `.agents/skills`, 22 under `.agents/agents`. The `.agents/` root is **67**, and the two extras are exactly `.agents/ANTIGRAVITY.md` and `.agents/settings.local.json.example` — so the parenthetical explaining the retired "67" is precisely correct.
- **16 call sites / 9 directories** — my own parser over `build-components.sh` returns the claimed list line-for-line (447, 448, 517, 518, 549, 550, 582, 583, 620, 630, 660, 679, 737, 752, 770, 785), all 16 targets literally beginning `"$out_root/`, and depth-2 extraction yielding exactly 9 directories.
- **"two characters"** — the widening is `"check_or_write "*` → `*"check_or_write "*` (and its twin): two `*`. Exact.
- **"15/19"** — the denominator is reachable: `8909b34` baselines at **19/19**, and j/k/l add exactly 5 assertions (2+1+2) to reach 24. The figure is correctly scoped *"before the fix"*.
- **"cases e through l, and one line in `run_check`"** — `main` carries a–d only; head adds e–l; `run_check` has exactly one functional line changed.
- **R1–R7** — checked against the seven requirements of `specs/0121-antigravity-outputs-in-core-paths.md` on `main`; each clause maps correctly. R4 verified literally: those three names appear only on comment lines (`:96`, `:100`), never as entries, and `.gitignore:47` keeps the last untracked.
- **"No policy column, matching every sibling"** — true of all 9 built-output entries, the two new ones included; none carries a second field.
- **Blast radius** — the PR touches exactly 4 files (`core-paths.txt`, `layers.md`, the guard, its suite). No `artifacts/`, no built outputs, no provenance carrier, so "no version bump" is correct and `check-skill-versions` agrees.

#### Non-blocking nit — the ⚠️ warning's scope is broader than the behaviour by two files

`class: tech`

Line 5: *"A fork that has hand-edited **anything under `.agents/`** will now have its next `scripts/sync-from-upstream.sh` run halt."*

The halt covers `.agents/skills` and `.agents/agents` — there is no bare `.agents` entry in the manifest, which I confirmed, and which is the deliberate design the same body defends at R4 and in the manifest's own comment block. So an edit confined to `.agents/ANTIGRAVITY.md` or `.agents/settings.local.json.example` does **not** halt a sync, before or after this PR.

I am flagging it as a nit and not a finding because it errs in the safe direction — it over-warns rather than under-warns — and because it is true of 65 of the 67 tracked files in that subtree. "anything under `.agents/skills` or `.agents/agents`" would make it exact. Fix it or don't; it does not gate the merge, and I would not spend a fifth round on it.

---

### Convention compliance

| Rule | Result |
|---|---|
| *Pull Request Format* — required sections present; *How to test* reproduces | ✅ |
| **No closing keyword for #755** | ✅ `closingIssuesReferences` → `[]`, and no `close/fix/resolve #N` in the body |
| *Language* — English only | ✅ |
| *Naming Convention* — Gitmoji title, no conventional-commit prefix | ✅ `🐛 Guarantee upstream sync for Antigravity built outputs` |
| *Version Bump* — no skill/agent/extension source touched | ✅ not applicable |
| *CLI Matrix Maintenance* — `check-*.sh` / `tests/test-*.sh` on neither trigger list | ✅ verified: the trigger list at `docs/cli-matrix-maintenance.md:13` covers `scripts/{build,install,setup,import,manage}-*.sh`, and the symmetric-script list at `:75-80` names setup/import/manage/build only — `check-core-paths.sh` matches neither brace set, and its suite sits under `scripts/tests/` |
| *Core-paths manifest co-maintenance* — manifest + `docs/layers.md` in the same diff | ✅ `102-103` ↔ rows `169-170`, R7 satisfied |
| *Logbook* — #755 is its own logbook per Rule A | ✅ |

---

### Test coverage

Not re-reviewed as code per the round's scoping, but the body's coverage claims were exercised as claims. Baseline `24/24` reproduces on both `/bin/bash` (3.2.57) and `bash` (5.3.15). The suite is wired on all three CI surfaces at the cited lines, so R6's "every CI run" holds by inheritance. Cases j, k, l are new at `1997b99` as stated, and case h does use the nested carve-out form (`.newcli strict` + `.newcli/skills excluded`) while case g asserts non-zero **and** `build-components.sh:2` — both exactly as the body describes.

Every mutation I applied was reverted and the tree re-verified clean with `git status --porcelain` between runs; the ticket worktree was never touched.

---

### Linter findings

`lint-markdown.sh` on `docs/layers.md` — exit 0, no findings.

`lint-shell.sh` on both changed scripts — exit 1, six findings, all `SC2016 (info)`:

```
In scripts/check-core-paths.sh line 160:
      *'$out_root/'*) ;;
       ^----------^ SC2016 (info): Expressions don't expand in single quotes, use double quotes for that.

In scripts/check-core-paths.sh line 163:
    rest="${rest#*'$out_root/'}"
                  ^----------^ SC2016 (info): Expressions don't expand in single quotes, use double quotes for that.

In scripts/tests/test-check-core-paths.sh lines 142, 147, 155, 169:
BUILD_STUB_NEWCLI='#!/bin/bash
                  ^-- SC2016 (info): Expressions don't expand in single quotes, use double quotes for that.
```

All six are correct-by-design and not findings against this PR: the guard must match the **literal** text `$out_root/` in the build script's source, and the fixtures are single-quoted so `$out_root` stays literal — the test file says so at its stub definitions. Double-quoting any of them would break the mechanism. No finding raised. `lint-skill.sh`, `lint-python.sh` and `lint-json.sh` do not apply — no `SKILL.md`, `.py` or `.json` in the diff.

---

### Summary

The body is now true of `1997b99`. Finding 1 was the one that mattered and it is closed with the strongest available evidence: the replacement credit names measurable red sets, and both of them reproduce exactly, as does the negative claim about row (b) that makes the correction necessary. Findings 2 and 3 are closed with their trees named and their shift attributed. Nothing in the body is now in the register of an observation without being one, no figure is unreachable from the artifact, and no citation is ambiguous about which revision it describes.

The carry-forward from the last round is also resolved: **#804** now tracks the not-a-call write path, so the deferred `--list-output-dirs` gap is no longer untracked.

One non-blocking nit on the warning's scope, which I would not hold a merge for. **Approving — the code was cleared in earlier rounds and the description now matches it.**

PR #773, opened 2026-08-09
✨ Add an exclusive, attributable claim on a shared ticket worktree

hcross:
A ticket's worktree is shared by every agent staffed on it, so one agent's
routine `git reset --hard` discards a sibling's uncommitted work outright and
announces nothing — the suite still runs green at its older case count, and the
traces left behind establish that a whole-tree operation happened but not which
agent ran it. This makes that operation refuse instead of proceed: an exclusive
claim held over a clean tree, an append-only ledger that outlives both the claim
and the worktree, and a documented verification path that touches no shared tree
at all.

Implements `specs/0114-shared-worktree-agent-isolation.md` (R2–R12). Closes #736.

## How to read this PR?

Six files, **2 441 insertions, 0 deletions** — no existing behaviour changes.
`git rev-list --count 3a0cd8f..216d1eb` = **24** commits, each self-contained.

| Read | File | Lines | What it is |
|---|---|---|---|
| 1 | `docs/agent-team-protocol.md` | +47 | The contract. 5 130 bytes inserted into *Worktree Isolation*. |
| 2 | `scripts/worktree-claim.sh` | +857 | The mechanism. |
| 3 | `scripts/tests/test-worktree-claim.sh` | +1 532 | 22 cases, each falsified before being trusted. |
| 4 | `.github/workflows/build.yml` · `ci/ci-capabilities.yml` · `.gitlab-ci.yml` | +3 · +1 · +1 | The suite wired into both pipelines. |

Every count above is measured at **`216d1eb`**, against merge-base `3a0cd8f`.
That pin is deliberate: a magnitude stated without its reference frame is a claim
that silently becomes false, and this body has now been through six REVIEW
iterations, of which iterations 3, 4 and 5 each found a magnitude that was true
where it was measured and false where it was written. Anything below that names a
magnitude names the commit it was taken at.

**No `file:line` citation appears in this body — by decision, not by omission.**
Every pointer is an anchor that survives an edit: an identifier, a subcommand, a
`--help` entry, a quoted heading, or a commit SHA. Rationale is at the end of
*What the REVIEW iterations changed*; the short version is that a line number in
a squash-merged body describes a branch that keeps moving, and this body carried
eleven wrong ones before they were removed. Please do not reintroduce them. The
standing exhibit: `run_release_on_exit` is defined on line 760 at `f8bd54b` and
on line 780 at `216d1eb`, because `44d0c9b` is a net twenty lines of insertion
above it.

**Start with the document, not the script.** It is what agents actually read, and
[...]

PR #803, opened 2026-08-10
🔐 Sign the graft commit wherever the fork signs its own commits

hcross:
Makes the one commit `scripts/sync-from-upstream.sh` authors under `--preserve-history` carry the same signature status as any other commit that repository would create — and forbids it from silently degrading to an unsigned one. Implements spec 0122 for issue #756.

## ⚠️ One adopter-visible break

`scripts/sync-from-upstream.sh` is a `strict` core path, so adopters receive this without opting in. **A fork whose signing configuration cannot be satisfied or parsed now exits non-zero where it previously produced an unsigned commit and exited 0.** That is spec 0122 R5 working as specified — degrading to an unsigned commit is not an outcome the script accepts, warning or no warning — but it is a behaviour change, and a fork with a broken signing setup meets it on its next sync rather than in a release note.

A fork that does not sign gets the same commit it got before, with no new interaction and no new failure mode (R3). One exit code moves on a path that already failed: a `commit-tree` failure unrelated to signing now surfaces as 1 rather than 128, because the call is wrapped. Nothing that succeeded before fails now.

## How to read this PR?

`git commit-tree` is plumbing. It **neither reads `commit.gpgsign` nor is overridden by it** — both measured — so neither direction is automatic and the script must resolve the setting itself and pass `-S` conditionally. R3 is therefore a real branch in the code, not a property that falls out for free.

Read the guard block at `scripts/sync-from-upstream.sh:637-707` and pay attention to three things. (A narrower `:645-675` in an earlier draft excluded the post-condition at `:679-700` and the `update-ref` at `:702` — the subject of point 2 below.)

**1. The presence probe is an `if` condition; the value read is not.** That asymmetry is the fix for a defect a plan review caught before it shipped. A non-zero status from the *presence* probe legitimately means "unset". But git exits **128** on a malformed boolean, and a command substitution in the word of a `[ … ]` test — or anywhere inside an `if` condition, where `errexit` is suspended — discards that status **twice over**, yielding an unsigned commit and exit 0. An assignment propagates it. Reproduced end to end: with `commit.gpgsign = yesplease`, the v1 shape prints `History-preserving commit created` and `Sync complete.`, exits 0, unsigned.

**2. The refusal is checked *before* `update-ref`.** Refusing after the ref has moved would satisfy R5 by violating R4's "leave the branch tip where it stood".

**3. The script emits its own English cause, and that is not decoration.** Git's signing diagnostic is **localized**. The same failing run:

```
erreur : Couldn't load public key …/absent-key.pub: No such file or directory?   # default locale
error: Couldn't load public key …/absent-key.pub: No such file or directory?     # LC_ALL=C
```

R4 requires a message naming the unproducible signature as the cause; git's cannot be that message, and no test may assert on it. **No assertion depends on git's wording for a positive match.** (`dd`'s grep does span git's fetch stderr; measured across C, fr_FR and de_DE it is two git lines, none matching, so the conclusion holds — but the stronger phrasing "nothing in the suite reads git's line" would be wrong.) That single artifact — one run, two locales, our line unmoved — is the R4 evidence.

## How to test this PR?

```sh
/bin/bash scripts/tests/test-sync-from-upstream.sh                                 # 65/65 on 3.2.57
bash      scripts/tests/test-sync-from-upstream.sh                                 # 65/65 on 5.3.15
GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null bash scripts/tests/test-sync-from-upstream.sh   # 65/65 hermetic
```

Eight new cases, `cc` through `jj`. The refusal on a configured-but-unsignable repository, verbatim:

```
Error: --preserve-history refuses to commit — this repository is configured to sign
commits (commit.gpgsign) but the signature could not be produced; see the git error above.
The branch tip is unchanged and the restored files remain in the working tree.
[exit=1]
HEAD before: 7b41c38f…
HEAD after : 7b41c38f…   (unmoved: R4)
core-file.txt in the working tree: 'v2'   (restored content kept: R4)
```

**The mutation table is the evidence.** Thirteen mutations, every result observed rather than predicted, and reported as **red-case counts** rather than pass totals — the suite grew from 64 to 65, so `62/64 → 62/65` would read as "no change" where one row genuinely changed:

| # | Mutation | Red cases | Which |
|---|---|---|---|
| M1 | drop the flag | **3** | `cc`, `ee`, **`jj`** — was 2 before `jj` existed |
| M2 | unconditional `-S` | 2 | `dd` signed; `ii` exit 1 |
| M3a | `\|\| true` | 1 | `ee`, stderr only |
| M3b | degrade via retry | 1 | `ee`, stderr only — FAIL line byte-identical to M3a |
| M4 | probe too early | 3 | `ff`, `ee`, `ii` — `jj` stays green, correctly |
| M5 | header-slice growth | 2 | `dd`, `ii` — `jj` stays green, correctly |
| M6 | relocate the post-condition after `update-ref` | 1 | `hh`, **HEAD assertion only** |
| M7 | delete the post-condition | 1 | `hh`, all three assertions |
| M8 | revert the config read to the swallowing shape | 1 | `gg`, all three |
| M9 | make the presence probe fatal | 1 | **`ii` alone** |
| M10script | restore the trailing space in `sync-from-upstream.sh` only | 1 | `jj`, **4 assertions** |
| M10helper | restore it in `commit_is_signed` only | 1 | `jj`, **1 assertion** — the sync itself stays green |
| M10both | the full reversion of the SHA-256 fix | 1 | `jj`, 4 assertions |

**M1 is the row that genuinely changed**: dropping `-S` breaks SHA-256 signing exactly as it breaks SHA-1, so `jj` now reddens beside `cc` and `ee`. M4 and M5 leave `jj` green and that is correct — `jj` has a usable key, so an early probe succeeds, and its commit is genuinely signed, so a full-output match still finds the header.

The three **M10** rows are what close the review finding. They matter separately: `M10script` and `M10helper` show each of the two readers is covered **independently**, which single-file mutations prove and a combined one only asserts.

**M6 versus M7 is the pair worth re-running** — one assertion red versus three. `hh` discriminates a *misplaced* post-condition from an *absent* one, not merely "something is wrong".

Each mutation was applied through a harness that **aborts when its anchor is not found exactly once**, so "the mutation did not take" cannot masquerade as "the mutation had no effect", and reverted from a byte-exact snapshot verified two ways (`diff -q` plus the suite back to 65) before the next.

Gates, with `$?` captured plainly rather than read off an OK line: `check-bash32-portability.sh` 0 · `check-core-paths.sh` 0 · `check-test-wiring.sh` 0 · `check-ci-parity.sh` 0 · `task lint-markdown` 0 (it prints no verdict — the rc is the whole signal) · `task spec:lint` 0 · the Rule 1 and Rule 3 convention greps, zero hits each · `shellcheck -S warning` clean.

**Requirement coverage.** R1/R2 the graft commit inherits the repository's own setting, signed under its designated identity. R3 a non-signing fork is unchanged — new case `dd`. R4 refusal leaves no commit, the tip unmoved, the restored files in the tree, and names the cause in a locale-stable line. R5 no unsigned-when-configured outcome is accepted — new cases `gg` (malformed value), `hh` (empty-signature degrade) and the post-condition before `update-ref`. R6/R7 both parents survive and imported SHAs are untouched — asserted with two **genuinely distinct** parents, since git deduplicates a repeated parent with a warning and yields a one-parent commit, which would assert the invariant against a commit that never had two parents to lose. R8 the no-op path exits zero even where signing is impossible — new case `ff`. R9 the three mandated configurations plus `ff` and `ii`.

**The signature-header predicate is `^gpgsig`, not `^gpgsig `.** Git's signature header name is hash-algorithm dependent: a repository created `--object-format=sha256` writes `gpgsig-sha256`, which a trailing space cannot match. Before this fix, such a repository had its **correctly signed** graft commit refused as unsigned — `commit-tree -S` exited 0 and produced a commit carrying `gpgsig-sha256` and `BEGIN SSH SIGNATURE`, and the script answered *"built an unsigned commit"* with exit 1. That breaches R1 and R2 rather than serving R5: for that fork an ordinary `git commit` yields a signed commit, while `--preserve-history` yielded **no commit at all** where it previously yielded an unsigned one.

Found by cold review, not by the suite — and the second half of the finding is the instructive one. **Applying the one-character fix as a mutation left the suite at 64/64**: the assertions could not tell the corrected predicate from the broken one, so the edit alone would have closed nothing. New case `jj` is what closes it, and the three `M10*` rows above prove it does.

Reach, stated rather than inflated: a SHA-256 fork of a SHA-1 upstream cannot reach the site at all — `git fetch` refuses first with an algorithm mismatch — so this needs a fully SHA-256 pair, which `canonical_repo` permits but no adopter has today. The direction was fail-safe throughout: no unsigned commit was ever produced, and the tip never moved. Narrow, not absent.

Dropping the space stays safe because the `sed -n '/^$/q;p'` slice exposes only header lines, continuation lines begin with a space, and `gpgsig-sha256` is the only other header that can match.

**One assertion inside `jj` was moved rather than written once.** A first draft checked the `gpgsig-sha256` header name *after* the sync, on the graft commit. Under the script-only reversion that assertion fired with *"this case no longer exercises the hash-dependent header name"* — which is false: the fixture produces the header fine, the script misreads it, and since HEAD never moves the check was reading the old tip and naming a cause that had not occurred. That is Finding 1's own defect class — a message naming a cause that did not happen — so the premise is now checked **before** the sync against a throwaway `commit-tree -S` that is referenced by nothing and moves no ref. It stays correctly green under the reversion.

**What that premise check does and does not promise.** It has two disjuncts, and the first fires whenever the probe cannot produce a signature **at all** — so it covers a superset of "git's header naming changed", and on the part outside that set its message names a cause that did not occur. Measured: with the signing program pointed at a nonexistent path and header naming untouched, `jj` reddens on five lines, the premise first. An earlier draft of this paragraph claimed it fires *only* on a naming change; that is false, and it is the same defect class the paragraph documents fixing, one level down. Splitting the disjuncts into two messages would let the stronger claim stand and is the natural follow-up; it is not done here.

The premise check earns its place regardless, and one mutation shows why: with the fixture silently **not** SHA-256, all four of `jj`'s behavioural assertions stay green — a SHA-1 fixture syncs fine — so without this check plus the format sanity loop, `jj` would report PASS while exercising the SHA-1 header name. Those two guards are the only thing that catches it.

**Signature presence is asserted without any verification infrastructure**, which matters because spec 0122 excludes verification from scope: the commit object's header block is sliced with `sed -n '/^$/q;p'` before matching `^gpgsig`. An unanchored `grep '^gpgsig'` over the whole object **false-positives** on an unsigned commit whose message body contains a line starting `gpgsig ` — measured, and the reason M5 exists.

**Four things DEV found that no review had named**, all measured:

- **The `gpg.ssh.program` stub must take the *last* argument, not `$3`.** Git invokes `<prog> -Y sign -n git -f <pubkey> <buffer>` — captured argv: `ARGC=7`, buffer last — so a `$3`-based stub would target `-n`, sign the wrong file, and make `commit-tree -S` fail. The stub loops to the last argument, staying independent of git's argument order. **Correction:** an earlier draft of this bullet added that a `$3` stub would therefore make `hh` *pass* for the wrong reason, refusing at step 2 instead of step 3. Cold review measured both stubs against `hh`'s own fixture and that is **false** — the `$3` stub makes `hh` **fail**, because `hh`'s third assertion requires the script's own `built an unsigned commit`, which is precisely what separates step 2's refusal from step 3's. The bullet understated the test it described: `hh` already discriminates the case it warned about. The mechanism was right, the conclusion drawn from it was not, and it appeared under a heading asserting the list was measured.
- **`ff` needs no signing capability**, contrary to the plan's Risks note: the R11 no-op short-circuit returns before `commit-tree` is reached, so it uses an unreachable key path and calls `ssh-keygen` not at all. Only `cc` and `hh` generate a key, and both carry an explicit precondition that **FAILs naming the capability** rather than skipping.
- **M4's count matched but its composition is environment-dependent.** Here the reds were `ff`, `ee`, `ii` — not `dd`, because the probe signs successfully with the operator's global key. On CI, or with no usable key, `dd` reddens too and the count drops below 61. `ff` is the row that pins the hazard; the rest is collateral from `set -e` aborting before the real path.
- **M3a and M3b are indistinguishable by design.** Both `rc=1`, HEAD unmoved, both landing on the post-condition's message. The only differing byte is `fatal : Not a valid object name` — localized, therefore unassertable on principle. `ee`'s stderr assertion separates the *config* refusal from the *post-condition* refusal; it does not separate M3a from M3b, and nothing in the suite does or should.

**One deliberate deviation from the plan**, flagged rather than hoped to read as in-scope: step 5 said "add one line to the file's front comment". DEV also added a third failure-mode bullet and changed *"Two failure modes apply only in this mode"* to *"Three"* — adding a refusal path while leaving the count at two would have shipped a false statement in the file's own front comment.

**Blast radius.** No build outputs, no `artifacts/` source, no version bump. **CLI matrix not triggered** — `sync-from-upstream.sh` matches none of `scripts/{build,install,setup,import,manage}-*.sh` and `docs/cli-matrix.md` has zero occurrences of it. Both files are `strict` core paths (`.crewrig/core-paths.txt:125` and the blanket `scripts` entry at `:51`), which is why the adopter-visible break is stated at the top of this body. `docs/adoption-guide.md`'s `--preserve-history` subsection deserves a paragraph on the new refusal; deliberately deferred to a follow-up and recorded here as a conscious omission rather than left silent.

**Known and accepted.** The refusal path leaves the index staged, because `git add -A` runs before the new code. R4's clause about the working tree holds; unwinding the index would discard whatever the operator had staged.

**Plan history.** Four revisions, two REQUEST CHANGES rounds, five findings — every one produced by prototyping and mutating rather than reading. The sharpest is the one in *How to read this PR?* point 1: the plan rejected `|| true` for swallowing a status, then shipped a substitution inside an `if` condition that swallows it twice over. The plan's own `### Risks` section ended up carrying an inventory of claims made without measuring; its count was wrong twice before it was right.

Issue #756 is the logbook per `AGENTS.md` → *Logbook Issues → Rule A*, and is closed manually after merge per Rule C — this body carries no closing keyword. Spec 0122's `status` moves `approved` → `implemented` in its own metadata-only pull request afterwards.

### DIFF
--- a/scripts/sync-from-upstream.sh
+++ b/scripts/sync-from-upstream.sh
@@ -682,9 +682,19 @@ if [ "$PRESERVE_HISTORY" = true ]; then
   # reachable: a signing program that exits 0 without writing a signature
   # makes `commit-tree -S` exit 0 and produce a commit with no gpgsig header.
   # The `sed -n '/^$/q;p'` slice stops at the blank line that ends the header
-  # block, so a message body beginning with `gpgsig ` cannot pass for one.
+  # block, so a message body beginning with `gpgsig` cannot pass for one.
+  #
+  # The pattern carries NO trailing space, deliberately: git's signature
+  # header name is hash-algorithm dependent — `gpgsig` in a SHA-1 repository,
+  # `gpgsig-sha256` in one created with `--object-format=sha256`. A trailing
+  # space matches the first and misses the second, which would make this
+  # post-condition refuse a signature it had just successfully produced. Not
+  # anchoring on the right stays safe inside the slice: only header lines are
+  # visible there, the signature's own continuation lines begin with a space,
+  # and `gpgsig-sha256` is the only other header name that can match. Case jj
+  # fails if the space comes back.
   if [ "$GRAFT_SIGN" = true ] && \
-     ! git -C "$REPO_DIR" cat-file commit "$GRAFT_COMMIT" | sed -n '/^$/q;p' | grep -q '^gpgsig '; then
+     ! git -C "$REPO_DIR" cat-file commit "$GRAFT_COMMIT" | sed -n '/^$/q;p' | grep -q '^gpgsig'; then
     echo "Error: --preserve-history built an unsigned commit though this repository is configured to sign; refusing to move the branch tip." >&2
     exit 1
   fi
--- a/scripts/sync-from-upstream.sh
+++ b/scripts/sync-from-upstream.sh
@@ -40,8 +40,10 @@
 # branch: it creates a single commit whose tree is byte-identical to what the
 # restore alone produced and whose second parent is FETCH_HEAD, so later
 # history-inspection commands (`git log`, `git merge-base`, `git bisect`)
-# surface the upstream lineage directly. Two failure modes apply only in this
-# mode:
+# surface the upstream lineage directly. That one commit inherits the
+# repository's own commit.gpgsign setting (spec 0122): it is signed wherever
+# an ordinary `git commit` in that repository would be. Three failure modes
+# apply only in this mode:
 #
 #   - Shallow-clone refusal: exits non-zero before any fetch, restore, or
 #     commit when the local repository is a shallow clone. This is a
@@ -54,6 +56,11 @@
 #     — every `strict`/`adopt-on-edit` manifest entry, minus any nested
 #     `excluded` child — plus the .crewrig/.synced-markers/ bookkeeping
 #     directory.
+#   - Signing refusal: aborts the history-preserving step (same posture as
+#     the anti-pollution guard — no commit, branch tip unmoved, restored
+#     files left in the working tree) when the repository is configured to
+#     sign commits but the signature cannot be determined or produced.
+#     Degrading to an unsigned commit is never an accepted outcome.
 #
 # When FETCH_HEAD is already an ancestor of the current branch tip, the mode
 # is a no-op: no commit is created and the script exits zero.
@@ -575,7 +582,8 @@ for i in "${!PATHS[@]}"; do
 done
 
 # ---------------------------------------------------------------------------
-# --preserve-history: history-preserving graft commit (spec 0086 R5-R9, R11).
+# --preserve-history: history-preserving graft commit (spec 0086 R5-R9, R11;
+# spec 0122 R1-R8 for that commit's signature status).
 # Reached only when the policy-aware restore above completed without
 # aborting (R4) — the strict dirty-core guard and any per-path failure exit
 # earlier, above this point, so no ancestry is ever recorded on that path.
@@ -625,9 +633,68 @@ if [ "$PRESERVE_HISTORY" = true ]; then
   GRAFT_TREE="$(git -C "$REPO_DIR" write-tree)"
   UPSTREAM_SHA="$(git -C "$REPO_DIR" rev-parse FETCH_HEAD)"
   COMMIT_MSG="🔀 Graft upstream history via --preserve-history ($(git -C "$REPO_DIR" rev-parse --short FETCH_HEAD))"
-  GRAFT_COMMIT="$(git -C "$REPO_DIR" commit-tree "$GRAFT_TREE" -p "$BRANCH_TIP" -p "$UPSTREAM_SHA" -m "$COMMIT_MSG")"
+
+  # spec 0122 R1-R2 — commit-tree neither reads commit.gpgsign nor is
+  # overridden by it, so neither direction is automatic: resolve the
+  # repository's own setting and pass -S ourselves.
+  #
+  # The presence probe stays an `if` condition because there a non-zero status
+  # legitimately means "unset". The VALUE read must not be one: git exits 128
+  # on a malformed boolean, and a command substitution in the word of a
+  # `[ … ]` test — or anywhere inside an `if` condition, where errexit is
+  # suspended — discards that status twice over, silently yielding an
+  # unsigned commit and exit 0. An assignment propagates it. The explicit
+  # branch below names the cause in English because git's own diagnostic here
+  # is localized, so it is not something an operator or a test can rely on.
+  GRAFT_SIGN=false
+  GRAFT_SIGN_ARGS=()
+  if git -C "$REPO_DIR" config --get commit.gpgsign >/dev/null 2>&1; then
+    if ! GRAFT_SIGN_RAW="$(git -C "$REPO_DIR" config --bool --get commit.gpgsign)"; then
+      echo "Error: --preserve-history cannot determine whether to sign this commit — commit.gpgsign holds a value git cannot read as a boolean; see the git error above." >&2
+      echo "The branch tip is unchanged and the restored files remain in the working tree. Correct commit.gpgsign and re-run." >&2
+      exit 1
+    fi
+    if [ "$GRAFT_SIGN_RAW" = true ]; then
+      GRAFT_SIGN=true
+      GRAFT_SIGN_ARGS=(-S)
+    fi
+  fi
+
+  # The array expansion is guarded per docs/scripting-conventions.md Rule 5's
+  # fourth, deliberately non-grep-detectable trap: under `set -u` Bash 3.2
+  # treats an empty array as unset, and empty is exactly what
+  # GRAFT_SIGN_ARGS holds on the non-signing path — every fork spec 0122 R3
+  # exists to leave alone.
+  if ! GRAFT_COMMIT="$(git -C "$REPO_DIR" commit-tree "$GRAFT_TREE" \
+    -p "$BRANCH_TIP" -p "$UPSTREAM_SHA" ${GRAFT_SIGN_ARGS[@]+"${GRAFT_SIGN_ARGS[@]}"} -m "$COMMIT_MSG")"; then
+    if [ "$GRAFT_SIGN" = true ]; then
+      echo "Error: --preserve-history refuses to commit — this repository is configured to sign commits (commit.gpgsign) but the signature could not be produced; see the git error above." >&2
+      echo "The branch tip is unchanged and the restored files remain in the working tree. Fix the signing setup (key, agent, gpg.format) and re-run, or clear commit.gpgsign for this repository." >&2
+    else
+      echo "Error: --preserve-history could not create the graft commit; see the git error above." >&2
+    fi
+    exit 1
+  fi
+
+  # spec 0122 R5 — refuse a degraded (unsigned) commit BEFORE the ref moves.
+  # The identical check placed after update-ref would leave the branch tip
+  # sitting on the unsigned commit, satisfying R5 by violating R4. The path is
+  # reachable: a signing program that exits 0 without writing a signature
+  # makes `commit-tree -S` exit 0 and produce a commit with no gpgsig header.
+  # The `sed -n '/^$/q;p'` slice stops at the blank line that ends the header
+  # block, so a message body beginning with `gpgsig ` cannot pass for one.
+  if [ "$GRAFT_SIGN" = true ] && \
+     ! git -C "$REPO_DIR" cat-file commit "$GRAFT_COMMIT" | sed -n '/^$/q;p' | grep -q '^gpgsig '; then
+    echo "Error: --preserve-history built an unsigned commit though this repository is configured to sign; refusing to move the branch tip." >&2
+    exit 1
+  fi
+
   git -C "$REPO_DIR" update-ref -m "sync-from-upstream --preserve-history" HEAD "$GRAFT_COMMIT"
-  echo "History-preserving commit created: $GRAFT_COMMIT (parents: $BRANCH_TIP, $UPSTREAM_SHA)"
+  # The suffix reports the signature the check above observed on the object,
+  # not the fact that -S was passed (docs/scripting-conventions.md Rule 2).
+  GRAFT_SIGN_NOTE=""
+  if [ "$GRAFT_SIGN" = true ]; then GRAFT_SIGN_NOTE=" (signed)"; fi
+  echo "History-preserving commit created: $GRAFT_COMMIT$GRAFT_SIGN_NOTE (parents: $BRANCH_TIP, $UPSTREAM_SHA)"
 fi
 
 echo "Sync complete. Review the changes with 'git diff' before committing."
--- a/scripts/tests/test-sync-from-upstream.sh
+++ b/scripts/tests/test-sync-from-upstream.sh
@@ -101,6 +101,39 @@
 #      produced by `backup_file()`, and does NOT match the sibling
 #      `extension.json` (pattern stayed narrow).
 #
+# Spec-0122 signed-graft-commit cases (issue #756):
+#  cc. R1, R2, R6, R7 — a repository configured to sign (gpg.format ssh, an
+#      ephemeral ed25519 key, commit.gpgsign true) gets a SIGNED graft commit
+#      that still carries both parents, whose second parent is the fetched
+#      upstream commit byte-for-byte, and whose creation is reported with a
+#      `(signed)` suffix on stdout.
+#  dd. R3 — a repository that does not sign (commit.gpgsign false) gets an
+#      unsigned graft commit, exit 0, no `(signed)` on stdout, and no
+#      signing-related line on stderr: the non-signing path gains nothing.
+#  ee. R4, R5 — commit.gpgsign true with a user.signingkey that does not
+#      exist: the sync refuses, creates no commit, leaves the branch tip
+#      where it stood, leaves the restored content in the working tree, and
+#      names the unproducible signature in its OWN message. git's signing
+#      diagnostic is localized, so no assertion here may read it.
+#  ff. R8 — the no-op path (FETCH_HEAD already an ancestor) still exits 0 in
+#      a repository configured to sign but unable to: no commit means no
+#      signature to produce, so signing capability is never probed.
+#  gg. R1, R5 — commit.gpgsign holds a value git cannot read as a boolean.
+#      `git commit` itself exits 128 in such a repository, so reporting a
+#      successful sync would breach R1 on its face. Guards the shape of the
+#      predicate: a config read placed in the word of a `[ … ]` test inside
+#      an `if` condition discards git's 128 (errexit is suspended there) and
+#      silently produces an unsigned commit with exit 0.
+#  hh. R5 ordering — a gpg.ssh.program that exits 0 without writing a
+#      signature makes `commit-tree -S` exit 0 and produce a commit with no
+#      gpgsig header. The sync must refuse with the branch tip UNMOVED; the
+#      same check placed after update-ref would refuse with the tip already
+#      on the unsigned commit, satisfying R5 by violating R4.
+#  ii. R3 by absence — commit.gpgsign unset, which is the modal non-signing
+#      shape and the one no other case reaches (init_git_repo pins it to
+#      `false`). Invoked with the operator's own configuration neutralised,
+#      because the predicate resolves through global config too.
+#
 # Usage:
 #   bash scripts/tests/test-sync-from-upstream.sh
 
@@ -134,6 +167,15 @@ init_git_repo() {
   git -C "$dir" config user.email "test@example.com"
   git -C "$dir" config user.name "Test"
   git -C "$dir" config commit.gpgsign false
+}
+
+# commit_is_signed <repo> <rev>
+# True when <rev> carries a gpgsig header. The sed slice stops at the blank
+# line that ends the header block: `git cat-file commit` also prints the
+# message, so a naive full-output match reports SIGNED for an unsigned commit
+# whose message body happens to begin with `gpgsig `.
+commit_is_signed() {
+  git -C "$1" cat-file commit "$2" | sed -n '/^$/q;p' | grep -q '^gpgsig '
 }
 
 # make_initial_commit <repo> [<file> <content>]...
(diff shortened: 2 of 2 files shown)


hcross:
## Verdict: REQUEST CHANGES

Cold review, no authoring context. Method: fresh clone checked out at `54cf59b`, both changed blobs hash-verified against the head ref (`99c3b79a1a5a…` for the script, `bc19f8180faf…` for the suite), every mutation applied through an anchor-exact harness that aborts unless its anchor matches exactly once, and reverted byte-exactly (`cmp -s`) between runs. Exit statuses captured plainly, never read off a pipeline's last stage. `grep -F` used for every asserted figure, because this environment's `grep` is `ugrep` and returns 0 silently on `${…}` patterns.

Two findings. The change is otherwise correct, and the mutation table reproduces exactly — all ten rows, including the discriminating pair.

---

### CI status

All 13 required checks **pass** on `54cf59b` (`gh pr checks 803`): `build`, `check-agents-size`, `check-ci-parity`, `check-components`, `check-extension-version-bump`, `check-feedback-routing`, `check-skill-versions`, `check-spec-id-reserved`, `gitlab-ci-check`, `grep-anti-patterns`, `lint-markdown`, `lint-specs`, `test-harness-curate`. None failing, none pending.

The new cases do **execute** in CI, not merely sit wired: `.github/workflows/build.yml:104` runs the suite inside `check-components`, and that job's log reports `Results: 64/64 passed` with `PASS  case-cc …` present. Preflight clear.

---

### Findings

#### Finding 1 — `^gpgsig ` misses a SHA-256 repository's signature header, so a correctly signed graft commit is refused as unsigned

`class: tech`

`scripts/sync-from-upstream.sh:687` and `scripts/tests/test-sync-from-upstream.sh:178` both match `'^gpgsig '`. Git's signature header name is **hash-algorithm dependent**: in a repository created with `--object-format=sha256`, git writes the signature under `gpgsig-sha256`, which `^gpgsig ` (trailing space) cannot match.

Measured, end to end, against the script at this head — a SHA-256 canonical repo and a SHA-256 adopter fork, `gpg.format ssh`, an ephemeral ed25519 key, `commit.gpgsign true`:

```text
object-format: adopter=sha256 upstream=sha256
exit_rc=1
Error: --preserve-history built an unsigned commit though this repository is configured
to sign; refusing to move the branch tip.
tip_moved: no
```

The commit `commit-tree -S` produced in that same fixture is **signed**: `commit_tree_S_rc=0`, header block carries `gpgsig-sha256`, and the value contains `BEGIN SSH SIGNATURE`. So the run refuses a signature it successfully produced, and the message misdiagnoses the cause.

Against the spec's plain terms this breaches R1 and R2 rather than serving R5: for that repository an ordinary `git commit` yields a signed commit, while `--preserve-history` now yields **no commit at all** and a non-zero exit — where before this change it completed with an unsigned one. Spec 0122 draws no object-format carve-out; R2 is written for any "repository whose configuration causes the commits it creates to be signed".

Reach, stated honestly rather than assumed: a SHA-256 fork of a **SHA-1** upstream cannot reach the site at all — `git fetch` refuses first, measured `fetch_rc=128`, `fatal: … client sha256 ; serveur sha1` — so this needs a fully SHA-256 pair, which `canonical_repo` permits but no crewrig adopter has today. The direction is fail-safe (no unsigned commit, tip unmoved). That makes it narrow, not absent.

The fix is one character in each file — `'^gpgsig'`, no trailing space — and it stays safe: inside the `sed -n '/^$/q;p'` slice only header lines are visible, continuation lines begin with a space, and `gpgsig-sha256` is the only other header that can match. I applied exactly that fix as a mutation (M10 below): the suite stays **64/64**, which is the second half of this finding — the suite cannot currently tell the fixed predicate from the broken one, so closing this needs a case, not only the edit.

#### Finding 2 — the `$3`-stub claim in the PR body is false as stated; measured, such a stub turns `hh` red, not green

`class: tech`

*Detailed description (for agents)* → first bullet of "Four things DEV found that no review had named, **all measured**" states that a `$3`-based stub "silently signs the wrong file, `commit-tree -S` then fails, and `hh` would pass **for the wrong reason** — refusing at step 2 instead of step 3."

The first half is confirmed. Captured argv: `ARGC=7`, `ARG1=-Y ARG2=sign ARG3=-n ARG4=git ARG5=-f ARG6=<pubkey> ARG7=<buffer>` — the buffer is last, so the last-argument stub the PR ships is required and `$3` would target `-n`.

The conclusion drawn from it is wrong. Both stubs run against `hh`'s own fixture and the script at this head:

```text
=== stub=dollar3  exit_rc=1  tip_moved=no
  stderr carries STEP-2 message (config/commit-tree refusal)
  stderr: no step-3 message   <- case hh's assertion would FAIL
=== stub=lastarg  exit_rc=1  tip_moved=no
  stderr: no step-2 message
  stderr carries STEP-3 message (post-condition refusal)  <- what case hh asserts
```

`hh`'s third assertion (`scripts/tests/test-sync-from-upstream.sh:2108`) requires the script's own `built an unsigned commit`. That string is precisely what separates step 2's refusal from step 3's, so a `$3`-based stub makes `hh` **fail**, not pass. The body understates its own test: `hh` already discriminates the wrong-reason pass it warns about.

The code and the test are right; the durable artifact records a false measurement inside a list asserted as measured, and a later reader would conclude `hh` is blind to a distinction it in fact catches. Correct the bullet.

---

### Test coverage — the ten-mutation table, re-derived here

Baseline before any mutation: **64/64, rc 0** on `/bin/bash` 3.2.57 and on `bash` 5.3.15, and **64/64, rc 0** under the body's third command (`GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null`). All three documented commands hold.

| # | Mutation | Body claims | Observed here | |
|---|---|---|---|---|
| M1 | drop the flag | 62 · `cc` on exit 1, `ee` on stderr | 62/64 · `cc` (8 lines, first is `expected exit 0, got 1`), `ee` (stderr) | ✅ |
| M2 | unconditional `-S` | 62 · `dd` signed, `ii` exit 1 | 62/64 · `dd` `carries a gpgsig header`, `ii` exit 1 + HEAD unmoved | ✅ |
| M3a | `\|\| true` | 63 · `ee`, stderr only | 63/64 · `ee`, 1 assertion | ✅ |
| M3b | degrade via retry | 63 · `ee` stderr only, FAIL line byte-identical to M3a | 63/64 · `ee`, 1 assertion, line **byte-identical** to M3a | ✅ |
| M4 | probe too early | 61 · `ff`, `ee`, `ii` | 61/64 · `ee`, `ff`, `ii` — `dd` green here, exactly as disclosed | ✅ |
| M5 | header-slice growth | 62 · `dd`, `ii` report SIGNED | 62/64 · `dd`, `ii` report SIGNED | ✅ |
| M6 | relocate the post-condition after `update-ref` | 63 · `hh`, **HEAD assertion only** | 63/64 · `hh`, **exactly 1** line: `the branch tip moved onto the unsigned commit` | ✅ |
| M7 | delete the post-condition | 63 · `hh`, all three | 63/64 · `hh`, **3** lines: exit 0 · tip moved · no message | ✅ |
| M8 | revert the config read to the swallowing shape | 63 · `gg`, all three | 63/64 · `gg`, 3 lines: exit 0 · tip moved · no message | ✅ |
| M9 | make the presence probe fatal | 63 · **`ii` alone** | 63/64 · `ii` alone, 2 lines | ✅ |

**M6 versus M7 — the pair asked about — discriminates, and is not a finding.** M6 leaves exit non-zero and the message intact and reddens only the tip assertion; M7 reddens all three. `hh` therefore distinguishes a *misplaced* post-condition from an *absent* one, not merely "something is wrong". The single M6 line is the placement line, verbatim.

Two rows need their mutation stated exactly as PLAN v4 writes it, or they read as no-ops — worth recording so the next re-run does not conclude the mutation failed to take:

- **M3b** must replace the *whole* failure branch with the retry. Swapping only `exit 1` for a retry, keeping the two `echo`s, leaves the suite at **64/64** — and correctly so: the post-condition still refuses, the tip still holds, and step 2's message is still printed, so that variant is warn-then-refuse, which is the specified behaviour, not the degradation R5 forbids.
- **M4** must be a throwaway `commit-tree -S` before the no-op short-circuit. Relocating the *config read* early leaves the suite at 64/64 — which is what PLAN v4 step 15 already says in writing ("`ff` does **not** detect an early config read, which cannot fail").
- **M5** is compound: the helper's slice removed **and** a `gpgsig `-prefixed commit message. Removing the helper's slice alone stays 64/64 (M12 below).

**Four mutations of my own, beyond the table:**

| # | Mutation | Result | Reading |
|---|---|---|---|
| M10 | `'^gpgsig '` → `'^gpgsig'` (the Finding 1 fix) | **64/64** | the sha256 predicate is uncovered in both directions |
| M11 | drop the empty-array guard, keep `"${GRAFT_SIGN_ARGS[@]}"` | **64/64** | inert here — see observations |
| M12 | remove the slice from `commit_is_signed` alone | **64/64** | the slice is exercised only by M5's compound fixture |
| M14 | make the ` (signed)` suffix unconditional | 62/64 · `dd`, `ii` | the suffix **is** pinned to the observed signature, per Rule 2 |

**Requirement coverage is real, not asserted.** Spec 0122 R9's three mandated configurations are present and each is reached by a distinct fixture: signing (`cc`), non-signing explicit (`dd`) and by absence (`ii`), refusal (`ee`, plus `gg` for the unparseable value and `hh` for the degrade). `ii`'s isolation is genuinely necessary and genuinely works — measured on this machine, a repository with `commit.gpgsign` unset resolves to `true` from `~/.gitconfig` (`unisolated_rc=0`, value `true`), and under `GIT_CONFIG_NOSYSTEM=1 HOME=<empty> XDG_CONFIG_HOME=<empty>` the presence probe returns 1, genuinely unset. Without that isolation `ii` would be a *signing* fixture on a maintainer's laptop.

---

### Correctness — claims checked rather than taken

Each of these was measured here against git 2.55.0:

- **`commit-tree` neither reads `commit.gpgsign` nor is overridden by it**, both directions. Without `-S` under `commit.gpgsign=true`: 0 `gpgsig` headers. With `-S` under `commit.gpgsign=false`: 1. The conditional `-S` is load-bearing, and R3 is a real branch.
- **The presence-probe / value-read asymmetry at `:651-656` matches the observed exit codes.** On a malformed value the presence probe exits **0** and `config --bool --get` exits **128** with a localized diagnostic (`fatal : valeur booléenne de configuration invalide 'yesplease' pour 'commit.gpgsign'`). The comment's reasoning is exact.
- **The `sed -n '/^$/q;p'` boundary is exact for a multi-line signature.** A synthetic PGP-armour `gpgsig` header whose continuation carries the armour's blank line — which git writes as a single space, not an empty line — is still detected as signed. The slice does not truncate on it.
- **The unanchored match genuinely false-positives.** An unsigned commit whose message body begins `gpgsig ` reports SIGNED without the slice and unsigned with it. M5's premise is load-bearing and true.
- **Two genuinely distinct parents.** Passing the same commit twice makes git warn (`le parent dupliqué … est ignoré`), exit **0**, and yield a **one**-parent commit. `cc`'s `parent_count -ne 2` assertion at `:1741` is what makes the R6/R7 claim self-proving rather than vacuous.
- **A duplicated `commit.gpgsign` key does not trip the refusal**: `--get` returns the last value with exit 0, and `git commit` also succeeds. No false refusal on that shape.
- **Locale.** No assertion depends on git's wording for a positive match. `ee`, `gg` and `hh` all assert the script's own English strings.
- **Bash 3.2.** `GRAFT_SIGN_ARGS` is expanded exactly **once** (`:669`), and that expansion is guarded. Baseline is 64/64 under `/bin/bash` 3.2.57, and the subject script itself parses and runs under it.

---

### Convention compliance

- **Title** — `🔐 Sign the graft commit wherever the fork signs its own commits`: Gitmoji, no conventional-commit prefix. ✅
- **No closing keyword for #756** — `gh pr view 803 --json closingIssuesReferences` returns `[]`. ✅ The body states the manual Rule C close explicitly.
- **Logbook** — Rule A satisfied: #756 is the logbook, 10 incremental comments spanning the open-question closure, two measurement corrections, four plan revisions and three plan reviews. ✅
- **Branch** — `fix/0756-signed-graft-commit` matches the `<prefix>/<NNNN>-<slug>` shape in use (`fix/0789-…`, `feat/0736-…`). ✅
- **Version bump** — not triggered: `docs/version-bump-convention.md` scopes the rule to `artifacts/**` and `extensions/**` skill/agent carriers; neither changed file is one. `check-skill-versions` passes. ✅
- **CLI matrix** — not triggered, verified rather than accepted: `scripts/sync-from-upstream.sh` matches none of `scripts/{build,install,setup,import,manage}-*.sh`, and `grep -cF 'sync-from-upstream' docs/cli-matrix.md` returns **0**. ✅
- **Core paths** — the body's citations are exact: `.crewrig/core-paths.txt:125` is `scripts/sync-from-upstream.sh	strict`, `:51` is the bare `scripts` entry, and lines 6-7 do define an absent policy column as `strict`. The adopter-visible break is disclosed at the top of the body, which is the right place for it. ✅
- **Built components** — `artifacts/` untouched, so no rebuild obligation. ✅
- **Front comment** — `Three failure modes` is now accurate: the block carries exactly three `#   - ` bullets (shallow-clone, anti-pollution, signing). ✅
- **PR body shape** — all three mandated headings present. It adds one extra top-level section (`## ⚠️ One adopter-visible break`) ahead of them, which `docs/pr-format.md` does not show; #773 already takes comparable preamble latitude, so I record it and do not raise it.

---

### Linter findings

Run against the pristine head checkout, in a directory no mutation ever touched:

| Gate | Exit | Output |
|---|---|---|
| `.claude/skills/pr-reviewer/scripts/lint-shell.sh` (both files) | 0 | no output, no degradation note |
| `shellcheck -S warning` (both files) | 0 | clean |
| `shellcheck` default severity, incl. style (script) | 0 | clean |
| `scripts/check-bash32-portability.sh` | 0 | `OK: no forbidden Bash 4+ construct in scripts hooks (3 declared construct(s), 151 file(s) scanned).` |
| `scripts/check-core-paths.sh` | 0 | `OK: all 50 strict/adopt-on-edit core-paths entries resolve at HEAD.` |
| `scripts/check-test-wiring.sh` | 0 | `OK: all test scripts are wired (67) or exempted with a reason (1); the exemption allowlist is honest.` |
| `scripts/check-ci-parity.sh` | 0 | `OK: reference, GitHub Actions, GitLab agree on the portable capability set; every pipeline job is traceable.` |
| Rule 3 grep (as `scripting-conventions.yml` runs it) | — | zero hits |
| Rule 1 grep | — | zero hits |

`shellcheck` 0.11.0 was present, so none of these is a degraded pass.

---

### The two disclosed items, judged

**The front-comment deviation is in scope — keep it.** PLAN v4 step 5 asked for one line; DEV also added the third failure-mode bullet and changed *Two* to *Three*. That is not scope creep, it is the same edit finished: the block enumerates the failure modes of this mode, the change adds one, and leaving the count at two would ship a false statement in the file's own front comment — a defect a reviewer would then have to raise. Verified: three bullets, count correct.

**M4's environment-dependent composition is acceptable in a mutation table, and the disclosure is accurate.** Reproduced exactly as described: 61/64 with `ee`, `ff`, `ii` red and `dd` green, because the unconditional probe signs with the operator's global ssh key (`commit.gpgsign=true`, `gpg.format=ssh` in `~/.gitconfig` here). The row's claim rests on `ff`, which reddens in either environment; the varying reds are collateral from `set -e` aborting early. What matters is that the *suite* is not environment-dependent, and it is not: 64/64 on two shells, 64/64 hermetically with global and system config nulled, and 64/64 in CI.

---

### Non-blocking observations

No `class:` tag — these are notes, not findings to route.

1. **The empty-array guard is inert in this file, and its comment implies otherwise.** `scripts/sync-from-upstream.sh:70` is the only `set` in the script and it is `set -e` — no `-u`, no `pipefail`. Rule 5's fourth trap is conditioned on `set -u`, so the guard at `:669` cannot fire here; M11 (guard removed, array kept) stays 64/64. Keeping the guarded form is right — it is the prescribed standing form and it costs nothing — but the comment at `:663-667` reads as though `set -u` were active in this script. A half-sentence would fix it.
2. **`dd`'s stderr assertion does span git's output.** `:1827` greps `[Ss]ign` over a file that also holds git's fetch lines, so the body's "Nothing in the suite reads git's line" is true of positive assertions and loose about this negative one. Measured across three locales the stderr is two git lines and none matches — `C`: `From …`; `fr_FR.UTF-8`: `Depuis …`; `de_DE.UTF-8`: `Von …` — so the R4 conclusion stands and this is wording, not a defect.
3. **"A fork that does not sign is bit-for-bit unaffected" is very slightly over-stated.** On the pre-existing path where `commit-tree` fails in a non-signing repository, the exit code moves from git's 128 (old `set -e` abort) to 1, plus one script-authored stderr line. R3 still holds — same failure mode, no new one — but "bit-for-bit" is not what changed hands.
4. **The `docs/adoption-guide.md` paragraph deferred in the body is the right call to have flagged**, and worth a follow-up issue rather than a memory: the refusal is reachable by an adopter with a broken signing setup on their next sync, and the guide is where they will look.

---

### What would clear this

Finding 1 needs the one-character predicate fix in both files plus a case that reddens when it regresses (M10 shows the suite is currently blind to it). Finding 2 needs the bullet corrected in the body. Nothing else in the change is blocked: the ten-row mutation table reproduces exactly, M6/M7 discriminate as claimed, the suite is hermetic in three configurations and green in CI, and every convention gate passes.

hcross:
## Verdict: REQUEST CHANGES

Re-review of head `735bd76`, targeted at the revision. Both blobs hash-verified against the head ref (`087a32ac5a21…`, `cb4d8ba9b60f…`). Scope as agreed: the fix, case `jj`, the relocated premise assertion, the changed counts, and the two standing caveats. Everything I verified on `54cf59b` and did not re-derive is noted as such.

**Finding 1 is closed** — the predicate fix, the new case, and the three `M10` rows all reproduce exactly as reported. **Finding 2 is closed** — the correction is in the body and states the measured half only.

One new finding, and it is one sentence in the body with no code change behind it. Everything else clears, including both standing caveats, which CI has now settled.

---

### CI status

All 13 required checks **pass** on `735bd76`. `check-components` was *pending* when I picked this up and has since completed (6m19s, pass); I held the verdict until it resolved rather than reviewing against a partial signal.

**The two standing caveats are now measured, not assumed.** From that job's log:

```text
check-components  Test sync-from-upstream script  PASS  case-jj: SHA-256 repository gets a signed graft commit (gpgsig-sha256 recognised)
check-components  Test sync-from-upstream script  Results: 65/65 passed
```

`case-jj` **passing** on `ubuntu-latest` establishes both capabilities at once — `ssh-keygen` signing support and `git init --object-format=sha256` — because the case FAILs naming each rather than skipping. `cc` and `hh` pass in the same run. Nothing about those preconditions rests on assumption any more.

---

### Finding — the sentence justifying the relocation claims an exclusivity the code does not have

`class: tech` · **body only, one sentence, no code change required**

*Detailed description (for agents)*, the paragraph on the moved assertion, closes: *"It fires **only if** git's header naming actually changes, and stays correctly green under the reversion."*

The second clause is confirmed. The first is false, and I measured it two ways. The check has two disjuncts (`test-sync-from-upstream.sh:2293`):

```sh
if [ -z "$probe_commit" ] || ! git -C "$adopter" cat-file commit "$probe_commit" 2>/dev/null | sed -n '/^$/q;p' | grep -q '^gpgsig-sha256 '; then
```

The first disjunct fires whenever the probe cannot produce a signature **at all** — nothing to do with header naming. Mutating the fixture's signing program to a nonexistent path (`JJ_NOSIGN`), leaving the header naming untouched:

```text
MUTATION JJ_NOSIGN | 64/65 | per_case: 5 case-jj
FAIL  case-jj: this fixture does not produce a 'gpgsig-sha256' header — the case no longer
               exercises the hash-dependent header name it exists to cover
FAIL  case-jj: expected exit 0, got 1 — a signature that WAS produced was read as absent
   … (3 more)
```

So it fires on a superset, and on the part of that superset outside the claim it **names a cause that did not occur** — which is the exact defect class this paragraph exists to describe fixing. The first clause of the message is literally true (no header, because no commit); the gloss that follows — *"the case no longer exercises the hash-dependent header name"* — is not, and neither is the neighbouring *"a signature that WAS produced was read as absent"* when none was produced.

This is reachable rather than theoretical, and by the suite's own reckoning: `jj`'s keygen precondition tests key **generation** while its message names signing support at `OpenSSH >= 8.2`. An environment that can generate an ed25519 key but not sign with it therefore clears the precondition and lands here. I could not measure that OpenSSH boundary directly — this machine is OpenSSH 10.2p1 — so I am citing the suite's own stated floor plus the measured `JJ_NOSIGN` behaviour, not a version claim of my own.

The narrow fix is the sentence: say it fires when the fixture stops producing the `gpgsig-sha256` header, whatever the reason. If you would rather the code matched the stronger claim, splitting the disjuncts into two messages does it — but that is optional and I am not asking for it here.

I am raising this rather than filing it as a note for consistency: last round's Finding 2 was also a body-only over-claim, of this same class, and was accepted and fixed. Waving through the same shape one iteration later would make the standard depend on who wrote the sentence.

---

### Finding 1 — closed, and the three `M10` rows reproduce exactly

Baseline at `735bd76`: **65/65, rc 0** on `/bin/bash` 3.2.57, on `bash` 5.3.15, and hermetically under `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null`. All three documented commands hold at their stated counts.

| Row | Reported | Observed here |
|---|---|---|
| **M10script** — space restored in `sync-from-upstream.sh` only | `jj`, 4 assertions | 64/65 · `jj`, **4** |
| **M10helper** — space restored in `commit_is_signed` only | `jj`, 1 assertion, sync green | 64/65 · `jj`, **1** |
| **M10both** — full reversion | `jj`, 4 assertions | 64/65 · `jj`, **4** |

`M10script`'s four, verbatim — note the first is a purpose-built diagnostic, not a generic mismatch:

```text
FAIL  case-jj: expected exit 0, got 1 — a signature that WAS produced was read as absent
FAIL  case-jj: HEAD did not move — the graft commit was refused though the repository can sign
FAIL  case-jj: commit_is_signed reports the graft commit unsigned in a SHA-256 repository
FAIL  case-jj: stdout does not report the commit as signed
```

**`M10helper`'s single red is genuine coverage, not a fixture accident** — which is what you asked me to establish, so here is the reasoning made observable rather than argued. The mutated helper has four callers. Under the mutation:

- `cc` (positive, SHA-1) — **PASS**. `gpgsig ` still matches with the space restored.
- `dd` (negative, unsigned) — **PASS**. No match either way.
- `ii` (negative, unsigned) — **PASS**. No match either way.
- `jj` (positive, SHA-256) — **FAIL**, on exactly the one assertion that consults the helper.

All three PASS lines are in the log. The mutation is discriminated by object format, which is the property under test — so the single red is the *only* correct outcome, and the other three callers prove the mutation was live rather than inert. The two readers are separable in the output as well as in the count: `M10script` reddens four lines including the exit-code diagnostic, `M10helper` reddens one and leaves the sync green.

---

### The relocated premise assertion — attacked as asked

**The throwaway is genuinely inert.** I snapshotted the fixture before and after running the exact probe from the case — all refs, `HEAD`, `symbolic-ref`, a digest of `.git/index`, a digest of `ls-files -s`, `status --porcelain`, a digest of every worktree file, and the reflog. Diff of the two snapshots:

```text
 --- reflog count:
       1
 --- loose objects:
-      7
+      8
```

That is the whole delta: **one loose object**, the probe commit itself. No ref moved, `HEAD` unchanged, index byte-identical, `status --porcelain` empty before and after (so the R8 anti-pollution guard sees nothing new), worktree digest unchanged, reflog unchanged. Reachability, three ways: not pointed at by any ref, absent from `log --all`, and `git fsck` reports it as `objet commit inatteignable`. Signing writes its buffer and `.sig` under `TMPDIR`, not into the repository. "Referenced by nothing and moves no ref, index, or file" is accurate; the one thing it does leave is an unreachable object in a `mktemp` fixture the suite's `EXIT` trap removes.

**It still detects what it was written for, and `JJ_SHA1` is the row that proves the relocation earned its place.** Making the fixture silently non-SHA-256 — git's header naming for that fixture genuinely changing to `gpgsig`, which is the closest reachable proxy for a rename:

```text
MUTATION JJ_SHA1 | 64/65 | per_case: 3 case-jj
FAIL  case-jj: fixture repo …/upstream-sha256.HrerdB reports object format 'sha1', expected 'sha256' …
FAIL  case-jj: fixture repo …/adopter-sha256.yPcElq reports object format 'sha1', expected 'sha256' …
FAIL  case-jj: this fixture does not produce a 'gpgsig-sha256' header — the case no longer exercises …
```

The premise check fires. The decisive detail is what *stayed green*: **all four behavioural assertions passed**, because a SHA-1 fixture syncs perfectly well. Without the premise check and the format sanity loop, `jj` would have reported PASS while exercising the SHA-1 header name — precisely the failure the case comment names. The two guards are load-bearing, and only they catch it.

**The relocation is also correct on its own terms**, verified rather than accepted: a plain `git init` re-init inside `init_git_repo` preserves `--object-format=sha256` (measured, `sha256` before and after), and the format sanity loop would catch it if a future git changed that. The premise check uses its own inline pattern rather than `commit_is_signed`, which is why it stays green under both single-file reversions — confirmed in the `M10*` logs.

One thing I checked and will *not* claim: I wondered whether `HEAD^{tree}` avoids an index write that `write-tree` would cause. Measured, `write-tree` left the index file byte-identical on a fresh commit, so that contrast does not hold and I am not crediting it.

---

### The changed counts — all thirteen rows re-derived at this head

The switch from pass totals to red-case counts is the right call: `62/64 → 62/65` would indeed have read as "no change" where M1 changed. Every row below was run at `735bd76`, not carried over.

| # | Body claims | Observed here | |
|---|---|---|---|
| M1 | **3** · `cc`, `ee`, **`jj`** | 3 · `cc`(8 lines), `ee`(1), `jj`(4) | ✅ |
| M2 | 2 · `dd`, `ii` | 2 · `dd`(1), `ii`(2) | ✅ |
| M3a | 1 · `ee`, stderr only | 1 · `ee`(1) | ✅ |
| M3b | 1 · `ee`, byte-identical to M3a | 1 · `ee`(1), line byte-identical | ✅ |
| M4 | 3 · `ff`, `ee`, `ii` — `jj` green | 3 · `ee`(1), `ff`(2), `ii`(2) — `jj` **green** | ✅ |
| M5 | 2 · `dd`, `ii` — `jj` green | 2 · `dd`(1), `ii`(1) — `jj` **green** | ✅ |
| M6 | 1 · `hh`, HEAD assertion only | 1 · `hh`(**1**), the tip line | ✅ |
| M7 | 1 · `hh`, all three | 1 · `hh`(**3**) | ✅ |
| M8 | 1 · `gg`, all three | 1 · `gg`(3) | ✅ |
| M9 | 1 · **`ii` alone** | 1 · `ii`(2 lines) | ✅ |
| M10script | 1 · `jj`, 4 assertions | 1 · `jj`(4) | ✅ |
| M10helper | 1 · `jj`, 1 assertion | 1 · `jj`(1) | ✅ |
| M10both | 1 · `jj`, 4 assertions | 1 · `jj`(4) | ✅ |

**M1 genuinely changed, as stated** — `jj` now reddens beside `cc` and `ee`. **M6 versus M7 still discriminates** at this head: one line versus three, M6's being the placement line. The `jj`-green claims for M4 and M5 hold, for the stated reasons.

I re-ran the seven rows I had only measured at `54cf59b` rather than reasoning that the predicate change could not touch them. It could not — but the body now asserts these counts at *this* head, and reasoning is not what this ticket has been settling arguments with.

---

### Carried forward from `54cf59b`, not re-derived

Per scope, unchanged and still standing: the sed-slice boundary (exact for a multi-line PGP-armour header), the unanchored-match false positive, two genuinely distinct parents (git dedups a repeated parent with a warning into one), `ii`'s isolation (measured mandatory — unisolated resolves `true` from `~/.gitconfig`), the config exit codes (presence 0, `--bool` 128), `commit-tree` neither reading nor being overridden by `commit.gpgsign`, the single guarded Bash 3.2 array expansion, the locale analysis, and both disclosed items I cleared (the front-comment deviation, M4's environment-dependent composition).

---

### Convention compliance and linters at this head

Re-run against a pristine checkout of `735bd76` in a directory no mutation touched:

| Gate | Exit | Output |
|---|---|---|
| `lint-shell.sh` (both files) | 0 | no output, no degradation note |
| `shellcheck -S warning` / default severity | 0 / 0 | clean (shellcheck 0.11.0 present — not a degraded pass) |
| `check-bash32-portability.sh` | 0 | `OK: no forbidden Bash 4+ construct … 151 file(s) scanned.` |
| `check-core-paths.sh` | 0 | `OK: all 50 strict/adopt-on-edit core-paths entries resolve at HEAD.` |
| `check-test-wiring.sh` | 0 | `OK: all test scripts are wired (67) or exempted with a reason (1) …` |
| `check-ci-parity.sh` | 0 | `OK: reference, GitHub Actions, GitLab agree …` |
| Rule 3 / Rule 1 greps | — | zero hits each |

Conventions: Gitmoji title unchanged ✅ · **no closing keyword** — `closingIssuesReferences` still `[]` ✅ · Rule A logbook on #756 ✅ · branch unchanged ✅ · version bump still not triggered (same two files, neither a provenance carrier) ✅ · CLI matrix still not triggered ✅ · body counts internally consistent: "Eight new cases, `cc` through `jj`" = 8, "Thirteen mutations" = 13 rows ✅.

---

### Non-blocking observations

No `class:` tag — notes, not findings to route.

1. **The precondition idiom under-tests its own stated capability, in `cc`, `hh` and `jj` alike.** All three FAIL with *"could not create an ed25519 key — ssh-keygen signing support (OpenSSH >= 8.2) is a precondition"* (`:1706`, `:2065`, `:2242`), but the command they guard is `ssh-keygen -t ed25519`, which tests **generation**. Nothing tests signing before the case depends on it. This is pre-existing — it arrived with `cc`/`hh` and `jj` inherits it, so it is not a regression in this revision, and CI has just shown `ubuntu-latest` clears both. Worth a follow-up rather than a fix here: one `ssh-keygen -Y sign` smoke check in the shared precondition would make all three fail-fast with the true cause, and would retire the finding above at the same time.
2. **The body's reading pointer excludes one of the three things it asks you to read.** *How to read this PR?* sends the reviewer to `scripts/sync-from-upstream.sh:645-675`; line 645 is mid-sentence in a comment and 675 is the `fi` of the failure branch, so the post-condition (`:679-700`) and `update-ref` (`:702`) — the subject of its own point 2, "the refusal is checked *before* `update-ref`" — fall outside the range. Identical at both heads, so pre-existing and not introduced here; I missed it last round. `:637-707` covers the block.
3. **Two wordings carried over from the previous head**, both already noted and neither re-raised: *"Nothing in the suite reads git's line"* is true of positive assertions but `dd`'s `[Ss]ign` grep (`:1839`) does span git's fetch stderr (measured clean in `C`, `fr_FR`, `de_DE`), and *"bit-for-bit unaffected"* overstates by the exit-code change from git's 128 to 1 on the pre-existing non-signing `commit-tree` failure path.
4. **The `docs/adoption-guide.md` paragraph is still deferred** and still worth a tracked follow-up rather than a body note.

---

### What would clear this

One sentence in the body. Drop the "only if" — the check fires whenever the fixture stops producing the `gpgsig-sha256` header, whatever the reason — and this is an APPROVE. The code needs no change: the fix is right, `jj` closes it, the three `M10` rows prove both readers are covered independently, the relocated premise assertion is inert and still detects what it was written for, all thirteen mutation rows reproduce at this head, the suite is 65/65 in three local configurations and in CI, and every gate passes.

hcross:
## Verdict: APPROVE

Body-only re-review at head `735bd76` — unchanged, and re-verified as unchanged rather than assumed: both blobs still hash to `087a32ac5a21…` and `cb4d8ba9b60f…`, the values I measured the suite and the thirteen mutation rows against. No code moved, so nothing from the previous pass is re-derived except where the body makes a claim I had not measured.

**Zero findings.** The finding is closed and all three non-blocking notes are taken.

---

### CI status

All 13 required checks **pass** on `735bd76`, `check-components` among them — the job that runs the suite and reports `Results: 65/65 passed` with `PASS case-jj`. Same run as the previous pass; the body edit does not disturb it.

---

### The finding — closed

The paragraph no longer claims the premise check fires *only* on a naming change (`grep -cF 'fires only if'` → 0). What replaced it is accurate against my measurements, clause by clause:

- *"two disjuncts, and the first fires whenever the probe cannot produce a signature at all"* — matches `test-sync-from-upstream.sh:2293`.
- *"on the part outside that set its message names a cause that did not occur"* — the misattribution, stated rather than glossed over.
- *"`jj` reddens on five lines, the premise first"* — exactly what `JJ_NOSIGN` produced: the premise line, then the four behavioural ones.
- *"An earlier draft claimed it fires only on a naming change; that is false, and it is the same defect class the paragraph documents fixing, one level down"* — the self-location is right, and worth more than a silent correction would have been.
- *"Splitting the disjuncts would let the stronger claim stand and is the natural follow-up; it is not done here"* — the honest disposition. I did not ask for the split and still don't.

**The added JJ_SHA1 row is a genuine strengthening, not a restatement.** It now records that with the fixture silently non-SHA-256 all four behavioural assertions stay green, so the premise check plus the format sanity loop are the only things standing between `jj` and a PASS that exercises the wrong header name. That is a better argument for the check than the paragraph previously made, and it belongs in the body rather than in a review comment that ages out of view.

---

### The three notes — all taken, and one of them I re-measured

**1. The reading pointer.** Now `:637-707`, with the narrowing recorded. Verified at this head: `:637` is the first line of the spec 0122 comment, `:707` the success `echo`, `:708` the closing `fi` — so the range brackets the guard block exactly. The post-condition (`:679-700`) and `update-ref` (`:702`) it previously excluded are both inside it now.

**2. The locale claim.** Now *"No assertion depends on git's wording for a positive match"*, with `dd`'s grep named as spanning git's fetch stderr, the three-locale measurement stated, and the overreach called out as wrong rather than quietly dropped. Matches what I measured: two git lines in `C`, `fr_FR` and `de_DE`, none matching.

**3. The R3 sentence — I measured this one, because the replacement asserts a number I had only reasoned about.** My previous pass said the exit code moves from 128 to 1 without ever running it, and the body now states it as fact, so it needed measuring before I could approve it. Correct baseline is `main`, not `54cf59b` — the wrapping landed with this PR, so `54cf59b` already exits 1. I extracted `main`'s script and confirmed it is the diff's pre-image (`0781424c2ccc…`, matching the `index 0781424..` line), then ran both against one fixture: a non-signing fork (`commit.gpgsign false`, no key) whose `commit-tree` fails for a reason unrelated to signing, via a pass-through `git` shim that appends an invalid `-p` so the failure and its status are real git's.

```text
=== PRE-PR-main   (blob 0781424)
    exit_rc=128   tip_moved=no   gpgsign=false
    (script printed no Error: line of its own)
    (no line mentions 'sign')
=== NEW-735bd76   (blob 087a32ac)
    exit_rc=1     tip_moved=no   gpgsign=false
    script line: Error: --preserve-history could not create the graft commit; see the git error above.
    (no line mentions 'sign')
```

The sentence holds in all three of its parts. **128 → 1**, confirmed. *"On a path that already failed"* — confirmed: git's `commit-tree` fails in both, the tip is unmoved in both. *"Nothing that succeeded before fails now"* — every new exit is gated on `GRAFT_SIGN=true` or on an unparseable value, neither reachable for a `false`/unset fork, and `dd` and `ii` pass. One thing worth noting beyond the claim: neither run emits a line matching `sign`, so the non-signing path stays free of signing noise on the *failure* path too, not only on the success path `dd` asserts.

---

### Convention compliance

Unchanged and re-checked at this head: Gitmoji title · **no closing keyword** (`closingIssuesReferences` still `[]`) · Rule A logbook on #756 · branch naming · version bump not triggered · CLI matrix not triggered. Body internally consistent: "Eight new cases, `cc` through `jj`" = 8, "Thirteen mutations" = 13 table rows, and the table is byte-identical to the one I verified row by row at this head. All three superseded phrasings are gone — `fires only if`, `bit-for-bit`, `Nothing in the suite reads` all return 0 occurrences.

Linters and gates were run against `735bd76` in the previous pass and the code has not moved since: `lint-shell.sh` 0 · `shellcheck` 0 at both severities · `check-bash32-portability` / `check-core-paths` / `check-test-wiring` / `check-ci-parity` all 0 · Rule 1 and Rule 3 greps zero hits.

---

### Termination

The ADR-0010 criterion is met: **APPROVE, zero findings of any class, CI green on the head reviewed** (`735bd76`, 13/13 including `check-components`).

One procedural note, since the next step is a merge and not mine to take: `AGENTS.md` → *Branching Strategy* requires the user's formal permission immediately before `gh pr merge`. This verdict clears the REVIEW stage; it is not merge authorization.

---

### What this ticket produced, worth keeping

Not a finding, just the part I would want a future reader of #756 to notice. Three defects in this batch were the same shape — a property checked on one instance and written as a property of the whole: the signature header name checked on SHA-1 and written as *the* header name, a conclusion inferred from a correct argv measurement and filed under a heading asserting the list was measured, and an assertion whose message named the one cause its author had in mind rather than the set its own disjunct admits. Each was found by making something *fail on purpose* — a reversion mutation, a stub swap, a broken signer — never by reading. The suite that came out of it now carries `M10script`/`M10helper` proving its two readers independently, and a precondition that fails loudly instead of skipping, which is what turned a CI log into a capability measurement for `ubuntu-latest`.

PR #807, opened 2026-08-11
🐛 Install Antigravity components where the assistant finds them

hcross:
Installs Antigravity skills and agents where the assistant actually finds them — `~/.gemini/config/`, the documented customization root — fixes the installer bug that placed no agent at all, and migrates machines set up under the superseded layout without touching anything the framework did not install. Implements spec 0123 for issue #761.

## ⚠️ The install location changes for existing users

Skills and agents move from `~/.gemini/antigravity-cli/{skills,agents}` to `~/.gemini/config/{skills,agents}`. A setup run **removes** the framework's own components from the superseded location. Content there that the framework did not install is left alone — that is spec 0123 R9, and it is the property the migration is most carefully tested against, because getting it wrong destroys a user's own files.

## How to read this PR?

**Start with the measurement, because this ticket exists because a path contract was inferred from documentation instead.** A live probe against `agy` 1.1.11 ran twice with identical verdicts, nine sentinels plus a duplicate-name cell plus a positive control:

| Location | Skills | Agents (flat) | Agents (directory) |
|---|---|---|---|
| `~/.gemini/antigravity-cli/` — what we installed | not found | not found | not found |
| `~/.gemini/config/` — the documented root | **FOUND** | **FOUND** | **FOUND** |
| `~/.gemini/skills`, `~/.gemini/agents` — Gemini CLI roots | not found | not found | not found |

Two consequences settle decisions the plan had deliberately routed through this step. The **directory** shape is discovered, so the install can keep the shape `scripts/build-components.sh:785` already stages, and **no `[GAP]` is needed on matrix row 4**. And a duplicate agent name across two roots is **doubled**, not shadowed and not refused — recorded in the runbook as an observation rather than forced into a rename, since both copies are the same crewrig component rendered for two CLIs.

**This corrects the probe recorded earlier on the issue.** That one used `agy agents` alone and concluded the application-data root *was* a discovery location for agents. Both readings reproduce: `agy agents` enumerates a **registry** and lists all three roots; the forced-choice model query measures **in-session availability** and finds only the documented root. Two instruments, two questions, and the cheaper one was the permissive one. The stricter governs.

Then read, in order:

1. **`scripts/lib/common.sh`** — `install_antigravity_tier_to_home()`, the shared installer. It is a **sourced** helper rather than a spawned script, deliberately: that is what lets `/bin/bash <suite>` genuinely exercise it on 3.2.57 instead of certifying only the harness.
2. **`scripts/setup-antigravity-interactive.sh`** — the flat-glob bug. It globbed `"$staging/agents"/*.md` while the build stages **directories**, so the glob matched nothing, `[ -f ] || continue` swallowed it, and the step exited 0. **The failure mode was silence**: no agent installed, nothing reported.
3. **`scripts/manage-antigravity-component.sh`** — the per-component surface. Note what is *not* changed: `ANTIGRAVITY_HOME` at `:30` still resolves `rules/` (`:154`) and `settings.json` (`:102`) — line numbers at **this head**, `02170ee`, both explicitly out of scope in spec 0123. Only the skills destination moves. The one-line fix would have silently relocated two kinds nobody probed.
4. **`scripts/probe-antigravity-discovery.sh`** and **`docs/runbooks/antigravity-discovery-probe.md`** — R13/R14. The procedure is re-runnable, and it returns **three** verdicts: FOUND, NOT-FOUND, and **INDETERMINATE**. That third one is not decoration — `agy agents` hung for over two minutes during the first probe, and a transient non-answer read as an absence would produce a confidently wrong record.

## How to test this PR?

```sh
/bin/bash scripts/tests/test-antigravity-component-install.sh   # 63 passed, 0 failed
bash      scripts/tests/test-antigravity-component-install.sh   # 63 passed, 0 failed
/bin/bash scripts/tests/test-antigravity-discovery-probe.sh     # 33 passed, 0 failed
bash      scripts/tests/test-antigravity-discovery-probe.sh     # 33 passed, 0 failed
bash      scripts/tests/test-component-tier-resolution.sh       # 26/26
```

Both new suites use `"${BASH:-bash}"` from the start, so the 3.2 run reaches the subject rather than only the harness — the repo-wide absence of that idiom is issue #798 and is not swept here.

**Twenty-five mutations, all as predicted** — twenty-two toward red and three that must stay **green**. The green side is the half a red-only pass cannot reach, and it is what proves the migration does not over-fire. The table below lists the ten from PLAN v4's step 14; the remaining twelve are grouped after it:

| | Mutation | Observed |
|---|---|---|
| (a) | flat `*.md` glob restored | 7 fails, including "no Installed line for an agent that IS present" **and** the R6 case |
| (b) | `placed = staged` | 5 fails (R4 tallies + R6) |
| (c) | R4 report deleted, R6 left intact | 3 fails, all R4 — R6 stays green, the separation the plan required |
| (d) | both surfaces retargeted at the superseded root | 3 fails (both setup assertions + R7) |
| (e) | missing forced-choice line folded into NOT-FOUND | 6 fails, both call shapes |
| (f) | plain `kill "$pid"` | 2 fails; a marker child **survived in both shapes**, PID reported |
| (g1)/(g2) | predicate degraded to name-only / provenance-only | fixture (i) removed / fixture (ii) removed, skill **and** agent shapes |
| (h) | frontmatter-exit rule dropped | fixture (i) removed, both shapes |
| (i) | name set narrowed to `library` | org **and** community orphans survived |
| (j) | `yaml_field` reader with `yq` off `PATH` | 9 fails |
| **(k)(l)(m)** | **green side** — user directory satisfying exactly one half; a staged skill's body rewritten without touching frontmatter; an empty staged agents directory beside a placed skill | **64 passed, 0 failed on both interpreters** |
| (n1)/(n2) | per-component call site, structural | both red |

**The eleven mutations added in review**, each proving an assertion this PR gained rather than shipped with:

| Group | Rows | Result |
|---|---|---|
| per-component call site, structural | n1 n2 | 2/2 RED |
| staging parity across the helper boundary | p1 p2 p3 p4 p5 | 5/5 RED |
| literal, whole-line name test | q1 q2 q3 | 3/3 RED |
| the reporting nit — silence the `Migrated away` line | q4 | RED |

Row **p4** is the one to re-run: it flips the build target `antigravity` → `gemini` and reddens the *second* Antigravity assertion in `test-setup-ensure-tier-built.sh`. Before this PR's fix that assertion was never reached at all — the guard's `continue` skipped both — so p4 is what makes it *executing* rather than merely *present*.

Mutation **(f)** is worth running yourself: plain `kill "$pid"` reaches the wrapper, not the process it forked, so a bounded-out `agy` outlives the probe — including the `EXIT` trap that removes the sentinels. `set -m` plus `kill -TERM -- "-$pid"` leaves nothing behind.

Gates, each captured as a plain exit status and never a pipeline's: `check-bash32-portability.sh`, `check-no-machine-paths.sh`, `check-core-paths.sh`, `check-test-wiring.sh`, `check-ci-parity.sh`, `build-ci.sh --check`, `build-docs-index.sh --check`, `build-components.sh --check`, `task lint-markdown`, `task spec:lint` — all `rc=0`.

**The migration predicate is a conjunction**: a directory at the superseded location is removed iff its name is one the framework serves **and** its `SKILL.md`/`AGENT.md` frontmatter carries `metadata.provenance.canonical`. Its fixtures discriminate **each half separately** — one satisfying only the name half, one satisfying only the provenance half — because a fixture failing both halves tests "the sweep does not delete arbitrary junk", not the AND. `user-validate`, `developer`, `tester` and `architect` are all plausible names for a user's own directory, and a degraded predicate destroys their content.

**The name set comes from `artifacts/{library,community,org}/`, not from `dist/`.** `ensure_tier_built` returns 0 on directory existence alone, so a stale staging tree would have orphaned any component that moved tiers after it was built — which is not hypothetical: `user-validate` moved into `library` on `7a10778` after `dist/library` was last built, and a `dist/`-keyed predicate removed two of three components on this machine and left the third behind.

**Frontmatter is read by a portable reader, not `yaml_field`.** That helper shells out to `yq`, which **neither install surface requires**, and it ends `|| echo ""` — so a missing `yq` would have yielded an empty name set, a migration that removes nothing, and **exit 0**. An empty name set is now an error.

**The membership test is `grep -qxF`, not `grep -qx`.** `$declared` comes off disk and was being compared as a *regexp*, so a directory named `a.b` would satisfy a served name `axb`, and a plain `harness` would satisfy `harness-report` by prefix — and that predicate's true branch **deletes**. Both are directories the code explicitly promises to preserve and report as residue.

**It shipped in an earlier revision of this PR with nothing testing it**, which is the more useful half. Cold review measured all three degradations — drop `-F`, drop `-x`, drop both — passing the entire suite on both interpreters. The cause was not laxity: **every component in this repository is kebab-case, so every fixture was, so nothing could exercise the discipline the code depends on.** A fixture corpus that shares a naming convention cannot test a property only non-conforming names reach.

Four fixtures now do: `a.b` against served `axb`; `harness` against served `harness-report`; `a.b-agent.md` in the flat agent shape, because `${item%.md}` strips one extension and a dotted name reaches the same test through code the directory branch does not share; and **a true positive in the same run**, without which a section where everything survives would pass equally on a migration that does nothing at all.

Scoped honestly: the provenance half of the conjunction holds throughout, so a user's own directory was never at risk. The victim class is a retired or user-edited *framework* component.

**Provenance carrier verified against the real tree — exhaustively, and it matters more than it sounds.** All four built Antigravity components (three skills, one agent) carry `metadata.provenance.canonical` **in frontmatter**, read through the shipped helpers rather than a hand-rolled equivalent; `grep -c -F 'crewrig-provenance:'` returns 0 on all four. The population is genuinely closed, not merely the built subset — `artifacts/community/*` and `artifacts/org/*` are empty, so `library` is the whole served corpus for both kinds.

That check was not ceremony. **`docs/cli-matrix.md` row 4b's Antigravity cell says the carrier is an HTML comment, "same carrier as Gemini (spec 0053)", and that is wrong about the current build** — verified structurally, not by sampling: the Gemini agent path at `build-components.sh:735` passes no `$source` and emits the HTML comment (its own comment says it bypasses `inject_provenance` deliberately), while the Antigravity path at `:785` **does** pass `$source`, so `metadata:` is spliced into frontmatter and no HTML comment is emitted anywhere. Had the reader been built from that row instead of from the tree, it would have looked for a comment, found none on any agent, and **the agents half of the migration would have been a silent no-op passing every fixture** — because the fixtures would have carried the same wrong assumption. Correcting row 4b is deliberately **not** done here: spec 0123 R10 binds matrix statements about where components are *placed, staged or found*, and 4b records a carrier. It is filed separately.

**Blast radius.** CLI-matrix trigger **fires** — both scripts match `scripts/{build,install,setup,import,manage}-*.sh` — so `docs/cli-matrix.md` rows **10, 12 and 27** are updated in the same diff; rows 3 and 4 are build/staging statements and are unchanged, and row 4 correctly carries no `[GAP]` because the probe found the directory shape. Three CI surfaces move together (`build.yml`, `ci-capabilities.yml`, regenerated `.gitlab-ci.yml`). No `artifacts/**` source, so no build outputs and no version bump. `.crewrig/core-paths.txt` unchanged — every touched path falls under an existing entry.

**One CI-only hazard checked rather than assumed.** `test-antigravity-discovery-probe.sh:236` asserts its own interpreter propagation with a BRE whose portability across grep implementations was not obvious. Verified in a `debian:stable-slim` container that GNU grep 3.11 matches the present case and rejects the absent one, so the assertion cannot be red in CI and green on a developer's machine. *(An earlier draft described the pattern as containing an unmatched `\{`. It does not — the bytes are `"\${BASH:-bash}" "\$SUBJECT"`, where the backslash escapes the `$` and `{` is bare. The verification stands; the characterisation of why it needed verifying was wrong.)*

**The `[GAP-soft]` marker** was chosen over inventing a new kind. The matrix carries `[GAP]`, `[GAP-soft]`, `[GAP-confirmation]` and `[GAP-content]`, and **nothing defines any of them** — in a vocabulary nothing validates, a fifth kind introduced by a bug fix would carry its meaning only here, and `[GAP-content]`'s single use is the standing example of a one-off nobody can now distinguish from its neighbours.

**Two blind spots this PR hit, and the rules that would have caught them.** Both were found after the code was written and correct, and neither was visible from inside the suites that had the gap.

*A change that **moves** code has a blast radius shaped by who references the **old** location, and no amount of testing the new location can see it.* Moving the tier-install into `scripts/lib/common.sh` silently blinded `test-setup-ensure-tier-built.sh`, whose parity assertion greps the setup script for a function that no longer lives there — and whose `continue` then skipped **both** remaining Antigravity assertions, guarding exactly the silent no-install mode this ticket was opened about. CI caught it; the mutation pass could not, because mutation proves *these* assertions have teeth and says nothing about assertions elsewhere that the change breaks. `check-test-wiring.sh` tells you which suites CI runs; it does not tell you which of them care about your diff. This does:

```sh
grep -l -F "$changed_file" scripts/tests/*.sh   # for every file the diff touches
```

Twenty-two suites for this change, all run, exactly one casualty — so the blast radius is **measured rather than assumed**, before the push rather than after.

*A fixture corpus that shares a naming convention cannot test a property that only non-conforming names exercise.* Every component in this repository is kebab-case, so every fixture was, so all three degradations of the served-name test passed the entire suite. The blind spot was in what the fixtures could **reach**, not in how hard they were run — the same shape as the first, wearing different clothes.

**Plan history.** Four revisions, three cold-review rounds, nine findings — every one produced by prototyping and mutating rather than reading. Two are worth carrying: the migration predicate was originally keyed to a stale build artifact, with the counterexample sitting on the machine the ticket was opened from; and the fixtures could not tell the correct predicate from either of its degraded halves, which the reviewer established by building all three and running the plan's own cases against each.

Issue #761 is the logbook per `AGENTS.md` → *Logbook Issues → Rule A*, closed manually after merge per Rule C — this body carries no closing keyword. Spec 0123's `status` moves `approved` → `implemented` in its own metadata-only pull request afterwards.


hcross:
## Verdict: REQUEST CHANGES

Cold review — no authoring context. Head `cf1c0ee`, PR `updatedAt` 2026-08-11T06:24:59Z (unchanged across this review). Branch contains `origin/main` tip `e3fdbc8`, so nothing here is scoped to a stale guard revision.

Two blocking findings, both `class: tech`. Everything else in this diff verified as claimed — including the migration predicate, which I attacked three ways and could not break.

---

## CI status

**FAILING — required check `check-components`.**

| Check | Result |
|---|---|
| `check-components` | **fail** (6m12s) — [job 93695972638](https://github.com/crewrig/crewrig/actions/runs/31464999989/job/93695972638) |
| build, check-agents-size, check-ci-parity, check-extension-version-bump, check-feedback-routing, check-skill-versions, check-spec-id-reserved, gitlab-ci-check, grep-anti-patterns, lint-markdown, lint-specs, test-harness-curate | pass (12) |

The failing step is `Test setup ensure-tier-built regression (spec 0107)`. Because the job shell is `bash -e`, the four steps after it never executed. I ran those four locally at `cf1c0ee` — `test-check-gemini-overlay-enrollment.sh`, `test-check-bash32-portability.sh`, `test-reserve-spec-id.sh`, `test-check-spec-id-reserved.sh` — all `rc=0`. The failure below is the only one.

---

## Correctness

### Finding 1 — the spec-0107 parity guard is dead for Antigravity, and the suite says so `class: tech` **BLOCKING**

`scripts/tests/test-setup-ensure-tier-built.sh` fails at head, in CI and locally:

```text
FAIL: setup-antigravity-interactive.sh: no 'local staging=' assignment found in its tier-install function
RESULT: 14 passed, 1 failed
```

That suite's §4 pins, per setup script, that the staging path handed to `ensure_tier_built` is the same one the script's own tier-install function reads from. It resolves the install function by grepping the setup script for two things:

- `scripts/tests/test-setup-ensure-tier-built.sh:159` — `install_tier(_skills)?_to_home [a-zA-Z0-9_]+`, for the tier literal.
- `scripts/tests/test-setup-ensure-tier-built.sh:169` — `local staging="[^"]*"`, for the staging pattern.

This PR moved that function out of the setup script into `scripts/lib/common.sh` and renamed it `install_antigravity_tier_to_home`. Neither grep can reach it any more: the new name does not contain the substring `install_tier_to_home`, and `local staging=` now lives at `scripts/lib/common.sh:1731`. `install_pattern` comes back empty, and the `continue` at `:205` skips **both** of Antigravity's remaining assertions — the `build_target is 'antigravity'` check at `:208` as well as the staging-path-match check. The other three CLIs contribute two `ok`s each; Antigravity now contributes one `FAIL` and zero.

The property itself still holds today — `setup-antigravity-interactive.sh:352` passes `"$REPO_DIR/dist/library/.agents"` and `common.sh:1731` computes `"$repo_dir/dist/$tier/.agents"`, which agree at `tier=library`. So this is a guard that went blind, not a live path bug. But it went blind for exactly the one CLI this PR changes, and the drift it exists to catch — build the tier at path A, install from path B — is the silent no-install failure mode this whole ticket was opened about.

`test-setup-ensure-tier-built.sh` is not in the twelve changed files. Moving a helper into `common.sh` is the sanctioned pattern here (spec 0116 R17, matrix row 10) — the resolver needs to follow it there, the same way the move was accounted for in row 10.

### Finding 2 — the membership test's matching discipline is unpinned; three degradations pass the suite green `class: tech` **BLOCKING**

`scripts/lib/common.sh:1947` is the name half of the delete predicate:

```sh
if printf '%s\n' "$names" | grep -qxF -- "$declared"; then
```

The body makes this a headline: *"The membership test is `grep -qxF`, not `grep -qx`. `$declared` comes off disk and was being compared as a regexp… and that predicate's true branch deletes."* That is correct, and I confirmed the fix is load-bearing. It is also entirely untested. I built the three degradations and ran the shipped suite against each, on both interpreters:

| Mutation at `common.sh:1947` | Suite result |
|---|---|
| `grep -qxF` → `grep -qx` (drop `-F`) | **55 passed, 0 failed** |
| `grep -qxF` → `grep -qF` (drop `-x`) | **55 passed, 0 failed** |
| `grep -qxF` → `grep -q` (drop both) | **55 passed, 0 failed** |

All three are silent regressions with real consequences. Against a served set `{axb, harness-report}`:

| Directory at the superseded root | shipped | `-F` dropped | `-x` dropped |
|---|---|---|---|
| `a.b` — provenance, regex-matches `axb` | SURVIVED (residue) | **DELETED** | SURVIVED |
| `harness` — provenance, substring of `harness-report` | SURVIVED (residue) | SURVIVED | **DELETED** |
| `a.b-user` — no provenance | SURVIVED | SURVIVED | SURVIVED |

The provenance half holds throughout, so arbitrary user content is never at risk — I want to be precise about that rather than overstate it. The victim class is a directory carrying framework provenance whose declared name is not served: a retired component, or a user's edited copy of one. Those are precisely what `common.sh:1868-1871` promises to preserve and report — *"a leftover that carries provenance but matches no served name is REPORTED by name with its removal command rather than removed — the one accepted failure mode here"*. Both degradations silently delete what that paragraph guarantees will survive, and the residue report drops from 2 lines to 1 with no other signal.

Why the fixtures miss it: every name in the suite is kebab-case with no metacharacter and no prefix collision — `alpha-agent alpha-skill beta-skill community-thing declared-name harness-curator harness-report my-own-notes one-skill other-skill real-agent real-skill retired-thing some-source-dir tester tester-agent user-validate whatever`. That is the body's own diagnosis turned on its test suite: *"Latent today only because names happen to be kebab-case, which is a property of the current corpus rather than of the code."*

§7 was purpose-built to discriminate each half of the conjunction, and §13 asserts call-site arguments structurally because "an emptied argument once survived a whole suite" (spec 0116 delta-01 R24). The same standard applied to `:1947` costs one fixture: a served name with a regex metacharacter, plus one whose name is a proper prefix of a served name.

---

## What I verified and could not break

**The migration predicate.** I built it three ways and ran the shipped fixtures against each. The fixtures discriminate — this is not a suite that would accept a degraded predicate:

| Mutation | Observed, both interpreters |
|---|---|
| (g1) conjunction → name-only | 52 passed, **3 failed** — fixture (i) removed, skill **and** agent shapes |
| (g2) conjunction → provenance-only | 51 passed, **4 failed** — fixture (ii) removed, plus §10 and §12 |
| (h) frontmatter-exit rule dropped from `_antigravity_has_provenance` | 52 passed, **3 failed** — fixture (i) removed, both shapes |

Fixture (i)'s body token **is** indented — `test-antigravity-component-install.sh:107-109` writes the `metadata:`/`provenance:`/`canonical:` block at 4/6/8 spaces. That is what makes (h) redden: a column-0 or mid-prose token would not match the indent-requiring pattern the mutant retains.

**My own fixtures, independent of the suite.** A user directory satisfying exactly one half, each way round, plus the cases nobody wrote a test for:

- served name + no provenance → SURVIVED, with its `private.txt` intact.
- provenance + unserved name → SURVIVED, reported as residue.
- `a.b` on disk vs served `axb` → SURVIVED. With `-F` removed → DELETED, which is how I know the fix is real.
- `axb` on disk vs served `a.b` → SURVIVED.
- A **symlink** at the superseded root named for a served component, pointing at a directory of user data → target untouched. `find -P` does not descend it, `rmdir` fails, and it surfaces as `(removal incomplete)` residue. Undocumented, untested, and correct.
- True positive still DELETED.

**End-to-end with the real corpus.** Built `dist/`, installed the four real components at `~/.gemini/antigravity-cli/`, added adversarial user content, ran the real migration: all 4 framework components removed; `tester` + `notes.md`, `user-validate2` (frontmatter `name: user-validate` — the served name — but no provenance), `no-marker`, `stray.txt`, `agents/harness-report.md` and `rules/` all survived. The name half alone does not kill anything.

**The rest of the mutation table**, all as the body predicts:

| | Mutation | Observed, both interpreters |
|---|---|---|
| (a) | flat `*.md` agents glob restored | 48 passed, **7 failed** — includes "no Installed line for an agent that IS present" and the R6 case |
| (b) | placement asserted rather than observed | 49 passed, **6 failed** (R5 + R4 tallies + R6) — body says 5; my mutation shape differs slightly, direction identical |
| (c) | R4 report deleted, R6 intact | 52 passed, **3 failed, all R4** — R6 stays green, the separation the plan required |
| (d) | both surfaces retargeted at the superseded root | 52 passed, **3 failed** (both setup assertions + R7) |
| (e) | missing forced-choice line folded into NOT-FOUND | 27 passed, **6 failed**, both call shapes |
| (f) | plain `kill "$pid"` | 31 passed, **2 failed**; a marker child **survived in both shapes**, PID reported |
| (i) | name set narrowed to `library` | 53 passed, **2 failed** — org **and** community orphans survived |
| (j) | name reader returns empty (`yq`-less `yaml_field`) | 46 passed, **9 failed** |
| (n1) | `PLACED_NAMES+=` removed | 54 passed, **1 failed** |
| (n2) | manage retargeted at `ANTIGRAVITY_HOME` | 54 passed, **1 failed** |

**The green side, which is the half a red-only pass cannot reach.** Shipped code: `install` 55/55 and `probe` 33/33 on `/bin/bash` 3.2.57 **and** `bash` 5.3.15, plus 55/55 and 33/33 in a `debian:stable-slim` container (bash 5.2.37, GNU coreutils). No over-firing in any of the independent fixtures above. Mutation (f) reversed: no marker process outlives the shipped `set -m` + `kill -TERM -- "-$pid"`.

**Specific claims, checked rather than accepted:**

- **`ANTIGRAVITY_HOME` still serves the two excluded kinds.** At head: `rules/` at `manage-antigravity-component.sh:154`, `settings.json` at `:102`, both still under `~/.gemini/antigravity-cli`. Only skills moved, to `AGY_CUSTOMIZATION_ROOT/skills` at `:136`. (The body's `:22`/`:129`/`:85` are pre-diff line numbers.)
- **The provenance carrier.** Both halves hold. Structurally: `build-components.sh:737` passes no `$source` for the Gemini agent and emits the HTML comment, `:785` passes `$source` for Antigravity so `metadata:` lands in frontmatter. Empirically, after a real build, all four Antigravity components (3 skills + 1 agent) carry `metadata.provenance.canonical` in frontmatter and return 0 for `grep -c -F 'crewrig-provenance:'`. Running the shipped `_antigravity_has_provenance` against the four real built files returns YES on each, and `_antigravity_framework_names` returns exactly `harness-curator harness-report user-validate` / `harness-curator`. Row 4b is wrong about the current build; the reader is right.
- **`install_antigravity_tier_to_home()` is sourced, not spawned** — the suite prints `Interpreter under test: 3.2.57(1)-release` under `/bin/bash`, so the 3.2 run reaches the subject. The probe suite does use `"${BASH:-bash}"` at `:104`, `:201`, `:222`. The install suite has no such idiom because it sources rather than spawns; the body's "both new suites" is loose there, though the property it stands for is verified.
- **R14's three verdicts.** `probe-antigravity-discovery.sh:282-291` returns INDETERMINATE for a non-`OK` status and for a missing `<name>=` line, before any YES/NO mapping. Mutation (e) reddens 6 cases across both call shapes.
- **`test-antigravity-discovery-probe.sh:236`.** Verified in `debian:stable-slim` with GNU grep 3.11: the present case exits 0, the absent case exits 1 — the assertion discriminates in CI. One correction to the body: the pattern contains no `\{`. The bytes are `"\${BASH:-bash}" "\$SUBJECT"` — the backslash escapes `$`, and `{` is bare, which BRE treats as literal. The conclusion stands; the characterisation does not.
- **CLI matrix.** Exactly rows 10, 12 and 27 changed. Rows 3 and 4 are byte-identical to `e3fdbc8`. Row 4 carries no bare `[GAP]` — its pre-existing `[GAP-confirmation]` is unchanged. Spec R12 is satisfied explicitly in row 10: *"Skills — documented AND observed"*, *"Agents — OBSERVATION ONLY"*, stated per kind.
- **Sourcing `common.sh` into `manage-antigravity-component.sh`** is safe: no function-name collision with `component-resolve.sh` or the script itself, no top-level executable statement, and no top-level assignment that clobbers `REPO_DIR`, `ANTIGRAVITY_HOME`, `AGY_CUSTOMIZATION_ROOT`, `MODE`, `TYPE`, `NAME` or `DEST`.

---

## Convention compliance

All clear.

- **Naming** — `🐛 Install Antigravity components where the assistant finds them`, Gitmoji, no conventional-commit prefix.
- **PR format** — purpose sentence, *How to read this PR?*, *How to test this PR?*, *Detailed description (for agents)* all present.
- **Logbook (Rule A)** — issue #761 is the logbook, 15 comments, incremental.
- **No closing keyword** — `closingIssuesReferences` is `[]`, verified at head. Correct for Rule C manual close.
- **Version bump** — no `artifacts/**` source touched, so the rule does not fire; `check-skill-versions` passes locally and in CI.
- **CLI matrix maintenance** — trigger fires on `scripts/{setup,manage}-*.sh`; matrix updated in the same diff. `.crewrig/core-paths.txt` needs no change, `check-core-paths.sh` rc=0.
- **CI parity** — the two new suites are wired into all three surfaces: `.github/workflows/build.yml`, `.gitlab-ci.yml`, `ci/ci-capabilities.yml`. `check-ci-parity.sh` and `check-test-wiring.sh` both rc=0.

Local gates, each captured as a plain exit status rather than a pipeline's: `check-ci-parity` `check-test-wiring` `check-bash32-portability` `check-no-machine-paths` `check-core-paths` `check-skill-versions` `build-ci.sh --check` `build-docs-index.sh --check` `build-components.sh --check` `task lint-markdown` — all **rc=0**.

---

## Test coverage

The install and migration helpers are covered to an unusually high standard, and the fourteen requirements of spec 0123 all map to assertions. Two gaps:

1. Finding 2 above — `:1947`'s matching discipline.
2. Non-blocking nit, `class: tech`: `test-antigravity-component-install.sh:260` captures `S6_OUT` and never asserts on it (shellcheck SC2034). §6 asserts filesystem state, which is stronger, so nothing is lost — but the capture reads as an intended assertion that was dropped.

---

## Linter findings

**`lint-markdown.sh`** on `docs/runbooks/antigravity-discovery-probe.md` and `docs/cli-matrix.md` — clean, exit 0. Verified the linter was live rather than silently degrading: `markdownlint` is on `PATH` at `/opt/homebrew/bin/markdownlint` and a deliberately malformed sentinel file returns exit 1 with three findings.

**`lint-shell.sh`** on the seven changed shell files — exit 1, but every item is `info` or `style` severity, and the two `warning`s are on pre-existing lines this PR does not touch (`common.sh:1398` SC2088, `manage-antigravity-component.sh:84` SC2115). The only new-code item is the SC2034 nit above. Two reported items are linter degradation, not defects: `scripts/lib/common.sh: not executable in git (mode 100644; expected 100755)` and `missing 'set -e'` — both correct for a sourced library.

---

**Routing:** both blocking findings are `class: tech` → DEV.

**Note on method** — `grep` on this machine is `ugrep` 7.5.0, which returns 0 silently on `${…}` patterns. Every figure asserted above was produced with `/usr/bin/grep -F`, a Python parser, or a direct read, and every exit status was captured plainly rather than from the end of a pipeline.

hcross:
## Verdict: APPROVE

Re-review at `02170ee`, `updatedAt` 2026-08-11T07:15:10Z. Both findings from my `cf1c0ee` review are closed, and both fixes survive attack. **Zero findings.** Three non-blocking observations at the end — none gates the loop.

Targeted, as scoped. `git diff cf1c0ee..02170ee` touches exactly two files, both test suites, and `cf1c0ee` is an unmodified ancestor — so no subject code moved and my earlier verification of the predicate, the fifteen original rows, the adversarial fixtures, the symlink case, `ANTIGRAVITY_HOME`, the provenance carrier, sourced-not-spawned, R14, the matrix rows and the conventions all carry forward unchanged. I did not re-spend on them, beyond re-running the eleven install-suite rows below because the suite itself grew.

---

## CI status

**All 13 required checks pass** on `02170ee` ([run 31467999312](https://github.com/crewrig/crewrig/actions/runs/31467999312)). `check-components` — the job that failed at `cf1c0ee` — now passes in 6m25s. No pending, no failing.

---

## Finding 1 — closed

`test-setup-ensure-tier-built.sh` now reports **16 passed, 0 failed** (was 14/1), and Antigravity contributes **both** of its assertions rather than one `FAIL` and nothing:

```text
ok: setup-antigravity-interactive.sh: ensure_tier_built build_target is 'antigravity'
ok: setup-antigravity-interactive.sh: ensure_tier_built staging path matches the install function's own staging path ($REPO_DIR/dist/library/.agents)
```

The fix reads across the file boundary rather than restoring a decoy into a script that no longer has a tier-install function. I agree that is the right call, and I attacked the three things that could make it dishonest.

**The call shape is genuinely fixed, not masked.** This was the one I most expected to find papered over. Instrumenting the resolver directly: `tier_arg=[library]`, raw pattern `[$REPO_DIR/dist/$tier/.agents]`, expected `[$REPO_DIR/dist/library/.agents]`, against an actual call site of `ensure_tier_built "$REPO_DIR" antigravity "$REPO_DIR/dist/library/.agents"`. No empty substitution, no `dist//.agents`. And mutation **d1** below proves the case still reddens if the call shape drifts again — it fails with the `dist//.agents` symptom, so that failure mode is reachable and visible rather than silently absorbed.

**Eight mutations, all as they should be:**

| | Mutation | Observed |
|---|---|---|
| **p4** | build target `antigravity` → `gemini` | **RED** — `build_target was 'gemini', expected 'antigravity'`. The assertion is *executing*, not merely present. |
| a1 | decoy `local staging="/bogus/after"` added later in `common.sh` | **green** — scoping holds, the decoy is not picked up |
| a2 | decoy `local staging="/bogus/before"` added above the function | **green** — scoping holds |
| a3 | decoy injected *between* the header and the real assignment | **RED**, naming `/bogus/inside` — fails loudly and legibly |
| b1 | `.agents` → `.antigravity` in the helper | **RED** — normalisation does not launder the path |
| b2 | `dist` → `build` in the helper | **RED** |
| c1 | helper renamed in `common.sh` | **RED** — `no 'local staging=' assignment found` |
| **d1** | library call-site tier quoted, so only the overlay shape remains matchable | **RED**, symptom `$REPO_DIR/dist//.agents` |

a1 and a2 green with b1/b2/a3 red is the pair that matters: the `awk` window is genuinely scoped to the function body, and it is scoped without becoming permissive about the path itself. The overlay call site is correctly excluded — its tier is `"$overlay_tier"`, quoted, so the `[a-zA-Z0-9_]+` class cannot reach it, and I confirmed only the `library` call matches.

`COMMON_LIB` is defined at `:44` with a `FATAL` guard at `:47`, so the cross-file read cannot silently resolve to nothing.

---

## Finding 2 — closed

§7b. All three degradations now redden, on both interpreters:

| Mutation at `common.sh:1947` | Before | Now |
|---|---|---|
| `grep -qxF` → `grep -qx` (drop `-F`) | 55/0 green | **60 passed, 3 failed** |
| `grep -qxF` → `grep -qF` (drop `-x`) | 55/0 green | **61 passed, 2 failed** |
| `grep -qxF` → `grep -q` (drop both) | 55/0 green | **58 passed, 5 failed** |

**The four fixtures each earn their place**, and I checked the two you flagged rather than taking them:

- `a.b` vs served `axb` — reddens under `-F` removal, not under `-x` removal. Discriminating.
- `harness` vs served `harness-report` — reddens under `-x` removal, not under `-F` removal. Discriminating in the opposite direction. Neither fixture is doing the other's work.
- **`a.b-agent.md`, flat shape** — reddens independently under `-F` removal (`agent shape: 'a.b-agent.md' was deleted by a non-literal name test`). It genuinely traverses the `elif [ -f ]` arm and the `rm -f` arm, which the directory branch does not share. Earns its place. One correction to the stated reason, in the observations below.
- **The true positive — I checked that it does.** I mutated `migrate_antigravity_superseded_components` to an unconditional `return 0` and ran the suite: **48 passed, 15 failed**, and among them `control: the genuinely served component was NOT migrated — this section would prove nothing`. Without that fixture, §7b's three survival assertions would all pass on a migration that does nothing whatsoever. With it, the section reddens. The control works.

**The generalisation is the right one.** *A fixture corpus that shares a naming convention cannot test a property only non-conforming names exercise* — that is exactly what I observed at `cf1c0ee`, where all eighteen fixture names were kebab-case with no metacharacter and no prefix relationship, and all three degradations passed the whole suite. The diagnosis holds.

**Scope stated accurately.** The §7b comment says the provenance half still holds, so a user's own directory is never at risk, and the victim class is a retired or user-edited framework component. That matches what I measured independently last round, and I am glad it was written down rather than inflated.

---

## The nit — taken, and the assertion is the right one

`S6_OUT` now asserts what the run *says*. I judged it two ways:

- **q4** — silence the `Migrated away` line while still removing the component: **62 passed, 1 failed**, `R8: a component was removed without being named`. The assertion executes.
- **q5** — the mirror I added, because an assertion that something is *never* named is vacuous unless over-announcement reddens it. I made the migration announce every entry it considers, including the user's own directory: **61 passed, 2 failed**, `R9: the user's own directory is named in the migration output`. Not vacuous.

The reasoning is sound: a removal under `$HOME` that is not announced is indistinguishable from one that never happened. Shellcheck confirms the original SC2034 is gone — the capture is now load-bearing rather than silenced.

---

## Regression check on the eleven originals

Re-run against the grown suite, since §7b changes the denominator. **No row went green:**

| Row | Result | Row | Result |
|---|---|---|---|
| (a) | 56/7 RED | (i) | 61/2 RED |
| (b) | 57/6 RED | (j) | 50/13 RED |
| (c) | 60/3 RED | (n1) | 62/1 RED |
| (d) | 60/3 RED | (n2) | 62/1 RED |
| (g1) | 60/3 RED | (h) | 60/3 RED |
| (g2) | 54/9 RED | | |

Two counts grew — (g2) from 4 to 9 and (j) from 9 to 13 — because §7b's fixtures also redden under those degradations. That is added sensitivity, not drift; every row kept its verdict.

(e) and (f) are probe-suite rows and `scripts/tests/test-antigravity-discovery-probe.sh` is untouched between the two heads, so those results carry unchanged.

---

## Suites, gates and linters at `02170ee`

| | macOS `/bin/bash` 3.2.57 | macOS `bash` 5.3.15 | `debian:stable-slim` (bash 5.2.37, GNU userland) |
|---|---|---|---|
| `test-antigravity-component-install.sh` | 63 / 0 | 63 / 0 | 63 / 0 |
| `test-antigravity-discovery-probe.sh` | 33 / 0 | 33 / 0 | 33 / 0 |
| `test-setup-ensure-tier-built.sh` | 16 / 0 | 16 / 0 | 16 / 0 |
| `test-component-tier-resolution.sh` | 26 / 0 | 26 / 0 | — (needs `jq`/`git`, absent in that image) |

Gates, each a plain exit status: `check-ci-parity` `check-test-wiring` `check-bash32-portability` `check-no-machine-paths` `check-core-paths` `check-skill-versions` `build-ci.sh --check` `build-docs-index.sh --check` `build-components.sh --check` `task lint-markdown` — all **rc=0**.

`lint-shell.sh` on both changed files: no `warning` or `error` severity, info/style only, and the SC2034 from last round is closed.

**Conventions.** Only test files changed, so no `artifacts/**` and no version bump owed. Gitmoji titles on both commits (`✅ Assert the tier-install staging parity across the helper boundary`, `✅ Pin the served-name test to a literal, whole-line comparison`). Body updated for the new head *before* review, and still carries **no closing keyword** — `closingIssuesReferences` is `[]` at `02170ee`.

---

## Observations — none is a finding, none gates the loop

**1. `${item%.md}` is not pinned by anything, and §7b's comment names it as the mechanism it does not exercise.** The comment reads *"`${item%.md}` strips one extension, so a dotted name reaches the same test through code the directory branch does not share."* The second clause is true and I verified it — the fixture reddens through the flat branch under `-F` removal. The first is not what that fixture exercises: `a.b-agent.md` carries `name: a.b-agent` in frontmatter, so `declared` comes from `_antigravity_frontmatter_name` and the `strip_ext` fallback at `common.sh:1940` never runs. Mutating `strip_ext="${item%.md}"` → `strip_ext="$item"` passes **63 / 0 on both interpreters**.

I checked the failure direction before deciding this was not a finding, and it is safe: on a flat `.md` with no frontmatter `name:`, the shipped code resolves the name from the filename and removes the component, while the mutant leaves it in place **and reports it as residue**. Under-removal with a printed `rm -rf`, never over-deletion. Worth one line in the comment rather than a fixture.

**2. The second mutation table lists `q4` twice.** It appears in the *literal, whole-line name test* group (`q1 q2 q3 q4`) and again as its own row, *the reporting nit — silence the `Migrated away` line*. So "The twelve mutations added in review" enumerates eleven distinct rows, and `q4` is filed under a group whose description does not match what it does. The headline **"twenty-two toward red" is correct** — I enumerated the union of both tables and got 22 distinct mutations — so only the inner tally is off.

**3. Carried from last round:** the body still cites `ANTIGRAVITY_HOME` at `:22`, `rules/` at `:129` and `settings.json` at `:85`. Those are pre-diff positions; at `02170ee` they are `:30`, `:154` and `:102`. The property is correct, the anchors are stale.

---

**Termination:** APPROVE, zero findings, CI green on the reviewed head `02170ee`.

*Method: `grep` here is `ugrep`, which returns 0 silently on `${…}` patterns — every figure above came from `/usr/bin/grep -F`, a Python parser, or a direct read, with exit statuses captured plainly rather than off a pipeline's end. Mutations were applied to isolated copies of `scripts/`; the working tree was never modified.*

Issue #771, opened 2026-08-09
✨ Block prohibited whole-tree git operations at the tool boundary


hcross:
Spec 0114 (#736) makes a whole-tree git operation in a shared ticket worktree
conditional on an exclusive claim held over a clean tree, and
`scripts/worktree-claim.sh` makes the compliant path cheap. Nothing in the
repository can *stop* the non-compliant one: an agent that types
`git reset --hard` into its own `Bash` call still runs it, and the residual
enforcement is a REVIEW-stage `class: tech` audit **after** the work is gone.
This ticket adds the one surface that can refuse the command before it executes.

Deferred deliberately by PLAN v2 for #736 — see *Alternatives considered and
rejected* in
[that comment](https://github.com/crewrig/crewrig/issues/736#issuecomment-5228485560),
which calls it "the **only** surface that could actually block the command" and
"the strongest option on the table".

## What to build

A Claude Code `PreToolUse` hook on `Bash` that matches the prohibited
operations — `git reset --hard`, `git checkout -- .`, `git stash`, `git clean`
in any form, and `git worktree remove --force` — when the resolved working
directory is under `.worktrees/`, and denies the call unless the acting agent
holds the claim over a clean tree.

The hook should **delegate, not reimplement**. `scripts/worktree-claim.sh`
already answers both halves of the question and its behaviour is pinned by
15 regression cases: `status --ticket <id>` reports `state: claimed` plus the
holder, and the clean-tree gate is evaluated on every `take` and every `run`,
including for the current holder. A hook that calls the script inherits the gate
and the ledger line; a hook that parses `git status` itself acquires a second
source of truth to keep in sync.

**Obligations it would enforce**, from `specs/0114-shared-worktree-agent-isolation.md`
on `main`:

- **R2** — "An agent SHALL NOT perform a whole-tree git operation in a shared
  worktree unless it holds the exclusive claim of requirement 5 on that worktree
  for the whole duration of the operation."
- **R3** — the prohibition "SHALL name `git reset --hard`, `git checkout -- .`,
  `git stash`, and `git clean` explicitly, and SHALL extend to every other
  operation that discards, reverts, or relocates uncommitted changes across a
  whole working tree rather than a named set of files."

R4's clean-tree condition rides along for free if the hook delegates, since the
script refuses over any non-empty `git status --porcelain --untracked-files=all`
regardless of who holds the claim.

## The four-CLI parity obligation — read this before scoping

[...]

PR #837, opened 2026-08-11
🐛 Run the wrapped command in the tree the claim gate certified

### LINKED ISSUE #779: 🐛 CREWRIG_REPO_DIR lets the claim gate certify one tree while the command mutates another
## The trap

`scripts/worktree-claim.sh` (shipping in #773, spec 0114) runs the wrapped command of `run` as a **child process**, so it inherits the **caller's** working directory. The clean-tree gate, meanwhile, evaluates `$TOPLEVEL`, derived from `CREWRIG_REPO_DIR` when that override is set.

Point the override at one tree while standing in another and the two diverge: **the gate certifies tree A while the command mutates tree B.**

## How it was found — by falling into it

The `developer` implementing #736 hit this while writing the fixture for the review's finding F1. They set `CREWRIG_REPO_DIR` to the fixture and left their shell inside the real `.worktrees/736`. The gate passed against the fixture, and `git clean -fdx` ran against the **real worktree**.

Nothing was lost, and the reason is worth stating precisely: that worktree happened to carry no untracked or ignored files at that moment. Luck, not a safeguard. They verified immediately — `git status` showed only the expected modifications, `HEAD` intact, a sibling agent's file still present, and the ledger had gone to the fixture rather than the real common dir, which is itself the signature of the divergence.

The header wording that misled them said the override redirects *"every git invocation"*, which reads as a whole-operation guarantee. It is not one: it redirects the script's own git invocations, not the wrapped command's cwd.

## Why it is not fixed in #773

The obvious fix — refuse when the toplevel is not an ancestor of the caller's cwd — **breaks the regression suite**, which drives its fixtures through exactly this divergence by design. That is a plan-level question about what `CREWRIG_REPO_DIR` is for, not something to smuggle into a REVIEW iteration on a different finding.

#773 therefore ships commit `e0fe55a`, **documentation only**: the header now states the boundary instead of implying a guarantee it does not provide.

## What this ticket owes

1. Decide what `CREWRIG_REPO_DIR` means: a test-harness affordance that may legitimately diverge from the cwd, or a repository override that must agree with it.
2. If the former, keep the divergence but make it observable — the gate should say which tree it certified, so a diagnostic that does not match the caller's expectation is visible rather than silent.
3. If the latter, add the ancestor check and rework the regression suite's fixtures to drive `cd` rather than the override.
4. Either way, cover it. The behaviour currently has documentation and no test.

## Related

Spec `0114-shared-worktree-agent-isolation.md` requirements 2 and 4 — the gate this weakens. The failure mode is adjacent to the one spec 0114 exists to remove: an agent believing a safety mechanism has cleared an operation that is in fact running somewhere the mechanism never looked.

hcross:
Realizes [spec 0126](https://github.com/crewrig/crewrig/blob/main/specs/0126-claim-gate-tree-convergence.md): `worktree-claim.sh run` now executes the wrapped command in the tree its clean-tree gate certified, so the certified tree and the mutated tree are one tree by construction rather than by a check. Three files, 134 insertions.

## How to read this PR?

**1. `scripts/worktree-claim.sh`, the change in `cmd_run`.** One statement —
`cd "$TOPLEVEL" || fail "…"` — between the post-claim TOCTOU re-check and `RC=0`. The comment above it is longer than the change, and deliberately so: it records why the two subshell forms are refused, because one of them is the natural thing to reach for and it is the dangerous one. Measured on `GNU bash 3.2.57(1)-release`, the repository's portability floor:

```text
$ /bin/bash -c 'set -euo pipefail; RC=0; ( cd /nonexistent; echo "REACHED cwd=$PWD" ) || RC=$?; echo "RC=$RC"'
/bin/bash: line 0: cd: /nonexistent: No such file or directory
REACHED cwd=/tmp
RC=0
```

Errexit is suppressed inside a subshell that is the left operand of `||`, so the wrapped command runs **in the caller's directory** and the invocation exits `0` — issue #779 re-created inside its own fix, wearing a success code. The `&&` variant at least fails loudly, but hands `RC` the status of `cd`, indistinguishable from a wrapped command exiting `1`. A change of directory in the function's own shell, guarded by the script's existing `fail`, avoids both and stays inside the exit-code vocabulary spec 0126 R4 freezes.

**2. The documentation of the accepted break — three surfaces, and this is where the review effort went.** The change moves a working directory, and *everything scoped to a working directory moves with it*. Two consequences beyond the relative-path one the spec anticipated:

- **`git clean -fdx` from a subdirectory now cleans the whole worktree.** Measured, with `.gitignore` = `*.log`: run from `sub/`, it destroyed `sub/deep/deep.log` and left `root.log` standing — cwd-scoped. After this change it takes everything. That widens the destruction scope of *the operation that caused the #736 incident this ticket closes*, over the file class the gate cannot see: the gate reads `git status`, which is blind to `.gitignore`-matched paths, as the script's own header and `--help` both state.
- **The documented safety procedure stopped matching it.** `docs/agent-team-protocol.md` told agents to enumerate with `git clean -ndx` first and treat the listing as a sibling's. That dry run is cwd-scoped too:

```text
$ git clean -ndx          (repo root)   → Would remove root.log
                                          Would remove sub/
$ cd sub && git clean -ndx              → Would remove ./deep      ← root.log absent
```

An agent in `sub/` following the procedure would have seen one subtree and deleted the worktree — the look that exists to surface the destruction, under-reporting it. Both surfaces now require the enumeration at the root and name the pathspec route (`git clean -fdx -- <subdir>/`) for a deliberate subtree clean.

The exposure set is **closed**, not open: of the whole-tree operations the mechanism governs, only `git clean` with `-x`/`-X` reaches state `git status` does not report. `git checkout -- .`, `git stash` and `git reset --hard` have their scope moved too and act on nothing over the empty status the gate requires.

**3. `scripts/tests/test-worktree-claim.sh`, cases 23 and 24.** Case 23 needs two fixtures, because one cannot exhibit a divergence. Both cases assert on an **observed mutation**, never on a diagnostic string — a message naming the certified tree can be perfectly correct while the file lands in the other one.

## How to test this PR?

```sh
bash scripts/tests/test-worktree-claim.sh     # expect: Results: 24 passed, 0 failed
bash scripts/check-bash32-portability.sh      # expect: OK
```

**Check the count, not just the exit status.** The harness requires a case to be registered in two places — `describe()` and the dispatch loop — and reports nothing when one is missed. `24 passed` is what proves both landed; `0 failed` would also be true of the 22-case suite.

**To confirm the two new cases actually discriminate**, run the new suite against the unmodified script. `git stash` is not the route: it is a whole-tree operation needing a claim over a clean tree, and the tree is dirty by construction because it carries the edit being stashed — the gate refuses with `5`. Use the copy recipe the suite documents:

```sh
git fetch crewrig main
d="$(mktemp -d)"; mkdir -p "$d/scripts/tests"
git show crewrig/main:scripts/worktree-claim.sh > "$d/scripts/worktree-claim.sh"
/bin/cp -f scripts/tests/test-worktree-claim.sh "$d/scripts/tests/"
bash "$d/scripts/tests/test-worktree-claim.sh"
```

Expected — and the criterion is not merely "red", since a case reds just as well from a broken fixture as from the absent behaviour:

```text
Results: 22 passed, 2 failed
FAIL  Case 23 …  expected path to exist: …/c23a/real/repo/.worktrees/736/marker.txt
FAIL  Case 24 …  wrapped command reported cwd '…/736/sub'; expected the worktree root '…/736'
```

Both failures land **at their mutation assertions**. Write the ref literally as shown: parameterized as `"$REF:scripts/…"` the `:s` is a zsh history modifier and silently corrupts the path.

**`scripts/worktree-claim.sh`** — four edits.

1. `cmd_run`: `cd "$TOPLEVEL" || fail "…"` inserted after the post-claim `tree_is_clean` re-check and before `RC=0`. A parent-shell change is safe here, verified rather than assumed: `git_here` has no call site after the toplevel resolution, `REPO_DIR` is captured as a string before any change, `tree_is_clean` uses `git -C "$TOPLEVEL"`, and everything the EXIT trap reaches (`ledger_append`, `claim_field`, `claim_exists`, `release_claim_dir`) resolves under `CLAIM_ROOT` / `CLAIM_DIR` / `LEDGER`, all derived from the absolutized `COMMON`. Release stays single: `run_release_on_exit` returns early unless `ACQUIRED` and clears it first — the same idempotence the existing `run_release_on_exit` + `refuse_dirty` pair relies on, proven by case 13 asserting `rc=5` with an intact ledger. `$TOPLEVEL` is absolute, so `CDPATH` cannot make this `cd` write to stdout.
2. The `Environment` block's `CREWRIG_REPO_DIR` entry — rewritten. It previously described the divergence as a standing property of the mechanism and told the reader to `cd` in and let the default apply; that hazard no longer exists. **This block is the designated carrier of the breaking-change notice** (spec 0126 R3, second clause), naming both consequences and the closed set.
3. The ignored-state block's `git clean -fdx` sentence — gains the extent. It was the only place in the file naming `run` and `git clean -fdx` in one sentence, so a reader landing there to learn *what* is destroyed now learns *where*. Nothing in that block became false; it was incomplete. The clause is placed in that sentence specifically and **not** in the paragraph below it, which argues that the limit is deliberate and that `--ignored` is not the fix — an argument the caller's working directory has nothing to do with.
4. `usage()`'s `run` entry — the working-directory clause, so the in-file surfaces agree. It does not repeat the breaking-change notice.

**`scripts/tests/test-worktree-claim.sh`** — `want_file` added as the mirror of the existing `want_no_file`; `case_23` and `case_24`; both registered in `describe()` and in the dispatch loop. `case_11`'s `run_claim_env … status` arm is deliberately **untouched** — that is spec 0126 R9's deliverable, discharged by inaction, and it is what keeps R5 (`CREWRIG_REPO_DIR` unchanged for subcommands that wrap no command) held by the suite rather than by intention.

`case_24` normalizes its expected value through `cd … && pwd -P`, mirroring `fx_claim_root`, the one helper that already does this. `$TOPLEVEL` is physical while `fx_wt` concatenates strings over a bare `mktemp -d`; on macOS those differ (`/var/folders/…` against `/private/var/folders/…`), so unnormalized this case would fail on the maintainer's machine and pass on `ubuntu-latest` — a green certifying the platform rather than the behaviour. It compares for **equality** rather than with `want_out`, because the root is a prefix of the subdirectory and a substring assertion would be satisfied by the exact value the case exists to reject.

**`docs/agent-team-protocol.md`** — two edits in *Worktree Isolation*: the invocation paragraph states where a wrapped command runs, and the enumerate-first sentence now requires the dry run at the worktree root, with the pathspec route for a deliberate subtree clean.

**Not touched, and why:** no `artifacts/` source, so no component rebuild and no `metadata.provenance.version` bump. No `docs/cli-matrix.md` row — the file carries zero occurrences of `worktree-claim` and the script matches none of the `{build,install,setup,import,manage}-*.sh` trigger prefixes. Searched by concept rather than by script name for surfaces stating the old contract: `AGENTS.md`'s summary bullet makes no scope claim and delegates to the protocol doc; `specs/0114` and `specs/0117` name the operations without asserting anything about the wrapped command's directory. Case 9's `git clean -fdxq` is not a `run` site — it executes directly — so this change does not reach it.

**Downstream:** `specs/0117-tool-boundary-command-guard.md` (issue #771, `approved`) delegates its decision to this script, which is why #779 sequences before it. A revert landing after 0117's implementation would return that guard to delegating to a gate that can read the wrong tree, so such a revert should re-open #779 rather than close it silently.

**Process record:** the PLAN went through five revisions and six cold-review passes before APPROVE. The reviews caught, among others, both cwd-scope consequences above, a test that would have passed in CI and failed on macOS, an acceptance criterion that a broken fixture would have satisfied, and four wrong figures — two mine, one the reviewer's own, one re-quoted after the correct value had already been published. The full trail is on #779.

Closes #779


hcross:
## Verdict: APPROVE

Cold review, no authoring context. Verified against `main`@`6efc4ce`, PR head `98a47af` (`headRefOid` confirmed unchanged at time of posting). All work below was done in an isolated `git worktree add --detach /tmp/crewrig-pr837-wt <headRef>` plus scratch copies under `/tmp` — nothing was executed against the shared checkout.

### CI status

All 13 checks pass, including `check-components` which was still `pending` (in_progress since 10:32:56Z) when I first queried and completed `success` on re-check. No required check is failing.

### Correctness — spec 0126 R1–R10

Walked every requirement against `scripts/worktree-claim.sh` at head:

- **R1/R2** — `cmd_run` does `cd "$TOPLEVEL" || fail …` unconditionally, and `tree_is_clean` gates on `git -C "$TOPLEVEL" status …` — same variable, no branch on whether `CREWRIG_REPO_DIR` is set. Confirmed `git_here` (which reads `$REPO_DIR`) has no call site after line 437 — nothing downstream can diverge from `$TOPLEVEL`.
- **R3** — accepted consequence documented in three in-file surfaces (Environment block, ignored-state block, `usage()`) plus `docs/agent-team-protocol.md`.
- **R4** — wrapped-command exit code still propagates via `RC=0; "$@" || RC=$?; … exit "$RC"`, unchanged. One documentation gap noted below (finding 3).
- **R5** — `take`/`release`/`takeover`/`status`/`history` never touch `cd`; case 11 (untouched) and case 12/13/14/16 all still exercise these paths correctly (see Test coverage).
- **R6** — Environment block rewritten to state R1–R3; verified it no longer describes the divergence as permanent.
- **R7** — case 23. Verified for real (see Test coverage): fails against pre-spec `main` at exactly the mutation assertion.
- **R8** — case 24. Same verification. Equality comparison (not substring) closes the prefix-match hole since the root is a prefix of the subdirectory — verified this is load-bearing.
- **R9** — case 11's `run_claim_env … status` arm is byte-identical to pre-PR; confirmed via diff.
- **R10** — `docs/agent-team-protocol.md` → *Worktree Isolation* now states the working directory (`run` executes at the worktree root …).

All ten satisfied.

### The `cd` in `cmd_run` — mechanics

- **Placement**: after the post-claim TOCTOU recheck, before `RC=0`. Correct — matches the comment's own description.
- **`fail` inside R4's contract**: `fail` calls `exit 1`, which re-fires the `EXIT` trap (`run_release_on_exit`) since the trap is armed at that point — release still happens, single (see below), and the diagnostic follows the documented `Error:`-prefix convention that separates `run`'s own-behalf failures from the wrapped command's output. Behaviorally sound. **Documentation gap** (finding 3, non-blocking): the header's "Exit contract — authoritative" block enumerates specific reasons for exit 1 and does not list this new one (TOPLEVEL becoming inaccessible between the recheck and the `cd`).
- **Release stays single**: `run_release_on_exit` guards on `ACQUIRED` and clears it as its first action, so a `cd`-failure exit (explicit `run_release_on_exit` call is not even reached here — the trap fires once) and the normal explicit-call-then-`trap -`-disarm path cannot double-release. Confirmed by reading, and case 10/14/16 continue to pass.
- **`git_here` / relative paths**: no call site after `$TOPLEVEL`/`$COMMON` resolution; `tree_is_clean`, `claim_field`, `claim_exists`, `ledger_append`, `release_claim_dir` all resolve off the absolutized `$COMMON` — confirmed by grep, nothing downstream of the `cd` resolves a path relative to the old cwd.
- **`CDPATH`**: `$TOPLEVEL` is always absolute (`pwd -P` at resolution). Empirically confirmed `cd` to an absolute path with `CDPATH` set produces no stdout pollution, with or without the target under a `CDPATH` entry.
- **`set -e` semantics**: `cd … || fail …` and `"$@" || RC=$?` are both the correct pattern (command as left operand of `||`) — no subshell involved, so the two hazards the code comments call out (RC-conflation, errexit suppression inside a subshell that is itself the left operand of `||`) don't apply to the shipped form.
- **Signals / stdin/stdout**: trap is armed (`EXIT INT TERM`) before the `cd`; `cd` is a builtin touching no file descriptors, so the wrapped command's stdin/stdout/stderr inheritance is unaffected.

### Adversarial check — mutation testing the `cd`, as requested

I mutated the shipped `cd "$TOPLEVEL" || fail …; RC=0; "$@" || RC=$?` into the two forms the code's own comment names as rejected, and reran the full 24-case suite against each, unmodified:

```
Mutant A: RC=0; ( cd "$TOPLEVEL" && "$@" ) || RC=$?
Mutant B: RC=0; ( cd "$TOPLEVEL"; "$@" ) || RC=$?
```

Both produce **`Results: 24 passed, 0 failed`** — identical to the real fix. Neither case 23 nor case 24 discriminates the shipped implementation from either explicitly-rejected alternative, **including Mutant B**, which the code's own comment says "is #779 itself, re-created inside its own fix, wearing a success code."

Root cause: both mutants are behaviorally identical to the shipped `cd` on the happy path — a subshell that successfully `cd`s and then execs `"$@"` puts the child in the same directory as a parent-shell `cd` would. The two forms only diverge from the shipped code when `cd "$TOPLEVEL"` itself **fails** (a TOCTOU race between the second `tree_is_clean` recheck and the `cd`, e.g. the worktree removed from under the running claim). Neither new case constructs that failure — both only exercise the successful-`cd` path.

**Finding 1 (`class: tech`, non-blocking).** The regression suite fully satisfies spec 0126 R7/R8 as literally written, and the shipped code itself is correct — this is not a defect in what merges today. But the suite provides zero protection against a future "simplification" back to either subshell form, despite the code's own commentary treating that regression as the primary risk this implementation choice defends against. I don't think a deterministic fixture for the exact TOCTOU race is cheap to build (making `git status` succeed against `$TOPLEVEL` and then `cd` to it fail requires injecting a removal in a window of a few interpreter instructions), which may be exactly why it wasn't attempted — but that also means the prose argument in the header is currently backed by reasoning alone, not by test evidence, and is worth recording as a residual gap rather than letting the "24 passed" figure imply otherwise.

### Test coverage — cases 9, 10, 12, 13, 14, 16 (premise check)

All six invoke at the fixture's worktree **root** (`fx_wt` returns `.worktrees/$TICKET`, never a subdirectory), so `cd "$TOPLEVEL"` is a no-op relative to their existing cwd — none of their premises are invalidated:

- **Case 9** doesn't go through `run` at all — the `git clean -fdxq` runs directly via `( cd "$(fx_wt c9)" && git clean -fdxq )` in the test harness, not through `cmd_run`. Unaffected by construction.
- **Cases 10, 12, 13, 14, 16** all pass `"$(fx_wt <name>)"` (the root) as the invocation cwd to `run_claim`/`run_claim_env`. Traced case 16's nested `run … -- bash interloper.sh` (which itself shells out to `takeover`) — the inner invocation's cwd is inherited from the outer `run`'s `cd`, which was already the worktree root both before and after this change, so no behavioral difference.

Re-ran the full suite (see below) — all still pass, and the rationale comments above each case remain true, not just the assertions.

### Verification performed (ran myself, not trusted from the PR body)

```
$ bash scripts/tests/test-worktree-claim.sh   # on PR head 98a47af
Results: 24 passed, 0 failed

$ bash scripts/check-bash32-portability.sh    # on PR head 98a47af
OK: no forbidden Bash 4+ construct … (160 files scanned, 148 array-scanned)

# Pre-change discrimination (PR body's own recipe: main's script + PR's tests)
Results: 22 passed, 2 failed
FAIL  Case 23 …  expected path to exist: …/c23a/real/repo/.worktrees/736/marker.txt
FAIL  Case 24 …  wrapped command reported cwd '…/736/sub'; expected the worktree root '…/736'
```

Both failures land exactly where the PR body says they do. Confirms cases 23/24 do discriminate the fix from the pre-fix behavior (as opposed to the narrower gap in Finding 1, which is about discriminating the fix from *plausible near-miss rewrites*, not from the pre-fix code).

### Documentation — the closed-exposure-set claim

Verified the headline claim (only `git clean -x/-X` reaches ignored state the gate can't see; `git checkout -- .`, `git stash`, `git reset --hard` "act on nothing over the empty status the gate requires") empirically:

```
$ git reset --hard, run from sub/  → reverts BOTH root.txt and sub/sub.txt (whole-repo)
$ git stash,        run from sub/  → stashes BOTH root.txt and sub/sub.txt (whole-repo)
$ git checkout -- ., run from sub/ → reverts ONLY sub/sub.txt (genuinely cwd-scoped)
```

**Finding 2 (`class: tech`, non-blocking).** `scripts/worktree-claim.sh`'s own Environment-block comment states `git checkout -- .`, `git stash` and `git reset --hard` "have their scope moved too" by this change. That's only true for `git checkout -- .` (its `.` pathspec is genuinely cwd-relative). `git stash` and `git reset --hard` take no path argument and have always acted on the **whole repository** regardless of the caller's cwd — nothing about their scope "moved," because they were never cwd-scoped in the first place, with or without this fix. The bottom-line conclusion (nothing to destroy, because the gate already requires an empty whole-repo `git status`) still holds — the *reason* just isn't "scope moved and then found empty" for those two commands, it's "never scoped, and separately guaranteed empty." Not repeated in `docs/agent-team-protocol.md` (which lists the same three commands but makes no scope-movement claim), so the inaccuracy is localized to this one comment block. Worth a follow-up wording fix given how much the PR foregrounds the precision of exactly this reasoning.

**Finding 3 (`class: tech`, non-blocking).** Restating from above: the "Exit contract — authoritative" header block wasn't updated to list the `cd`-failure path as a new route to exit 1. Cosmetic/completeness only — the code's behavior on that path is correct and consistent with the rest of the contract.

Swept the rest of the repository for surfaces still stating the pre-0126 contract (searched by concept — "caller's cwd", "inherits… working directory", "child process" — not just by grepping the script's own name):

- `specs/0114-shared-worktree-agent-isolation.md`, `specs/0117-tool-boundary-command-guard.md`, `specs/0118-spec-pr-merge-status-guard.md` — reference `worktree-claim.sh` but assert nothing about the wrapped command's directory. No contradiction.
- `AGENTS.md` line 261 (`Whole-tree git ops` bullet) — delegates to `docs/agent-team-protocol.md`, makes no scope claim itself.
- `docs/cli-matrix.md` — zero occurrences of `worktree-claim`; confirmed the script matches none of the `{build,install,setup,import,manage}-*.sh` trigger prefixes, so no row is owed.

Closed, as claimed, modulo Finding 2's wording.

### Convention compliance

- Gitmoji-only title (`🐛 …`), no conventional-commit prefix. Compliant.
- PR body carries the required sections (*How to read this PR?*, *How to test this PR?*, *Detailed description (for agents)*). Compliant.
- No `artifacts/` source touched → no `metadata.provenance.version` bump owed. Confirmed via file list.
- No `.claude/**`, `.gemini/**`, `artifacts/**`, `extensions/**`, build/install/setup script, or CLI-workflow path touched → no `docs/cli-matrix.md` update owed. Confirmed above.
- `Closes #779` is the correct keyword here (this is the implementation PR for its own logbook issue, not a spec-PR) — no close-keyword trap.

### Linter findings

`lint-shell.sh scripts/worktree-claim.sh scripts/tests/test-worktree-claim.sh` → exit 0, no output (shellcheck present, clean).
`lint-markdown.sh docs/agent-team-protocol.md` → exit 0, no output (markdownlint present, clean).

### Summary

Spec 0126 R1–R10 all satisfied and independently verified against the code, not just the PR body. The `cd` insertion in `cmd_run` is correctly placed, safe under `set -e`/`CDPATH`/signals, and releases the claim exactly once. Cases 23/24 genuinely discriminate the fix from the pre-0126 behavior (verified by actually running the suite against `main`'s script). CI is green. Three non-blocking `class: tech` findings recorded above — the mutation-testing gap (Finding 1) is the one worth a follow-up look, the other two are wording fixes.

hcross:
All three findings addressed in `390178a`. None was contested, and Finding 1 is the most useful thing said about this change.

**Finding 1 — the mutation test succeeded, and that is now recorded at the call site.**

Both rejected forms reporting `24 passed, 0 failed` is the finding I asked for and did not expect to get. The diagnosis is exactly right: all three forms agree whenever the `cd` succeeds, and cases 23/24 only exercise that path. They discriminate this fix from the behaviour that *preceded* it — which is what spec 0126 R7/R8 ask — not from a rewrite of itself.

I did not attempt a fixture for the divergent path, for the reason you give: it needs `git -C "$TOPLEVEL" status` to succeed at the post-claim re-check and the `cd` to that same path to fail a few interpreter instructions later, and anything blunt enough to break the second also breaks the first, at which point the case refuses with 5 and proves nothing.

What I did instead is refuse to let the count imply a protection that does not exist. The comment above the `cd` now opens with `NO TEST DEFENDS THIS CHOICE, AND THE SUITE'S GREEN MUST NOT BE READ AS ONE`, states that both mutants pass, and names what actually guards the line: the comment, not the suite. Follow-up is #839, which carries your measurement, the reason a fixture is expensive, and three candidate approaches — the cheapest being to add both mutants to the falsification matrix the suite's header already maintains, where a row that stops discriminating is itself treated as a defect.

**Finding 2 — you are right, and the error was mine in a comment this PR foregrounds.**

Re-measured before rewriting rather than taking the correction on trust. From a subdirectory:

```text
git reset --hard    → reverts root.txt AND sub/sub.txt   (whole repository)
git stash           → reverts root.txt AND sub/sub.txt   (whole repository)
git checkout -- .   → reverts sub/sub.txt only            (genuinely cwd-scoped)
```

So the closed set holds for two different reasons, and the comment now says so: `git checkout -- .` is the one other cwd-scoped operation and acts on nothing over the empty status the gate requires, while `git stash` and `git reset --hard` take no pathspec and were never the caller's directory to scope — their scope does not move because it was never there. Saying "their scope moved too" was uniform and wrong for two of the three.

**Finding 3 — added.** The authoritative exit contract now lists the `cd`-failure route to exit 1, with the note that the wrapped command contributes no output on that path, which is the disambiguation rule the header already documents for `run`'s own-behalf failures.

Suite still `24 passed, 0 failed` and `check-bash32-portability.sh` OK after the three edits; CI re-running on the new head.

Thanks for running the mutation — it is the only check in this whole ticket that told me something the reasoning had not.

PR #856, opened 2026-08-11
🐛 Name the modified files in the sync refusal, never their directory

### LINKED ISSUE #719: 🐛 sync-from-upstream.sh: Directory-level dirty checks lead to silent over-erasure of new upstream files during restoration
### Problem Description

In `scripts/sync-from-upstream.sh`, directory-level paths (like `scripts`) that do not have an explicit policy in `.crewrig/core-paths.txt` default to the `strict` policy.

When a downstream project customizes a single file inside a `strict` directory (for example, modifying `scripts/sync-from-upstream.sh` itself for macOS/Bash 3.2 compatibility), the sync script's dirty check flags the **entire directory** as dirty:

```text
Error: the following core-layer paths have local modifications:
  scripts
```

To run the sync, the downstream project is forced to temporarily align the entire directory with upstream (`git checkout upstream/main -- scripts`), run the sync (creating the graft/merge commit), and then restore their custom modifications by checking out the directory from their previous commit:

```bash
git checkout develop -- scripts
```

**The Regression:**
Because the downstream project checks out the entire directory from their previous local commit, they silently **overwrite and revert** all other files in that directory that were newly added or modified by upstream (for example, losing the `mempalace-http-wrapper.py` version guard from Spec 0108 and the `setup-*-interactive.sh` catalogue picker updates from Spec 0096). This leads to silent regressions and test failures in CI.

### Proposed Solutions

To prevent this silent over-erasure during downstream synchronizations, we propose the following improvements:

1. **Fine-Grained Dirty Checks:**
   Modify `scripts/sync-from-upstream.sh` to evaluate and report dirty status **per-file** rather than per-directory, even for directory entries. Only the specific customized files should be reported as dirty, allowing downstream projects to temporarily revert and restore only those specific files rather than the entire directory.

2. **Fine-Grained Restoration Guidance:**
   Update the error message and documentation of `scripts/sync-from-upstream.sh` to explicitly warn against directory-level checkouts during restoration, advising instead to only restore the specific files that carry local modifications:

   ```bash
   # Avoid: git checkout develop -- scripts
   # Recommend: git checkout develop -- scripts/sync-from-upstream.sh
   ```

hcross:
Realizes [spec 0129](https://github.com/crewrig/crewrig/blob/main/specs/0129-dirty-report-names-files.md): when the dirty guard refuses a sync, it names the modified **files** rather than the manifest entry containing them, so a downstream adopter restores what they modified instead of the directory holding it. Three files, +250/−18.

Here is the whole ticket in one screen. Before, with one file customized inside the `tools` directory entry:

```text
Error: the following core-layer paths have local modifications:
  tools
```

After, with two:

```text
Error: the following core-layer paths have local modifications:
  tools/alpha.sh
  tools/nested/gamma.sh
Revert these changes before running sync, or promote them to overlay overrides.

Restore ONLY the files listed above, one path at a time:
  git checkout <your-ref> -- <path listed above>
Never restore the containing directory. A directory-level checkout also
reverts every file upstream added or changed in it, silently.
```

## How to read this PR?

**1. `scripts/sync-from-upstream.sh`, the detection loop.** The change is small and the comment above it is not, because two of the three decisions it records were wrong in earlier drafts and were only caught by measuring.

The detection was *already* per-file — `strict_blob_is_dirty` runs on each member. The knowledge was discarded by a `break` on the first hit and an append of the entry rather than the member. Both are gone.

**2. The deduplication, and why it sits after the loop rather than at append time.** A file can be governed by two strict entries at once, and the manifest ships two such pairs. They fail *differently*:

| pair | duplicate produced by | append-time dedup in the tree branch | post-loop pass |
|---|---|---|---|
| `docs` + `docs/index.json` | one append from **each** branch | still duplicated | fixed |
| `artifacts/core` + `artifacts/core/system-context` | two appends from the **tree** branch | fixed | fixed |

Spec R5 keeps the blob branch out of scope, so an append-time filter can only live in the tree branch — where it cannot see the first pair. Only the post-loop pass closes both. `case-mm` is pinned to the nested-**file** shape for that reason: the nested-directory shape goes green against the wrong placement and would not discriminate.

**3. The newline delimiter.** The seen-string membership test is `case "$NL$seen$NL" in *"$NL$p$NL"*`. With a space delimiter and `docs/a b.md` stored, the probes `docs/a` and `b.md` both report as already-seen — and a false hit there **drops a genuinely modified file from the refusal**, which is this ticket's own defect reached through its fix. `git ls-tree` prints a space in a path verbatim but C-quotes control characters, so a member can contain a space and can never contain a newline.

**4. `docs/adoption-guide.md`.** Not adjacent to the work — part of it. The guide reproduced the refusal verbatim and its troubleshooting entry taught the recovery this change alters, so a fix landing without it would ship a guide describing a message the script no longer prints, next to guidance for the move the script now warns against. The example's `config/SOUL.md` also goes: `.crewrig/core-paths.txt` has no `config` entry and no `config/SOUL.md` entry, so the guard could never emit it. It is replaced by `scripts/sync-from-upstream.sh` — a real member of a strict directory entry, and the file the reporting adopter had actually customized.

**5. The three cases, and why every assertion is whole-line.** The string that must be absent (`tools`) is a substring of the string that must be present (`tools/alpha.sh`). So a substring **absence** check fails whether or not the fix is in — a test that can never go green — and a substring **presence** check on the directory name passes with the defect fully present, because the pre-fix refusal printed exactly `  tools`. `run_case_dirty_report` compares whole lines with `grep -cxF` and requires a count of exactly 1, which is also what pins the deduplication.

## How to test this PR?

```sh
bash scripts/tests/test-sync-from-upstream.sh    # expect: Results: 68/68 passed
bash scripts/check-bash32-portability.sh         # expect: OK
```

**To confirm the three new cases actually discriminate**, run the new suite against the unmodified script. Stay inside the repository while doing it — `case-bb` resolves its subject with `git -C "$SCRIPT_DIR" rev-parse --show-toplevel`, which from a `mktemp` copy falls through to the process cwd and reports a spurious failure if you have wandered out:

```sh
git fetch crewrig main
d="$(mktemp -d)"; mkdir -p "$d/scripts/tests"
git show crewrig/main:scripts/sync-from-upstream.sh > "$d/scripts/sync-from-upstream.sh"
/bin/cp -f scripts/tests/test-sync-from-upstream.sh "$d/scripts/tests/"
bash "$d/scripts/tests/test-sync-from-upstream.sh"
```

Expected `Results: 65/68 passed` — every pre-existing case still green, and the three new ones failing **at their own mutation assertions**:

```text
FAIL  case-kk … (expected the line '  tools/alpha.sh' exactly once, saw 0)
FAIL  case-kk … (stderr names 'tools' as its own line, which it must not)
FAIL  case-ll … (expected the line '  tools/beta.sh' exactly once, saw 0)
FAIL  case-mm … (stderr names 'papers' as its own line, which it must not)
```

That distinction is the acceptance criterion: a case that reds because its fixture never built would satisfy a bare "it failed". Write the ref literally as shown — parameterized, `:s` is a zsh history modifier that corrupts the path — and note `cp` is aliased interactive in this environment, hence `/bin/cp -f`.

**`scripts/sync-from-upstream.sh`** — the tree branch of the dirty-core detection loop appends each modified `$member` and nothing for the entry; the `break` is removed (R1/R2/R3). The `excluded_children_of` skip is untouched (R6) — **`case-z` is its sentinel**, verified by deleting the skip and observing which case reds; `case-aa` cannot fail on that mutation because its `settings.json` is byte-identical to upstream, and `case-a`/`case-e` cannot either because their excluded children are adopter-only and never appear in `git ls-tree FETCH_HEAD`. The blob branch is untouched (R5). A deduplication pass follows the loop, order-preserving, first occurrence wins — the duplicates are identical strings, so what that preserves is manifest order, which is the order the refusal already printed. The write-back is `DIRTY=(${UNIQ[@]+"${UNIQ[@]}"})`, the guarded form spec 0124 R2 requires. The refusal gains the targeted-restoration guidance (R4); its exit status is unchanged (R7).

**R8 is satisfied by construction and required nothing:** the removed `break` was unreachable when nothing is modified, so a clean run already visited every member and the non-refusing path pays nothing new.

**`scripts/tests/test-sync-from-upstream.sh`** — `run_case_dirty_report` added, taking present-paths and absent-lines around a `--` separator, asserting exit 1, `grep -cxF "  $p"` equal to 1 for each present path, and no whole-line match for each absent one. Cases `kk` (R9, one modified member), `ll` (R10, three modified members including a nested one), and `mm` (the deduplication, pinned to the nested-file shape). Case `ll` is what discriminates R3: on `kk`'s single-member fixture an implementation that keeps the `break` passes by accident.

**`docs/adoption-guide.md`** — three edits: the fenced example block, the `git checkout` resolution step, and the cause/effect pair above it, which described a listed path as a manifest entry. That vocabulary is what spec R2 corrects.

**Not touched, and why:** no `artifacts/` source, so no component rebuild and no `metadata.provenance.version` bump. No `docs/cli-matrix.md` row — the file carries no occurrence of `sync-from-upstream` and the script matches none of the `{build,install,setup,import,manage}-*.sh` trigger prefixes. `.crewrig/core-paths.txt` is read, not modified; its format is out of scope per spec 0129, so no manifest co-maintenance obligation arises. The consumer set for the refusal text is exactly two files — `git grep -ln "have local modifications"` returns the script and the guide — and both are in this diff.

**Process record:** the PLAN went through three revisions and three cold-review passes. The reviews caught two defects that would have shipped — the deduplication placed where it cannot see half the duplicates, and a membership test that would have silently dropped modified files — plus a blast-radius claim of mine asserting no in-repo consumer of the refusal text existed, which a three-second `grep` refutes, and one finding I recorded as resolved without making the edit. None was reachable by reading; each came from applying the change in a scratch copy and measuring. The trail is on #719.

Closes #719

### DIFF
--- a/docs/adoption-guide.md
+++ b/docs/adoption-guide.md
@@ -319,16 +319,26 @@ listed in `.crewrig/core-paths.txt` from the URL set in
 `config/TOOLS.md`, `crewrig.config.toml`, and `artifacts/community/`)
 untouched.
 
-**Most-likely error — dirty-core guard:** If at least one core-layer path
-has been locally modified, the script will list the offending paths and
-exit 1 with a message similar to:
+**Most-likely error — dirty-core guard:** If at least one core-layer file
+has been locally modified, the script will list the offending **files** — never
+the directory containing them — and exit 1 with a message similar to:
 
 ```text
 Error: the following core-layer paths have local modifications:
-  config/SOUL.md
+  scripts/sync-from-upstream.sh
   artifacts/core/skills/developer/SKILL.md
 Revert these changes before running sync, or promote them to overlay overrides.
+
+Restore ONLY the files listed above, one path at a time:
+  git checkout <your-ref> -- <path listed above>
+Never restore the containing directory. A directory-level checkout also
+reverts every file upstream added or changed in it, silently.
 ```
+
+Both example paths are *members* of directory entries in
+`.crewrig/core-paths.txt` (`scripts` and `artifacts/core`). The guard names the
+member rather than the entry precisely so that the restoration below can be
+targeted: told `scripts`, an adopter restores `scripts`.
 
 Resolution: see [Troubleshooting — dirty-core refusal](#dirty-core-refusal) below.
 
@@ -468,22 +478,30 @@ missing directories, re-run `bash scripts/build-components.sh`.
 
 ### Dirty-core refusal during sync {#dirty-core-refusal}
 
-**Cause:** At least one path listed in `.crewrig/core-paths.txt` has been
-locally modified. The sync script enforces a dirty-core guard to prevent
-upstream changes from silently overwriting local modifications to core-layer
-files.
+**Cause:** At least one **file** governed by `.crewrig/core-paths.txt` has been
+locally modified — either because it is listed there itself, or because it sits
+inside a directory that is. The sync script enforces a dirty-core guard to
+prevent upstream changes from silently overwriting local modifications to
+core-layer files.
 
 **Effect:** `bash scripts/sync-from-upstream.sh` exits 1 and lists the
-offending paths.
+offending **files**, each at its full path. A directory listed in the manifest
+is never itself reported: what you see is what you modified.
 
 **Resolution:** Choose one of two paths for each offending file:
 
 1. **Revert the modification** — if the change was experimental or
-   unintended, restore the file to its committed state:
+   unintended, restore that file, and only that file, to its committed state:
 
    ```bash
    git checkout -- <path/to/core-file>
    ```
+
+   **Restore file by file, never the containing directory.** A directory-level
+   checkout — `git checkout -- scripts` — also reverts every file upstream
+   added or changed inside it, with no message and no error. That is how an
+   adopter loses a new upstream guard and discovers it later through unrelated
+   test failures; it is why the guard lists files rather than directories.
 
    Then re-run `bash scripts/sync-from-upstream.sh`.
 
--- a/scripts/sync-from-upstream.sh
+++ b/scripts/sync-from-upstream.sh
@@ -432,8 +432,16 @@ for i in "${!PATHS[@]}"; do
   resolves_at_fetch_head "$path" || continue
 
   if [ "$(git cat-file -t "FETCH_HEAD:$path" 2>/dev/null)" = "tree" ]; then
-    # Directory entry: check each upstream member individually.
-    dir_dirty=0
+    # Directory entry: report the MEMBERS, never the entry (spec 0129 R1/R2).
+    # An adopter restores what they are shown. Reporting `scripts` is what makes
+    # them run `git checkout <ref> -- scripts` afterwards, reverting every file
+    # upstream added in between — the silent over-erasure of issue #719.
+    #
+    # No `break` (spec 0129 R3): a partial list sends the adopter back for a
+    # second refusal, and the second list is the one they act on in a hurry.
+    # This costs nothing on the path that runs every time — when nothing is
+    # modified the loop already visits every member, because the early exit was
+    # only ever reachable once something was dirty (spec 0129 R8).
     while IFS= read -r member; do
       [ -n "$member" ] || continue
       skip=0
@@ -441,17 +449,45 @@ for i in "${!PATHS[@]}"; do
         case "$member" in "$excl"/*|"$excl") skip=1; break ;; esac
       done < <(excluded_children_of "$path")
       [ "$skip" -eq 1 ] && continue
-      if strict_blob_is_dirty "$member"; then
-        dir_dirty=1
-        break
-      fi
+      strict_blob_is_dirty "$member" && DIRTY+=("$member")
     done < <(git ls-tree -r --name-only FETCH_HEAD -- "$path/" 2>/dev/null)
-    [ "$dir_dirty" -eq 1 ] && DIRTY+=("$path")
   else
-    # Blob entry: upstream-history membership check.
+    # Blob entry: upstream-history membership check. Untouched by spec 0129 —
+    # for a file entry the entry IS the file, so this already reports precisely
+    # (R5).
     strict_blob_is_dirty "$path" && DIRTY+=("$path")
   fi
 done
+
+# One file can be governed by two strict entries — the manifest ships
+# `docs` + `docs/index.json` and `artifacts/core` + `artifacts/core/system-context`
+# — and would then be appended twice, putting a duplicated line in the list the
+# adopter is asked to copy. Dedup HERE, in one pass after the loop, and not at
+# append time: the `docs` pair is produced by one append from the tree branch and
+# one from the blob branch, so an append-time filter inside either branch cannot
+# see it. Order-preserving, first occurrence wins — the duplicates are identical
+# strings, so what this preserves is manifest order, which is the order the
+# refusal already prints.
+#
+# Delimited by NEWLINE, not by a space: a space-delimited membership test reports
+# a false hit for any probe that is a space-separated fragment of a stored value
+# (with `docs/a b.md` stored, both `docs/a` and `b.md` read as already-seen), and
+# a false hit here silently DROPS a modified file from the refusal — this
+# ticket's own defect, reached through its fix. `git ls-tree` prints a space in a
+# path verbatim but C-quotes control characters, so a member string can contain a
+# space and can never contain a newline.
+UNIQ=()
+seen=""
+NL='
+'
+for p in ${DIRTY[@]+"${DIRTY[@]}"}; do
+  case "$NL$seen$NL" in
+    *"$NL$p$NL"*) continue ;;
+  esac
+  seen="$seen$NL$p"
+  UNIQ+=("$p")
+done
+DIRTY=(${UNIQ[@]+"${UNIQ[@]}"})
 
 if [ ${#DIRTY[@]} -gt 0 ]; then
   echo "Error: the following core-layer paths have local modifications:" >&2
@@ -459,6 +495,11 @@ if [ ${#DIRTY[@]} -gt 0 ]; then
     echo "  $p" >&2
   done
   echo "Revert these changes before running sync, or promote them to overlay overrides." >&2
+  echo "" >&2
+  echo "Restore ONLY the files listed above, one path at a time:" >&2
+  echo "  git checkout <your-ref> -- <path listed above>" >&2
+  echo "Never restore the containing directory. A directory-level checkout also" >&2
+  echo "reverts every file upstream added or changed in it, silently." >&2
   exit 1
 fi
 
--- a/scripts/sync-from-upstream.sh
+++ b/scripts/sync-from-upstream.sh
@@ -459,10 +459,13 @@ for i in "${!PATHS[@]}"; do
   fi
 done
 
-# One file can be governed by two strict entries — the manifest ships
-# `docs` + `docs/index.json` and `artifacts/core` + `artifacts/core/system-context`
-# — and would then be appended twice, putting a duplicated line in the list the
-# adopter is asked to copy. Dedup HERE, in one pass after the loop, and not at
+# One file can be governed by two strict entries and would then be appended
+# twice, putting a duplicated line in the list the adopter is asked to copy. The
+# manifest ships FOUR such pairs today — `docs` + `docs/index.json`,
+# `artifacts/core` + `artifacts/core/system-context`, and `scripts` with each of
+# `scripts/sync-from-upstream.sh` and `scripts/build-docs-index.sh` — but the
+# dedup is placement-general and does not depend on that count. Dedup HERE, in
+# one pass after the loop, and not at
 # append time: the `docs` pair is produced by one append from the tree branch and
 # one from the blob branch, so an append-time filter inside either branch cannot
 # see it. Order-preserving, first occurrence wins — the duplicates are identical
--- a/scripts/tests/test-sync-from-upstream.sh
+++ b/scripts/tests/test-sync-from-upstream.sh
@@ -258,6 +258,61 @@ run_case_stderr() {
   fi
 }
 
+# run_case_dirty_report <name> <repo> <present-path>… -- <absent-line>…
+#
+# Spec 0129's report assertions, and the reason this helper exists rather than a
+# second `run_case_stderr` call: R2 is an ABSENCE requirement, and the string
+# that must be absent (`scripts`) is a substring of the string that must be
+# present (`scripts/sync-from-upstream.sh`). A substring absence check therefore
+# fails whether or not the defect is fixed — a test that can never go green —
+# while a substring PRESENCE check on the directory name passes with the defect
+# fully present, because the pre-0129 refusal prints `  scripts`. Only a
+# WHOLE-LINE comparison discriminates, and the refusal prints each path as two
+# spaces followed by the path.
+#
+# Present paths are matched with the same two-space prefix, for symmetry and
+# because a bare substring would match a sibling path that contains this one.
+run_case_dirty_report() {
+  local name="$1" repo="$2"
+  shift 2
+  local present=() absent=() bucket="present"
+  local a
+  for a in "$@"; do
+    if [ "$a" = "--" ]; then bucket="absent"; continue; fi
+    if [ "$bucket" = "present" ]; then present+=("$a"); else absent+=("$a"); fi
+  done
+
+  local actual_exit=0 stderr_out
+  stderr_out="$(cd "$repo" && CREWRIG_REPO_DIR="$repo" bash "$SCRIPT_UNDER_TEST" 2>&1 >/dev/null)" || actual_exit=$?
+
+  local ok=1 p n
+  if [ "$actual_exit" -ne 1 ]; then
+    echo "FAIL  $name (expected exit 1, got $actual_exit)"
+    ok=0
+  fi
+  for p in ${present[@]+"${present[@]}"}; do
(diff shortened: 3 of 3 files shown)


hcross:
## Verdict: REQUEST CHANGES

Cold-start review (no authoring context). All checks below were run against
content pinned to the PR head commit (`5065e1d286e0ad644a16686e1e78a5b335773712`),
extracted via `git show <sha>:<path>` into an isolated `/tmp` tree — not the
shared `.worktrees/719` checkout, which another concurrent session was actively
switching branches on mid-review (confirmed via `git reflog`: `main` →
`chore/709-spec-status-implemented` → `main`, commits and pulls, while this
review was in progress). No destructive git operation was run against that
shared tree; its live state was simply not trusted for content extraction once
the interference was noticed.

### CI status

10 of 13 checks passed: `build`, `grep-anti-patterns`, `check-components`,
`lint-markdown`, `lint-specs`, `test-harness-curate`, `check-skill-versions`,
`check-extension-version-bump`, `check-agents-size`, `check-ci-parity`. Three
are still in progress at review time: `check-spec-id-reserved`,
`check-feedback-routing`, `gitlab-ci-check`. `main` carries no configured
branch-protection required-status-check list (`gh api
repos/crewrig/crewrig/branches/main/protection` → 404 "Branch not protected"),
so none of these formally gate the merge, and none of the three pending checks
look related to this diff's file set (no spec added, no `ci/`/`.gitlab-ci.yml`
touch). Not a blocker for this verdict, but worth confirming green before
merge.

### Correctness

The detection-loop rewrite (`scripts/sync-from-upstream.sh`, tree branch,
~432-460) correctly realizes spec 0129 R1/R2/R3/R5/R6/R7/R8:

- R1/R2 — the tree branch now appends the per-member path (`DIRTY+=("$member")`)
  and never appends `$path` itself; the old `dir_dirty`/`break`/`DIRTY+=("$path")`
  path is gone. Verified directly (manual fixtures) and via `case-kk`/`case-ll`.
- R3 — no `break`; confirmed live by reintroducing one (see Mutation testing
  below): `case-ll` catches it, `case-kk` does not (matches the PR body's own
  claim about why a single-member fixture can't discriminate R3).
- R5 — blob branch untouched; confirmed by diff inspection.
- R6 — `excluded_children_of` skip untouched; confirmed by deleting it (see
  below): `case-z` catches it, nothing else does.
- R7/R8 — exit status and non-refusing-path cost unaffected; confirmed by
  reasoning (the clean-path loop body is unchanged per member) and by every
  clean-sync fixture still exiting 0.

The post-loop deduplication is correctly placed (after the loop, not at
append-time inside the tree branch) and correctly newline-delimited. Both
placement decisions are real, not decorative — confirmed by mutation (below).

`docs/adoption-guide.md`'s quoted refusal text (lines 327-336) is byte-for-byte
identical to the script's actual `echo` statements (`scripts/sync-from-upstream.sh:493-502`).
Both example paths in the guide (`scripts/sync-from-upstream.sh`,
`artifacts/core/skills/developer/SKILL.md`) are real files, genuinely members of
`strict` directory entries (`scripts`, `artifacts/core`) in
`.crewrig/core-paths.txt`, and genuinely emittable by the guard. `git grep -ln
"have local modifications"` against `main` returns exactly the script and the
guide — no other prose in the repository describes the old
directory-naming behavior.

**Finding (non-blocking) — code comment undercounts the manifest's actual
duplicate-producing shape.**
`scripts/sync-from-upstream.sh:462-464` states "the manifest ships two such
pairs" (`docs`+`docs/index.json`, `artifacts/core`+`artifacts/core/system-context`)
as the justification for why the dedup must be a post-loop pass. `.crewrig/core-paths.txt`
actually contains at least two more instances of the identical shape: `scripts`
(implicit `strict`) + `scripts/sync-from-upstream.sh` (explicit `strict`,
manifest line ~76), and `scripts` + `scripts/build-docs-index.sh` (explicit
`strict`, manifest line ~85). Confirmed empirically: modifying
`scripts/sync-from-upstream.sh` under the real manifest produces one append
from each branch, correctly deduplicated to a single line by the shipped
code — so the *fix* is unaffected (the dedup is placement-general, not
count-dependent) — but the comment's own enumeration is factually incomplete,
which matters because this PR's stated methodology is exactly "verified by
measuring, not by reading."
`class: tech`

### Convention compliance

Gitmoji title (`🐛 …`), PR body follows the project's template sections,
`docs/cli-matrix.md` correctly not touched (no `sync-from-upstream` occurrence
there, script matches none of the `{build,install,setup,import,manage}-*.sh`
trigger prefixes — confirmed via `git grep`), no `artifacts/` source touched so
no component rebuild/version-bump obligation. `scripts/check-bash32-portability.sh`
passes clean, and I independently re-ran the full suite under genuine
`/bin/bash` 3.2 (macOS-shipped, `3.2.57(1)-release`) rather than trusting a
`bash` on `$PATH` that on this machine actually resolves to Homebrew bash
5.3.15 — the test harness's own internal invocations (`bash
"$SCRIPT_UNDER_TEST"`) use unqualified `bash`, so a naive "it passed" only
proves 5.3 unless `$PATH` is forced. Forced (`PATH` prepended with a symlink to
`/bin/bash`), the full 68/68 result, the 65/68 pre-fix discrimination, and every
mutation result below reproduce identically under real 3.2.

### Test coverage — mutation testing (the adversarial core of this review)

Ran the shipped 68-case suite against the pre-fix script (`main` at `94c1020`)
and against four hand-authored mutations of the shipped fix, all from content
pinned by commit SHA (immune to the concurrent-checkout interference on the
shared worktree), under both Homebrew bash 5.3 and real `/bin/bash` 3.2:

| Mutation | Result | Discriminated by |
|---|---|---|
| Pre-fix script (no spec-0129 changes at all) | **65/68** | `case-kk`, `case-ll`, `case-mm` fail at their own present/absent assertions (not fixture-build failures) — matches the PR body's claimed discrimination recipe, reproduced independently. |
| Dedup moved to append-time inside the tree branch | **67/68** | `case-mm` (expects `papers/index.json` once, sees 2) — confirms the placement decision is load-bearing, not stylistic. |
| `break` restored after the first dirty member append | **67/68** | `case-ll` (misses `tools/beta.sh` and `tools/nested/gamma.sh`); `case-kk` does **not** fail, exactly as the PR body predicts for its single-member fixture. |
| `excluded_children_of` skip deleted | **67/68** | `case-z` only (`expected exit 0, got 1`) — matches the PR body's claim that `case-z` is R6's sentinel. |
| Newline delimiter (`NL='\n'`) reverted to a space (`NL=' '`) | **68/68 — NOT CAUGHT** | Nothing. |
| All four R4 restoration-guidance lines deleted (`Restore ONLY…` / `git checkout <your-ref>…` / `Never restore…`) | **68/68 — NOT CAUGHT** | Nothing. |

Three of the four mutations the team asked me to try are caught, exactly as
the PR body claims for each. The other two checks I added on top are **not**
caught, and both are concrete:

**Finding 1 (blocking) — the newline-vs-space delimiter defense has zero
regression coverage.** `scripts/sync-from-upstream.sh:481` (`NL='<newline>'`);
gap is in `scripts/tests/test-sync-from-upstream.sh` (no case anywhere
constructs a dirty path containing a space). This is not a hypothetical: I
built a concrete failure scenario. Manifest with two `strict` entries in this
order — a directory `container` (containing dirty member `container/name b`,
a real space in the filename) and, separately, a top-level file `b` (also
dirty). Under the shipped (newline-delimited) code both are reported. Under
the space-delimited mutation, `b` is **silently dropped** from the refusal
entirely — the adopter is told only about `container/name b` and never learns
`b` needs restoring, which is a direct R1 violation waiting to happen. This is
*exactly* the defect the PR's own comment (`scripts/sync-from-upstream.sh:471-478`)
names by description ("a false hit here silently DROPS a modified file from
the refusal — this ticket's own defect, reached through its fix") — the code
comment identifies the risk precisely, but nothing in the 68-case suite
protects against a future regression reintroducing it. Recommend a case
shaped like `case-mm` but pinning a directory member with a space against a
separate same-fragment top-level entry.
`class: tech`

**Finding 2 (blocking) — spec 0129 R4 (the restoration guidance text) has zero
regression coverage.** `scripts/sync-from-upstream.sh:497-502` (the four
`echo` lines realizing R4); gap is in `scripts/tests/test-sync-from-upstream.sh`
(`grep -n "Restore ONLY\|Never restore\|git checkout <your-ref>"` returns
nothing). Deleting the entire R4 block and going straight from "Revert these
changes…" to `exit 1` leaves the full suite green. R4 is a numbered,
independently-testable spec requirement — realized correctly by the code, but
with no assertion anywhere that the guidance text is actually emitted.
Recommend a `run_case_stderr` (or `-dxF` whole-line) assertion on at least the
`Restore ONLY` and `Never restore` lines, in the same fixture already used for
`case-kk` or a dedicated case.
`class: tech`

Existing cases whose premise touches this diff were individually re-checked
and remain valid: `dirty-core refusal` and `case-d` (single-file entries, R5,
substring-pattern assertions unaffected by the new trailing guidance lines
since `run_case_stderr` uses `grep -q`, not an anchored/whole match), `case-z`
(R6 sentinel, confirmed above), `case-aa` (still passes, though its
`.github/copilot` substring assertion now matches by virtue of being a prefix
of the reported member path rather than the reported entry itself — not a
regression, just a coincidental rather than a targeted match), `case-bb` (gitignore
pattern check, unrelated to this diff's logic, re-confirmed passing — note for
future reviewers: it resolves its subject via `git -C "$SCRIPT_DIR" rev-parse
--show-toplevel`, which falls through to process `cwd` outside a real repo
checkout and gives a spurious failure if you `cd` into a bare `mktemp`
directory before invoking the suite — this cost me one contaminated run
before I read the PR body's own warning on it), `case-m` (shallow-clone guard,
unrelated to the tree-branch change, re-confirmed passing).

### Linter findings

`shellcheck` (0.11.0, present and actually invoked — not degraded) against
both `scripts/sync-from-upstream.sh` and `scripts/tests/test-sync-from-upstream.sh`:
clean, no findings. `markdownlint` (present, actually invoked) against
`docs/adoption-guide.md`: clean, no findings.

### Summary

Solid, well-reasoned fix — every requirement the team asked me to probe
(placement, break, exclusion skip) is genuinely protected by the shipped
tests, and the documentation is byte-accurate. The two gaps above are why this
is Request Changes rather than Approve: both are mutations that leave the
*entire* 68-case suite green while regressing user-visible behavior the PR's
own prose identifies as important (one is the ticket's self-named defect
class; the other is a numbered spec requirement). Both are small, additive
fixes — one or two more `run_case_dirty_report`/`run_case_stderr` cases, no
production-code change needed.

hcross:
All three findings addressed in `178f148`. None contested, and the two blocking ones are the same finding twice: a defence stated in prose with no signal behind it — which is the shape of the defect this whole ticket is about.

**Findings 1 and 2 — both now discriminate, verified by re-running your mutations myself rather than assuming the new cases work.**

```text
space delimiter reverted   →  70/71   caught by: case-nn
guidance lines deleted     →  69/71   caught by: case-oo
```

`case-nn` builds the false hit deliberately, on your repro: the `container` entry is processed first and stores the member `container/name b`; the separate `b` entry then probes `b`, a space-separated fragment of what is stored. Under a space delimiter that probe reads as already-seen and `b` never reaches the report — a modified file, silently missing, which is exactly the failure class spec 0129 R3 forbids.

`case-oo` asserts the sentence warning against the directory-level checkout and the one stating the targeted restoration, rather than the whole block. The wording may well improve; a refusal that stops warning has lost R4.

Worth stating plainly, because it is the second time this session: **the mutation exercise is the only check that found either gap.** Both requirements were realized correctly in the shipped code, the suite was green, and reading the diff would not have revealed that deleting the defence changed nothing. On #779 the same exercise showed the new cases did not discriminate the shipped construct from the two forms its own comment rejected. Twice now, the difference between "the tests pass" and "the tests would fail if the behaviour were removed" has been invisible to everything except mutating the code and re-running.

**Finding 3 — you were right, and the count was worse than "at least four".**

I enumerated every nested manifest pair and classified both ends by policy rather than taking either figure on trust:

```text
artifacts/core/system-context   strict   under artifacts/core    strict    <== duplicate-producing
docs/index.json                 strict   under docs              strict    <== duplicate-producing
scripts/build-docs-index.sh     strict   under scripts           strict    <== duplicate-producing
scripts/sync-from-upstream.sh   strict   under scripts           strict    <== duplicate-producing
.crewrig/.synced-markers        excluded under .crewrig          strict
.crewrig/spec-id-carrier        excluded under .crewrig          strict
.github/copilot/settings.json   excluded under .github/copilot   strict
docs/org                        excluded under docs              strict
specs/org                       excluded under specs             strict
```

Four, not two — and the two you named are the ones I had missed. The comment now says four, names them, and states that the dedup is placement-general and does not depend on the count. Worth the fix even though behaviour is unaffected: a wrong number in the comment explaining *why* a guard exists is how the next reader concludes the guard is narrower than it is.

(A note on my own first attempt at that enumeration: I wrote it as a `zsh` loop with `set -- $pair`, and zsh does not word-split unquoted parameters, so every row came back `ABSENT` — a uniform result that looked like an answer. The table above is from `/bin/bash`.)

**On the concurrent branch switching you observed.** Confirmed harmless for this PR, checked rather than assumed: `.worktrees/719` is on `feat/0719-dirty-report-names-files` at the pushed commit, and `git diff` against the remote branch is empty, so nothing you reviewed was contaminated on my side. Your decision to stop trusting its live file state and re-extract through `git show <sha>:<path>` was the right call regardless — a shared worktree's working tree is not a stable read for a reviewer, and that generalizes beyond this incident.

Suite at 71/71, `check-bash32-portability.sh` OK. CI re-running on `178f148`.

Thank you for adding two mutations beyond the four I asked for. Both landed.

PR #861, opened 2026-08-11
🧪 Close MCP daemon coverage gaps (spec 0133)

### LINKED ISSUE #754: 🧪 Close the MCP daemon coverage gaps named but not fixed in #742
The specialist reviews of #742 named several gaps that were acknowledged in commit messages and then not fixed. Collecting them so they are tracked rather than remembered.

## Untested, ranked by consequence

**1. `switch_assistants_to_http` has no end-to-end test.** The transaction is the delta's entire correctness surface, and the one most likely to be wrong, because it only ever runs when something else has already gone wrong. R12/R13/R14/R15 are four requirements and five scenarios currently discharged by primitives plus one structural proxy — an `awk` scrape asserting the rollback list is built before the attempt. That proxy fails closed and is not tautological, but it cannot catch a rollback that runs and restores the *wrong* capture.

Fully hermetic and cheap: fake `$HOME` configs for gemini+copilot+antigravity, stub `mcp_assistant_present`, force a failure on the second CLI, then assert the first is back to `stdio`, the failure was reported, and a re-run converges and says so.

**2. `status-mcp-server.sh` is never executed by any test.** It is the sole discharge of R10, half of R6's operator-readability, and all of R16 — and since #742 it also **gates the switch transaction's exit code**. An untested script became load-bearing. Its three interesting branches (NOT SERVING plus log tail, auth NOT ENFORCED, launcher DRIFTED) are all `rc=1` paths and none is exercised.

**3. Drift is only tested in the direction that cannot bite.** A fresh install is asserted to report IN SYNC. Nothing asserts a *drifted* launcher is reported as drifted — which is the branch an operator's `git pull` actually triggers.

**4. `_materialise_mcp_unit` has no test.** Two placeholders, and `sed` never errors on a non-match, so an unsubstituted `__MEMPALACE_HOME__` silently yields a unit logging to a literal path — breaking the log file R6's readability argument depends on. Three lines close it: `grep -q '__' "$dst"` → fail.

**5. The `null`-capture arm of `restore_mempalace_registration` is untested.** It runs for an assistant that had no mempalace entry before the run — a fresh machine, i.e. the common case for a first-time adopter. If broken, R13 leaves an orphan HTTP entry pointing at a daemon setup just failed to stand up.

## Two review findings left unfixed

**The launcher reimplements the token-path derivation without `common.sh`'s fix.** `mcp_token_path` creates the parent before resolving, precisely because `cd` fails on a missing parent and the path then falls back to its unresolved form, yielding a *different* key — the "the token keeps changing" bug fixed during #742. The launcher's inline copy of that derivation did not get the fix, so the two can diverge under exactly the condition the fix addressed. Its `|| resolved=` fallback is also dead code (verified on Bash 3.2 and 5).

**An assertion that cannot fail:** the "no token value is baked into the launcher" check passes regardless, since the launcher is materialised before any token is minted into the variable it greps for.

## Note on method

Every gap above was found by **mutation** — injecting the regression and re-running — not by reading. Assertions added here should be verified the same way, because five separate assertions in this suite passed against the defect they claimed to guard before being fixed.

Refs #739, ADR 0016.

hcross:
Closes every coverage gap the specialist review of the shared MemPalace MCP HTTP daemon named but did not fix: the switch transaction, the status probe, the drift check, the unit-materialisation guard and the restore null-arm now all carry assertions that can actually fail. The two unfixed review findings — the launcher's token-path derivation and the no-baked-token check — stop being able to diverge or to pass vacuously.

Closes #754

## How to read this PR?

The changeset is three files; read them in this order:

1. **`scripts/tests/test-mcp-daemon.sh`** (+228) — the bulk of the PR. Read the new sections in the order they run: **11** (switch transaction end-to-end, R2), **11b** (repeated run converges a partial state, R15), **12** (null-capture restore, R8), **13a/13b/13c** (the three `status-mcp-server.sh` branches, R4/R5/R6), **14** (placeholder guard, R7), **15** (launcher token path, R9). The no-baked-token assertion (R10) lives in the pre-existing section 3, at the `grep -q "${tok}" "${launcher}"` check.
2. **`scripts/lib/common.sh`** (+13) — the R7 placeholder guard inside `_materialise_mcp_unit`.
3. **`scripts/lib/mcp-daemon-launcher.sh`** (+10/−1) — the R9 token-path derivation.

The non-obvious design decision: **`scripts/status-mcp-server.sh` is byte-identical to `main`.** The new tests exercise it hermetically — a fake `fake-mcp.py` HTTP server stands in for the daemon, and the launcher is installed into a fake `$HOME` — so the status probe's three `rc=1` branches are asserted without touching production code. Every new assertion was verified by mutation (inject the regression, watch it fail, remove it) per spec R1.

## How to test this PR?

The suite is fully hermetic — no real mempalace install, no network, no daemon. Run it from the worktree root:

```sh
```

Expected outcome: `passed: 61   failed: 0` (baseline was 36 passing before this PR). The script exits non-zero on any failure.

### `scripts/lib/common.sh` — R7 placeholder guard (+13)

`_materialise_mcp_unit` now greps the materialised unit for any residual template placeholder matching `__[A-Z][A-Z0-9_]*__`. On a match it prints `ERROR: <dst> still contains an unsubstituted placeholder.` to stderr, removes the output file, and returns 1. Rationale (in the code comment): a unit that logs to `/__MEMPALACE_HOME__/...` or execs the literal `/__LAUNCHER_PATH__` points at a path that does not exist, and `sed` never errors on a non-match, so without the guard an unsubstituted placeholder silently yields a unit logging to a literal path. The only shape `install_daemon_supervisor` ever builds passes through exactly the two known placeholders, so a residual placeholder means the template gained a token this materialiser does not know how to fill.

### `scripts/lib/mcp-daemon-launcher.sh` — R9 token-path derivation (+10/−1)

The launcher's inline token-path derivation now `mkdir -p`s the palace's parent directory before resolving, matching `mcp_token_path` (spec 0113 R8). The dead `|| resolved="${palace}"` fallback is removed. Rationale (in the code comment): `cd` fails on a missing parent, the path falls back to its unresolved form, and a later call — after some other step created the directory — resolves differently and computes a *different* key, i.e. two token files for one palace reads as "the token keeps changing". Creating the parent first makes the key a function of the path alone.

### `scripts/tests/test-mcp-daemon.sh` — new coverage (+228)

- **Section 11 (R2)** — `switch_assistants_to_http` end-to-end against four hermetic configs (gemini, copilot, antigravity, claude). Tools are stubbed on `PATH`; claude starts with no mempalace entry (captures null), the other three start stdio-shaped. A mid-transaction failure is forced on the second assistant switched (gemini) by making its config directory unwritable. Asserts: the failed switch returns non-zero, the failing assistant is reported, the first assistant's entry is restored to its prior arrangement, and the restored claude carries no orphan HTTP entry.
- **Section 11b (R15)** — a repeated run with one assistant (copilot) already switched converges every present assistant, returns success, and reports `found a partial state`.
- **Section 12 (R8)** — the null-capture arm of `restore_mempalace_registration`: an assistant with no prior entry captures `null`, registration switches it to http, and a null-capture restore removes the http entry with no orphan mempalace entry left behind.
- **Section 13a (R4)** — NOT SERVING: with no server on the probe port the probe exits non-zero, prints `NOT SERVING`, and states `no log at` when the log is absent; with a log present it tails the daemon log.
- **Section 13b (R5)** — authentication NOT ENFORCED: a daemon "serving" (healthz 200) but an unauthenticated `/mcp` returning 200 (not 401) makes the probe exit non-zero and report `NOT ENFORCED`.
- **Section 13c (R6)** — launcher drift in the direction an operator's update triggers: a daemon serving with auth enforced (healthz 200, `/mcp` 401) isolates the drift branch, and an installed launcher whose `LAUNCHER_SOURCE_SHA` is rewritten to `deadbeef` is reported `DRIFTED` with a non-zero exit.
- **Section 14 (R7)** — a template with only the two known placeholders materialises and emits a unit; a template carrying an unknown placeholder (`__UNKNOWN_KEYS__`) is refused and emits no unit.
- **Section 15 (R9)** — a palace whose dirname is reached through a symlink and whose subdirectory does not yet exist: the launcher reports a token path and it equals `mcp_token_path` (without the fix the two derivations diverge and this assertion fails).
- **No-baked-token assertion (R10)** — in section 3, `grep -q "${tok}" "${launcher}"` asserts no token value is substituted into the launcher. Per spec R10, the finding that this check cannot be made to fail meaningfully (the launcher is materialised before any token is minted into the variable it greps for) is recorded as moot in the logbook issue's implementation comment.

### Untouched

`scripts/status-mcp-server.sh` is byte-identical to `main`; the tests exercise it hermetically. No new daemon or setup-script behaviour beyond the two named code fixes (R7, R9), per the spec's out-of-scope section.

### DIFF
--- a/scripts/lib/common.sh
+++ b/scripts/lib/common.sh
@@ -786,6 +786,19 @@ _materialise_mcp_unit() {
     -e "s|__LAUNCHER_PATH__|$(mcp_launcher_installed_path)|g" \
     -e "s|__MEMPALACE_HOME__|${HOME}/.mempalace|g" \
     "$src" > "$dst"
+  # Refuse to emit a unit that still carries an unsubstituted placeholder
+  # (spec 0133 R7). A materialised unit that logs to /__MEMPALACE_HOME__/...
+  # or execs the literal /__LAUNCHER_PATH__ points at a path that does not
+  # exist; failing here beats a supervisor that silently restarts a program
+  # that was never installed. A unit that reaches the supervisor only through
+  # the two placeholders above is the only shape install_daemon_supervisor
+  # ever builds, so a residual placeholder means the template gained a token
+  # this materialiser does not know how to fill.
+  if grep -qE '__[A-Z][A-Z0-9_]*__' "$dst" 2>/dev/null; then
+    echo "  ERROR: $dst still contains an unsubstituted placeholder." >&2
+    rm -f "$dst"
+    return 1
+  fi
   return 0
 }
 
--- a/scripts/lib/mcp-daemon-launcher.sh
+++ b/scripts/lib/mcp-daemon-launcher.sh
@@ -59,7 +59,15 @@ die() { printf '%s ERROR: %s\n' "$(date '+%Y-%m-%dT%H:%M:%S%z')" "$*" >&2; exit 
 TOKEN_FILE="${MEMPALACE_MCP_TOKEN_FILE:-}"
 if [ -z "${TOKEN_FILE}" ]; then
   palace="${MEMPALACE_PALACE_PATH:-${HOME}/.mempalace/palace}"
-  resolved="$(cd "$(dirname "${palace}")" 2>/dev/null && pwd)/$(basename "${palace}")" || resolved="${palace}"
+  # Derive the key exactly as mcp_token_path does (spec 0113 R8, spec 0133 R9).
+  # The parent must exist before resolving, or `cd` fails on a missing parent,
+  # the path falls back to its unresolved form, and a later call — after some
+  # other step created the directory — resolves differently and computes a
+  # DIFFERENT key. Two token files for one palace reads as "the token keeps
+  # changing". Creating the parent first makes the key a function of the path
+  # alone, and removes the dead `|| resolved=` fallback that masked the bug.
+  mkdir -p "$(dirname "${palace}")" 2>/dev/null || true
+  resolved="$(cd "$(dirname "${palace}")" 2>/dev/null && pwd)/$(basename "${palace}")"
   if command -v shasum >/dev/null 2>&1; then
     key="$(printf '%s' "${resolved}" | shasum -a 256 | cut -c1-24)"
   else
--- a/scripts/tests/test-mcp-daemon.sh
+++ b/scripts/tests/test-mcp-daemon.sh
@@ -418,6 +418,234 @@ else
   echo "       start a real daemon, so the behavioural lifetime check is skipped."
 fi
 
+# --- 11. Switch transaction, end-to-end (spec 0133 R2, R15) ------------------
+# The transaction's "present" gate is `command -v`, so stub each tool on PATH;
+# the stubs must exist for capture/arrangement/register to treat the four
+# assistants as present without a real install.
+echo ""
+echo "Switch transaction, end-to-end (R2, R11-R15):"
+mkdir -p "${TEST_HOME}/bin"
+for c in claude gemini copilot agy; do
+  printf '#!/usr/bin/env bash\nexit 0\n' > "${TEST_HOME}/bin/${c}"
+  chmod +x "${TEST_HOME}/bin/${c}"
+done
+export ORIG_PATH="${PATH}"
+export PATH="${TEST_HOME}/bin:${PATH}"
+
+# Four hermetic configs. claude starts with no mempalace entry (captures null);
+# the other three start stdio-shaped, giving the transaction prior arrangements
+# to restore to and to converge.
+mkdir -p "${TEST_HOME}/.gemini/config" "${TEST_HOME}/.copilot"
+echo '{"mcpServers":{}}' > "${TEST_HOME}/.claude.json"
+echo '{"mcpServers":{"mempalace":{"command":"bash","args":["c"]}}}' > "${TEST_HOME}/.gemini/settings.json"
+echo '{"mcpServers":{"mempalace":{"command":"bash","args":["p"]}}}' > "${TEST_HOME}/.copilot/mcp-config.json"
+echo '{"mcpServers":{"mempalace":{"command":"bash","args":["a"]}}}' > "${TEST_HOME}/.gemini/config/mcp_config.json"
+chmod 600 "${TEST_HOME}/.claude.json"
+
+# Force a mid-transaction failure on the SECOND assistant switched. `present`
+# is ordered claude, gemini, copilot, antigravity, so gemini is switched second.
+# Making its config DIRECTORY unwritable passes the R12 floor (the file itself
+# stays readable and writable) but makes write_json_config_secure's mktemp fail
+# — the failure lands in the APPLY loop, after claude has already been switched,
+# so the rollback must restore claude. gemini's backup_file silently no-ops
+# (it writes into the same unwritable directory), which is harmless.
+chmod 555 "${TEST_HOME}/.gemini"
+out1="$(switch_assistants_to_http "test-token" 2>&1)"; rc1=$?
+chmod 755 "${TEST_HOME}/.gemini"
+
+[ "${rc1}" -ne 0 ] \
+  && ok "a switch that fails partway returns non-zero" \
+  || nope "the failed switch returned success"
+case "${out1}" in
+  *"gemini could not be switched"*) ok "the failing assistant is reported" ;;
+  *) nope "no report of the gemini failure: ${out1}" ;;
+esac
+[ "$(mcp_assistant_arrangement claude)" = "none" ] \
+  && ok "the FIRST assistant's entry is restored to its prior arrangement" \
+  || nope "claude was not restored after the rollback"
+jq -e '(.mcpServers | has("mempalace")) | not' "${TEST_HOME}/.claude.json" >/dev/null 2>&1 \
+  && ok "the restored claude carries no orphan HTTP entry" \
+  || nope "claude still carries a mempalace entry after the rollback"
+
+# --- 11b. A repeated run converges a partial state (R15) ----------------------
+# Leave one assistant (copilot) already switched to the shared daemon and the
+# rest on their previous arrangement; the repeated run must find that partial
+# state, converge every present assistant, and say it did.
+register_mempalace_mcp copilot "test-token" >/dev/null 2>&1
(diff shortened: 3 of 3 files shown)


hcross:
## Verdict: COMMENT

**Verdict deferred until CI resolves.** All required checks are **pending** (see CI status below), so per the review protocol the formal verdict is deferred. The editorial review below is complete and found **no blocking issues** — the change is sound, the new assertions are meaningful, and the suite passes 61/0. Re-verify once CI is green.

---

### CI status

**Pending** — every required check is still pending (queued) at observation time (`gh pr checks 861`):

- `build`, `check-agents-size`, `check-ci-parity`, `check-components`, `check-feedback-routing`, `check-skill-versions`, `check-spec-id-reserved`, `gitlab-ci-check`, `grep-anti-patterns`, `lint-markdown`, `lint-specs`, `test-harness-curate` — **pending**
- `check-extension-version-bump` — **pass**

No check is failing. The verdict is deferred until the pending jobs resolve; the editorial review below stands independent of CI.

### Correctness

- **R7 placeholder guard** (`scripts/lib/common.sh:789-801`): `_materialise_mcp_unit` now greps the materialised unit for `__[A-Z][A-Z0-9_]*__` after the two known substitutions, and on a match prints an error, removes the output, and returns 1. Correct: `sed` never errors on a non-match, so the guard is the only thing that catches a residual `__MEMPALACE_HOME__`/`__LAUNCHER_PATH__` or a newly introduced placeholder. The regex also catches unknown tokens like `__UNKNOWN_KEYS__` (exercised in section 14). The guard is scoped to the only shape `install_daemon_supervisor` builds, so no false-positive path is introduced.
- **R9 token-path derivation** (`scripts/lib/mcp-daemon-launcher.sh:62-70`): now `mkdir -p`s the palace parent before resolving, matching `mcp_token_path` (`scripts/lib/common.sh:1290-1307`), and drops the dead `|| resolved="${palace}"` fallback. Section 15's symlink + missing-subdir test genuinely exercises the divergence: without the fix the launcher falls back to the literal unresolved path and computes a different key, so the assertion fails. Verified meaningful.
- **Test sections 11-15** (`scripts/tests/test-mcp-daemon.sh:421+`): each new section targets a real branch — the switch transaction's rollback (R2), partial-state convergence (R15), null-capture restore (R8), the three `status-mcp-server.sh` rc=1 branches (R4/R5/R6), the placeholder guard (R7), and the token-path agreement (R9). The hermetic `fake-mcp.py` stand-in and fake `$HOME`/`PATH` stubs make the status probe and switch testable without a real daemon. `scripts/status-mcp-server.sh` is correctly left untouched (test-only coverage, per spec out-of-scope).

### Convention compliance

- **Gitmoji** — PR title `🧪 Close MCP daemon coverage gaps (spec 0133)` and issue title both use the 🧪 emoji; no conventional-commit prefix. Compliant (`AGENTS.md` → Naming Convention).
- **English-only** — code comments and PR body are English. Compliant.
- **Version-bump convention** — **N/A.** The three touched files (`scripts/lib/common.sh`, `scripts/lib/mcp-daemon-launcher.sh`, `scripts/tests/test-mcp-daemon.sh`) are not skill/agent provenance carriers; none carries `metadata.provenance.version` (verified: no match in any file; affected paths in `docs/version-bump-convention.md` are `artifacts/**`/`extensions/**` skill/agent sources only). No bump required.
- **CLI-matrix maintenance** — **N/A.** The touched paths (`scripts/lib/*`, `scripts/tests/*`) are not in the trigger surface (`docs/cli-matrix-maintenance.md` lists `scripts/build-components.sh` and `scripts/{build,install,setup,import,manage}-*.sh`, not `scripts/lib/*` or `scripts/tests/*`). No matrix update required.

### Test coverage

Ran `bash scripts/tests/test-mcp-daemon.sh` in the PR worktree (`.worktrees/754`, branch `feat/0133-close-mcp-daemon-coverage-gaps`): **`passed: 61   failed: 0`**, exit 0 (baseline 36). All new sections 11-15 and the R10 no-baked-token assertion (section 3, `test-mcp-daemon.sh:235-236`) pass. The new assertions are non-tautological — each targets a branch that previously had no failing path, and the R9/R7 tests are written to fail against the pre-fix code.

### Linter findings

`lint-shell.sh` on the three changed files:

- **`scripts/lib/mcp-daemon-launcher.sh`** — clean (exit 0).
- **`scripts/lib/common.sh`** — findings present, but all **pre-existing** (SC2015 at :606/:622, SC2016 at :870/:901/:909/:919, SC2088 at :1398; plus the library's non-executable mode and absent `set -e`, expected for a sourced library). None is in the new guard block (:789-801).
- **`scripts/tests/test-mcp-daemon.sh`** — findings present, but all **pre-existing** (SC2015/SC2016/SC1091/SC2034 at lines ≤371). The new sections (lines 421+) introduce **no** shellcheck findings.

No new linter findings are introduced by this PR.

### Findings

- **class: tech** (non-blocking) — Residual divergence in the launcher token-path derivation. `mcp_token_path` (`common.sh:1298`) retains the `|| resolved="$palace_path"` fallback, but the launcher (`mcp-daemon-launcher.sh:69`) removed it. In the pathological case where `cd` fails even after `mkdir -p` (parent path exists but is not a directory, or permission-denied), the launcher's `resolved` becomes `/$(basename)` (empty dirname) while `mcp_token_path` falls back to the literal palace path — yielding a different key. This does not affect the tested missing-parent condition (where `mkdir` makes the fallback unreachable), so it is non-blocking, but for strict R9 agreement the launcher should mirror `mcp_token_path` verbatim (keep the fallback) or the residual divergence should be acknowledged as accepted. This matches the plan-review finding on issue #754.
- **class: spec** (non-blocking) — R10 moot evidence is documented in the PR body but not yet in the logbook issue's implementation comment. Spec R10 requires the no-baked-token finding to be "recorded as moot with the evidence stated in the implementation's logbook comment." Issue #754 currently carries only the PLAN and PLAN-review comments — no implementation comment. The evidence is present in the PR body ("the launcher is materialised before any token is minted into the variable it greps for"), so this is likely posted at merge time when the logbook issue is closed; flagging so it is not dropped.

---

*Reviewed cold from PR #861 diff, spec `0133-close-mcp-daemon-coverage-gaps` (main), issue #754, and a local run of the suite. Observation marker: PR `updatedAt` 2026-08-11T16:35:11Z.*

PR #875, opened 2026-08-12
✨ Preserve tool capabilities for Antigravity agents (Spec 0137)

hcross:
This PR implements tool capability mapping for the Antigravity CLI compilation target according to Spec 0137. It allows compiling agents to retain their required capabilities (like shell write access or MCP tools) when built for Antigravity.

## How to read this PR?

- [scripts/build-components.sh](file:///Users/hoanicross/devel/perso/genai/framework/crewrig/scripts/build-components.sh): Review the `build_agents` function. We now parse the agent source frontmatter using `yq` to extract `license`, `compatibility`, and an explicit `antigravity` configuration block (`enable_write_tools`, `enable_mcp_tools`, `enable_subagent_tools`). We also check if `claude.allowed-tools` contains `Bash` as a fallback to map `enable_write_tools: true`.
- [artifacts/core/agents/pr-reviewer/AGENT.md](file:///Users/hoanicross/devel/perso/genai/framework/crewrig/artifacts/core/agents/pr-reviewer/AGENT.md): The `pr-reviewer` source definition was updated with a `claude.allowed-tools` section requiring `Bash` and its version bumped to `1.2.0` to verify the mapping logic.
- [docs/cli-matrix.md](file:///Users/hoanicross/devel/perso/genai/framework/crewrig/docs/cli-matrix.md): Documents the new agent tool mapping capability under row 4c.
- The rest of the diff contains the compiled agent definitions under `.agents/agents/...`, `.claude/agents/...`, `.gemini/agents/...`, and `.github/agents/...`.

## How to test this PR?

1. Build the components using `bash scripts/build-components.sh`.
2. Inspect the generated agent definition under `.agents/agents/pr-reviewer/AGENT.md` and check that its frontmatter contains `enable_write_tools: true`.
3. Inspect other compiled agents in `.agents/agents/` (e.g., `accessibility-auditor`) and check that fields like `license` are now preserved.

## Detailed description (for agents)

- **Build Script ([scripts/build-components.sh](file:///Users/hoanicross/devel/perso/genai/framework/crewrig/scripts/build-components.sh))**:
  - Enhanced the agent compilation section for the `antigravity` target.
  - Added extraction of YAML frontmatter metadata using `yq`.
  - Extracted fields: `license` and `compatibility`.
  - Extracted `enable_write_tools`, `enable_mcp_tools`, and `enable_subagent_tools` from an explicit `antigravity` configuration block in the source agent frontmatter.
[...]

PR #877, opened 2026-08-12
📝 Document the shared MemPalace MCP daemon

### LINKED ISSUE #751: 📝 Document the shared MemPalace MCP daemon (ADR 0016 derived spec 5)
The shared MCP HTTP daemon shipped (#742, #750) with its CLI-matrix rows updated, but the narrative documentation still describes only the tier below it. ADR 0016's *Derived spec plan* step 5 planned this; no ticket existed for it until now.

## What is stale or missing, verified

**`README.md` lines 134-150 — the misleading one.** It explains the HNSW corruption class, presents the shared `chroma run` daemon as the fix, and concludes it reduces *"multi-writer contention to a single process"*. That was already optimistic before this work — the MCP tier's contention was untouched, which is what cost an operator ten hours of refused writes on 2026-08-08 — and it is now half the story. A reader learns tier 1 and never discovers tier 2 exists. Zero mentions of the MCP daemon, `task mempalace:switch-http`, or the port.

**No runbook.** `docs/runbooks/` has `chroma-http-server.md` and nothing for the MCP daemon. It needs the same shape: start/stop/status, the stop-versus-uninstall distinction (under a supervisor a stop is a restart request), log location, the port-collision diagnostic, and the token rotation procedure that `uninstall-mcp-daemon.sh` currently prints inline because nothing else documents it.

**ADR 0016 is unreachable from the entry documentation** — referenced nowhere outside `docs/adr/`.

**`docs/concepts.md`** describes the memory tier without its topology.

**`artifacts/core/rules/60-tools.md`** still presents `-32001` ("a peer holds the write lock") as the documented failure mode for friction tagging. Not wrong — it remains the degraded-path net, and spec 0103 stays valid — but it does not say that the case becomes rare once a machine is switched, so an agent reading it will still treat contention as the expected condition.

## Also in scope: an ADR that states something false

ADR 0016 claims `cmd_serve` mints a bearer token on a loopback bind. It does not — `mempalace/cli.py:1450` reads `if not token and not loopback and not args.allow_insecure`, so upstream mints only on a **non-loopback** bind. The claim was mine, it was wrong, it was caught during the plan review, and the correction landed in the plan rather than in the ADR. A merged decision record should not carry a false statement about the code it depends on.

## Not in scope

The CLI matrix rows (7c, 7d, 10) are already updated and stay the machine-readable source of truth; this is the prose that points at them.

Refs #739, ADR 0016 derived spec 5.

hcross:
Documents the shared MemPalace MCP HTTP daemon that ADR 0016 designed but
that no operator-facing page ever explained, and corrects two false
statements in that ADR discovered while writing the correction.

## How to read this PR?

Start with `docs/runbooks/mempalace-mcp-server.md` — it is the new page and
the one every other edit points at. Then `README.md`'s retitled
"Multi-agent concurrency: ChromaDB and MCP daemons" section (the two-layer
framing a reader meets first). `docs/adr/0016-shared-mempalace-mcp-http-server.md`
carries two inline `**Correction (spec 0136, #751):**` blocks — the loopback
bearer-token claim was backwards; both are cited against the installed
MemPalace 3.6.0 source rather than asserted. `docs/concepts.md` and
`artifacts/core/rules/60-tools.md` are short, single-paragraph additions.
`docs/cli-matrix.md` gains one sentence (a runbook pointer) in row 10 — no
per-CLI facts restated. The three files under `docs/assets/mempalace-mcp/`
are generated figures, each with a `.prompt.md` sidecar carrying the exact
prompt, model id, and invocation used to produce it.

Every MemPalace-behaviour claim this PR writes or corrects was verified
against the installed `mempalace` 3.6.0 source before being written, cited
in a logbook comment on #751 at the moment of verification (not batched):
see issuecomment-5266457828, issuecomment-5266485189, and
issuecomment-5266497212.

## How to test this PR?

1. Read the runbook end to end as if converting a real machine — it should
   let you start the daemon, check it, read its log, recognise a port
   collision, and replace its token without opening any script's source
   (the spec's own acceptance scenario).
2. Render the three images and check every label character by character
   against the facts they depict (port `41893`, port `8001`, `ChromaDB`,
   `HNSW`, the four CLI names for the topology figure). For the two
   procedure figures, check both spelling and step order against the code
   they depict — `scripts/lib/common.sh`'s `install_mcp_daemon` for the
   conversion figure, `scripts/switch-mempalace-http.sh` +
   `scripts/stop-mcp-server.sh` for the rotation figure. (A cold review
   caught both figures with a wrong or incomplete step order on the first
   pass — see the correction note below — so this check is not
   theoretical.)
3. Run the four checks this changeset can actually redden:
   ```sh
   markdownlint "**/*.md" --ignore node_modules --ignore extension-skeleton --ignore communication --ignore 'extensions/**/commands/*.md'
   bash scripts/build-docs-index.sh --check
   bash scripts/check-core-paths.sh
   bash scripts/check-no-machine-paths.sh
   ```
   All four exit `0`. (`main` was merged into this branch at `90409a5`,
   which brings in PR #875's fix for `specs/0137` — the two-job red that
   briefly affected every open PR, this one included, is gone.)
4. `bash scripts/build-components.sh` regenerates nothing under version
   control (`artifacts/core/rules/` is deployed by the setup scripts, not
   compiled by this script) — confirm with `git status` before and after.

**`README.md` (lines 133-151 of the prior text).** Retitled
"Multi-agent concurrency: ChromaDB and MCP daemons". Removed the sentence
claiming the ChromaDB daemon alone "reduc[es] multi-writer contention to a
single process" — that conclusion only ever described the ChromaDB tier and
now overclaims. Split into an explicit **Tier 1** (unchanged ChromaDB
content) and **Tier 2** paragraph (the MCP daemon, `task mempalace:switch-http`,
explicitly opt-in and per-machine, linking the new runbook and ADR 0016).
Embeds `docs/assets/mempalace-mcp/topology.png` with alt text, placed before
either tier's prose — the first point a reader meets the topology.

**`docs/runbooks/mempalace-mcp-server.md` (new).** Shaped on the sibling
`chroma-http-server.md`: same `<!-- crewrig-doc: published=false -->`
marker, prerequisites-then-procedures order. Covers conversion
(`task mempalace:switch-http`), daily operations, the stop-vs-uninstall
distinction (a stop under the supervisor is a restart request), the log at
`~/.mempalace/mcp-server.log`, the port-collision diagnostic, and the
token-replacement procedure — ordered so the daemon never keeps serving the
superseded token past the single command in step 2. States plainly that no
automated rotation path exists yet. Embeds the two procedure figures with
alt text.

**`docs/adr/0016-shared-mempalace-mcp-http-server.md`.** Two corrections,
each marked inline, each citing the refuting source, each leaving the
Decision section untouched:
- Lines 150-151 (pre-edit): "on loopback [a token] is optional, and
  `cmd_serve` mints one anyway" was backwards. `mempalace/cli.py:1450`'s
  condition (`if not token and not loopback and not args.allow_insecure`)
  only auto-generates a token for a *non*-loopback bind; the docstring at
  `cli.py:1421` already says so.
- Lines 411-412 (pre-edit, an open question): its premise — that minting a
  token on loopback would "match `cmd_serve`'s own behaviour" — was wrong
  for the same reason; corrected to say it would diverge from it. The
  question itself stays open.

**`docs/concepts.md` (lines 68-77, pre-edit).** One added sentence: the
agent-memory tier is served by a shared process rather than one per session
*under the daemon topology*, linking ADR 0016 and the runbook. Phrased
conditionally to stay consistent with the README's opt-in framing.

**`artifacts/core/rules/60-tools.md` (lines 417 and 463, pre-edit).** Both
places presenting a peer-held write lock as the operative condition now add
that the case becomes rare — not routine — on a converted machine, while
keeping the refusal documented as the degraded path an agent must still
handle either way. The second site cites the mechanism:
`mempalace/palace.py:1114-1120`, `mine_palace_lock`'s re-entrancy docstring
— the same function ADR 0016 already identifies as
`_acquire_mcp_writer_lock()`'s backing lease.

**`docs/cli-matrix.md` row 10.** One added sentence, a pointer to the new
runbook, following the row's existing link-syntax pattern (visible text
prefixed `docs/`, href relative from `docs/` with no prefix) — the doubled-
prefix defect this pattern avoids is live today at
`docs/logbook-issues.md:36`, unrelated to this PR and out of scope to fix
here.

**`docs/assets/mempalace-mcp/*`.** Three figures (`topology.png`,
`convert-machine.png`, `rotate-token.png`), `gemini-3-pro-image`, each with
a `.prompt.md` sidecar (prompt, model id, invocation) and a
`published=false` `crewrig-doc` marker (required by
`scripts/build-docs-index.sh` for every `.md` under `docs/`, sidecars
included — not called out in the plan and caught only by running the
check). `convert-machine.png` needed a second generation attempt in the
original round: the first broke the word "switch-http" across a box
boundary with no hyphen — legible individually but a poor wrap, not a wrong
label; the retry rephrased the step without a long unbroken command token.
`topology.png` and `rotate-token.png` shipped on the first attempt.
`topology.png` is unchanged since; both procedure figures were regenerated
once more in the correction round below.

**Two blocking corrections from cold review (issuecomment-5268119392),
both fixed.** The review found the two procedure figures' underlying claims
wrong, not just their pictures:

1. **`convert-machine.png` had the daemon starting before its token.** The
   real order, per `install_mcp_daemon` (`scripts/lib/common.sh:812-830`):
   the bearer token is minted **first** — the function's first statement,
   and the comment above it says why ("the launcher refuses to serve
   without one") — *then* the launcher and supervisor start the daemon.
   Fixed the runbook's numbered list and regenerated the figure; dropped a
   redundant framing step in the process, so the figure now shows the 4
   steps the code actually performs rather than 5.
2. **The token-rotation procedure's central safety claim was false.** Both
   the runbook and the original figure asserted that `switch-mempalace-http.sh`
   restarts the daemon, which is the step that stops the superseded token
   from being honored. It does not: `install_daemon_supervisor` skips an
   already-loaded launchd agent and no-ops `enable --now` on an
   already-active systemd unit, and the launcher `exec`s the wrapper once
   and never re-reads the token file (`scripts/lib/mcp-daemon-launcher.sh:187`).
   The claim traces to `scripts/uninstall-mcp-daemon.sh`'s own inline
   comment ("restarts, and re-registers") — an upstream defect in
   CrewRig's own tooling, not in MemPalace, filed separately as **#880**.
   Added the missing restart as an explicit step
   (`task mempalace:stop`, between minting the token and cleaning up stale
   backups) and corrected the claim to name that step as the actual point
   of effect. Regenerated `rotate-token.png` with the added step — its
   original four-step order was otherwise correct and is unchanged, so this
   is one inserted step, not a reorder.

Both sidecars record the correction inline, citing the same source lines.
One non-blocking finding from the same review: the R1 citation for the idle
watchdog (logbook comment issuecomment-5266457828) named
`mempalace/mcp_server.py:5071-5081`; the def/docstring actually spans
`5073-5082` (lines 5071-5072 are blank). Corrected in the follow-up logbook
comment; the quoted text itself was byte-accurate.

**Spec status.** A separate, metadata-only commit takes
`specs/0136-document-shared-mempalace-mcp-daemon.md` from `status: approved`
to `status: implemented` per `docs/spec-format.md`'s in-commit transition
mechanic — no other field touched.

**One genuine, non-inherited CI catch, already fixed (`773a47a`).** The
three prompt sidecars originally carried this machine's own absolute path
(`/Users/hoanicross/.claude/skills/nano-banana/scripts/nano_banana.py`) in
their `Invocation:` line. `check-no-machine-paths.sh` (spec 0081) caught it
on CI; my own local run of that check had passed falsely because it was run
before `git add` and the script scans `git grep` (tracked content only) —
an untracked file is invisible to it regardless of content. Fixed by
switching to `$HOME/...`, the pattern's own documented exemption; logged on
#751 (issuecomment-5266785144).

**Rules-file deployment note.** `artifacts/core/rules/60-tools.md` is
deployed to user directories by the four setup scripts; a machine that has
already run setup keeps the prior text until its next setup run.

Closes #751


hcross:
## Verdict: REQUEST CHANGES

Cold review, no authoring context. Identity check: `gh api user --jq .login` and
`gh pr view 877 --json author --jq .author.login` both resolve to `hcross` — the
solo-maintainer case. Per the `pr-reviewer` skill's step-6 ladder, no formal
`gh pr review` event is attempted (GitHub would reject it as a self-review); this
plain comment, opening with the verdict header, is the canonical artifact.

### CI status

10/12 checks pass. `lint-markdown` and `lint-specs` show `FAILURE`.

The PR body's claim that both are **inherited, not introduced** is independently
verified true, historically: `gh run view 31586717103 --json jobs` (the `main`
run at this PR's own recorded base, `374dd64`) fails with exactly
`lint-markdown` and `lint-specs`, nothing else. The root cause matches too —
`specs/0137-antigravity-agent-allowed-tools.md:11` opens with `## Intent`, no H1
(`MD041`), and line 42 uses `*(none)*` as a heading (`MD036`); this PR's own
changed files lint clean (`git diff --name-only crewrig/main...HEAD -- '*.md' |
xargs markdownlint` exits `0`, reproduced independently).

**But the claim is now stale, not false.** `gh api repos/crewrig/crewrig/pulls/877
--jq .base.sha` returns `374dd64…`, while the actual current tip of `main` is
`771f498` (`gh api repos/crewrig/crewrig/git/refs/heads/main`) — one merge ahead
(`git merge-base --is-ancestor 374dd64 771f498` exits `0`; the reverse exits `1`).
That one merge, PR #875 (`771f498`), already carries "`- Add H1 heading to spec
0137 (MD041)`" in its own merge-commit message. Extracting the file at `771f498`
and linting it directly confirms it: `git show 771f498:specs/0137-….md |
markdownlint -` exits `0`, clean. The failing run analyzed here
(`31596377675`) literally logs `Merge 773a47a9… into 374dd64…` — it tested
against the **stale** recorded base, not current `main`. Issue #878 / PR #879
(both still `OPEN`) propose a fix for a defect that a *different*, already-merged
PR (#875) incidentally already repaired.

**Actionable, not just informational:** updating this branch against current
`main` should turn both checks green on its own, without waiting on #879. This
doesn't change the verdict on this PR's own content (below), but the "how to
test" narrative in the PR body should not point at #879 as the pending fix —
that fix already landed, under a different number, for a different reason.

### Correctness

**R1 citations — verified against the pinned MemPalace 3.6.0 source.** All three
citations named for verification are byte-accurate:

- `cli.py:1421` — the `cmd_serve` docstring reads "auto-generating a strong one
  for non-loopback binds" verbatim, confirming the loopback claim was backwards
  as originally written.
- `cli.py:1450` — `if not token and not loopback and not args.allow_insecure:`
  matches verbatim; the branch fires only when `not loopback`.
- `palace.py:1114-1120` — the `mine_palace_lock` re-entrancy docstring matches
  the quoted block verbatim, including the exact composition rationale.

Both ADR 0016 corrections (lines ~150-151, ~408-420 pre-edit) are marked inline,
cite these lines, and leave `## Decision` (line 189 of the current file, well
outside both edited ranges) untouched — confirmed by reading the full section
heading list.

One supplementary citation in the same logbook thread is imprecise: the runbook
comment (`issuecomment-5266457828`) cites `mempalace/mcp_server.py:5071-5081` for
the idle-watchdog docstring, but `_start_idle_exit_watchdog` actually starts at
line 5073 and its docstring closes at 5082 — lines 5071-5072 are blank. The
quoted text itself is accurate; only the line range is off by two. `class: tech`,
non-blocking.

**Figure 1 (`convert-machine.png`) depicts the wrong step order.** The figure's
box 2, the runbook's own numbered list ("1. Installs the supervisor unit… 2.
Provisions a bearer token…"), and the prompt sidecar's claimed verification
("`install_mcp_daemon` → `mcp_token_read_or_create` → …") all place token
provisioning *after* the daemon starts installing. The actual code is the
opposite: `install_mcp_daemon` (`scripts/lib/common.sh:812-830`) calls
`mcp_token_read_or_create` as its **first** statement (line 819), before
`install_mcp_launcher` (824) or `install_daemon_supervisor` (825-830) even run —
per the comment right above it, "The token must exist before the launcher runs:
it refuses to serve without one, by design" (lines 817-818). The sidecar's
claimed order also mischaracterizes the call graph itself: `mcp_token_read_or_create`
is not a step *after* `install_mcp_daemon` returns, it is the first line
*inside* it. This is exactly the "correct words, wrong order" failure mode a
generated procedure figure is supposed to avoid. `class: tech`, blocking.

**Figure 2 (`rotate-token.png`) step order matches `uninstall-mcp-daemon.sh`'s
printed ROTATE text exactly** (delete token file → re-run
`switch-mempalace-http.sh` → delete `.bak` configs → restart sessions) — this
part is correct and I found no discrepancy.

**But the runbook's and figure's central safety claim about that same step 2 does
not hold up against the code.** Both assert that re-running
`switch-mempalace-http.sh` "restarts the daemon under the new value — **the
point at which the superseded token stops being honored**." Tracing
`install_daemon_supervisor` (`scripts/lib/common.sh:583-655`) for the case that
actually applies during a rotation — the daemon is already running, which is
the whole premise of rotating its token — shows it explicitly skips restarting:

- macOS: `if launchctl list | grep -q "$label"; then echo "… already loaded —
  skipping load."` (lines 602-603) — no `launchctl load`/unload happens when the
  agent is already loaded.
- Linux: `systemctl --user daemon-reload && systemctl --user enable --now
  "$systemd_unit"` (lines 620-623) — `enable --now` on an already-active unit
  does not restart it; systemd treats `start` on an active unit as a no-op.

The launcher (`scripts/lib/mcp-daemon-launcher.sh:59-111`) reads the token file
once, validates it, and `export`s it, then `exec`s the wrapper at line 187 —
fixed for the lifetime of that process, never re-read. `mempalace`'s own
`cmd_serve` resolves the token once at server start
(`cli.py`, already cited above), not per-request. No `WatchPaths`
(`config/launchd/com.mempalace.mcp-server.plist`) or path-triggered restart
(`config/systemd/mempalace-mcp-server.service`) exists in either supervisor unit
to pick up a token-file change without an explicit restart, and neither this
script nor the wrapper (`scripts/lib/mempalace-http-wrapper.py`, checked for
`SIGHUP`/reload handling — none) provides one.

Net effect, as far as static tracing can establish: on the realistic path —
rotating a token on a daemon that is already running — step 2 does not appear
to actually stop the daemon from honoring the deleted/superseded token. It
keeps running under the old in-memory value until something else restarts the
OS process (a crash under `KeepAlive`/`Restart=always`, a full
`uninstall-daemon` + reinstall, or a manual `launchctl`/`systemctl` restart) —
none of which this procedure performs. This is a direct hit on spec R5 ("leaves
no daemon serving a superseded value") and on the documentation's own stated
purpose for the step order. I did not run the daemon lifecycle live to avoid
mutating this machine's real launchd/systemd state; the evidence above is from
static source tracing across `common.sh`, the launcher, the plist/unit files,
and the wrapper, and I'd weight it as strong enough to require a resolution
before merge — either a corrected claim (a restart step added to the procedure
and the figure) or a demonstration that I've mis-traced something. `class: tech`,
blocking.

**Everything else checked out.** All 15 spec requirements map to a concrete diff
location (R1 citations above; R2/R3 in the README Tier 1/Tier 2 split; R4 in the
new runbook's coverage of start/stop/check/log/port-collision/token-replacement;
R6 — every task name and env var the runbook cites exists, verified against
`Taskfile.yml` and `scripts/lib/common.sh`/`mcp-daemon-launcher.sh`; R7/R8/R9/R11
in `docs/concepts.md`, `60-tools.md`, and the `cli-matrix.md` pointer-only edit;
R12 in the topology figure's placement; R14 in the three `.prompt.md` sidecars;
R15 in the images' opaque, non-transparent, high-contrast rendering). Scope is
exactly the 13 changed files (`git diff --name-only crewrig/main...HEAD | wc -l`),
matching PLAN v5's declared blast radius plus the one file DEV instruction 3
explicitly authorized (the `specs/0136` status-only commit, `a43b85b`, diffed to
confirm it touches only the `status` field). The "genuine, non-inherited" machine
path defect the PR body describes is real and already fixed in the branch
(`773a47a`); `check-no-machine-paths.sh` now passes.

### Convention compliance

- Every new/added relative link and image reference resolves (target exists and
  is not gitignored, resolved from the linking file's directory): checked all 13
  distinct hrefs/srcs introduced across `README.md`, `docs/cli-matrix.md`,
  `docs/concepts.md`, and `docs/runbooks/mempalace-mcp-server.md`. No misses.
- `docs/cli-matrix.md` row 10's new link follows the documented convention
  (visible text prefixed `docs/`, href relative from `docs/` with no prefix) and
  resolves correctly. The doubled-prefix defect the PR body cites as pre-existing
  and out of scope (`docs/logbook-issues.md:36`) is confirmed real and unrelated
  to this diff — the href there literally reads
  `[`docs/spec-pr-workflow.md`](docs/spec-pr-workflow.md)` from inside `docs/`
  itself, doubling the prefix.
- `docs/index.json:207`'s ADR 0016 entry (title, `nav_order`) is untouched by
  this diff, as claimed.
- The two self-identified process gaps in the logbook — the `crewrig-doc` marker
  requirement on the `.prompt.md` sidecars, and the staleness of PLAN v5's
  "`task spec:lint` excluded because no spec is touched" reasoning once the
  status-only commit landed — are not defects in the shipped artifact. Both were
  caught during DEV, correctly resolved (the marker is present in all three
  sidecars; the reasoning was re-derived and still holds for a different reason),
  and transparently logged (`issuecomment-5266542674`/`…-657311`,
  `issuecomment-5266718...` region). This is the lifecycle working as intended,
  not a gap to route back.

### Test coverage

Not applicable in the usual sense — this is a documentation-only change with no
source/test files touched. The closest analog, verification of behavioral
claims against source, is covered under Correctness above.

### Linter findings

- `markdownlint` (scoped to this PR's own changed `.md` files): clean, exit `0`.
- `bash scripts/build-docs-index.sh --check`: `OK: docs/index.json is up to date
  (32 published pages).`
- `bash scripts/check-core-paths.sh`: clean, both sub-checks `OK`.
- `bash scripts/check-no-machine-paths.sh`: clean.
- `node scripts/lib/spec-linter.js -- specs/0136-document-shared-mempalace-mcp-daemon.md`
  run in isolation: `Linting passed!` — spec 0136 itself carries no defect; the
  repo-wide `task spec:lint` failure is entirely attributable to
  `specs/0137-…md`, per the CI-status section above.

### Summary

Two blocking findings, both `class: tech`, both about figure/runbook accuracy
against the scripts they claim to depict — not about the prose-writing or the
ADR corrections, which check out cleanly:

1. `convert-machine.png` (+ runbook text + prompt sidecar) states the daemon
   starts before the token is provisioned; the code provisions the token first,
   as a precondition for the daemon to even install.
2. The token-rotation procedure's stated safety property — that step 2 is "the
   point at which the superseded token stops being honored" — does not appear to
   hold for the realistic case (daemon already running), based on static tracing
   of `install_daemon_supervisor`, the launcher, and the supervisor unit files.

Neither is a five-minute fix disguised as a big finding: #1 needs the figure
regenerated and the prose/sidecar corrected; #2 needs either a real fix to the
rotation story (a restart step) or, at minimum, an accurate caveat, and probably
a decision from the maintainer on whether that's in scope for spec 0136 or a
follow-up ticket. Everything else — the R1 citations, the ADR corrections, link
resolution, scope discipline, and the CI-inheritance claim's historical accuracy
— held up under independent verification.

hcross:
## Verdict: APPROVE

REVIEW iteration 2 of #877, following up on the first pass
(https://github.com/crewrig/crewrig/pull/877#issuecomment-5268119392). Same
identity check as before (`gh api user` and the PR author both resolve to
`hcross`), so this is again a plain comment rather than a formal review event.

### CI status

Rechecked at `c11baa9` (current PR head, base now `771f498` — current `main`
tip). All required checks pass:
`build`, `check-agents-size`, `check-feedback-routing`, `check-ci-parity`,
`check-extension-version-bump`, `check-skill-versions`,
`check-spec-id-reserved`, `gitlab-ci-check`, `lint-markdown`, `lint-specs`,
`test-harness-curate`, and `check-components` (waited for it to finish rather
than approve while pending — it's a real ~13-14 min job, not flaky).
`lint-markdown`/`lint-specs` are green now that the branch has picked up
`main`'s `771f498` — consistent with what the first pass flagged as stale.

### Correctness — the three findings from iteration 1

**Blocking 1 (step order) — fixed, and re-verified independently.** Read
`convert-machine.png` directly (not just the sidecar's description): four
boxes, "Provision the bearer token" → "Install and start the daemon" →
"Every CLI registered with the token" → "Verify status and auth". That
matches `install_mcp_daemon` (`scripts/lib/common.sh:812-830`) — token first,
daemon second — and matches the runbook's reworded numbered list and its alt
text word for word. No remaining discrepancy.

**Blocking 2 (false safety claim) — fixed, and the corrected procedure holds
up under an adversarial pass.** Read `rotate-token.png` directly: five boxes,
"Delete the old token file" → "Run switch-mempalace-http.sh: mints new token"
→ "Restart the daemon: new token takes effect" → "Delete each CLI's .bak
config (old token)" → "Restart every running session". I traced the two
specific hazards asked about:

- *Does anything re-read a `.bak` before step 4 deletes it?* No.
  `switch_assistants_to_http`'s failure-rollback path (`_switch_rollback`,
  `scripts/lib/common.sh:1053-1075`) restores from the **in-memory** captured
  JSON (`capture_mempalace_registration`, called before any write), not from
  the `.bak` file on disk. The `.bak` is mentioned only as a *manual* recovery
  hint printed if the automatic restore itself fails — an error branch of the
  ordinary conversion path, unrelated to and unaffected by this rotation
  procedure's step order. Nothing in the automated flow reads it, so deleting
  it at step 4 (after steps 2-3 have already fully run) is safe.
- *Does the restart at step 3 race the re-registration step 2 performed?* No.
  These are sequential manual commands a human runs one at a time; step 2
  (`task mempalace:switch-http`) is synchronous and runs to completion —
  including its own trailing `status-mcp-server.sh` check — before returning
  control, so step 3 can't observe a half-finished step 2. `mcp_token_read_or_create`
  (`common.sh:1325-1359`) is read-or-create with `set -C` (noclobber) on the
  write, so the token minted inside `install_mcp_daemon` and the token the
  top-level script re-reads afterward for CLI registration are guaranteed to
  be the same value — no split-token risk between "what the CLIs got" and
  "what step 3's restart will make the daemon read."

One thing I looked for and am **setting aside rather than flagging**: `task
mempalace:stop` (step 3) doesn't wait for or verify the daemon comes back
healthy before the runbook moves to step 4/5, and — untested, speculative —
a fast respawn-after-`launchctl stop` could theoretically race the old
process's socket release on the same port. Neither step 4 nor step 5 depends
on the daemon being immediately reachable, so this isn't an ordering hazard
in the sense asked about; it would only matter if someone tried to use the
daemon in the same instant, which the procedure doesn't ask them to do. Not
worth a finding, but naming it since it's the kind of thing #880's own test
(rotate → probe with old token → expect 401) would incidentally also catch
if it were real.

**Non-blocking 3 (citation line range) — upheld and correctly amended.** The
correction is at `issuecomment-5268322931` on #751 (not `...254433` as
relayed to me — likely a transcription slip, worth knowing about but not
worth re-flagging as a finding). Re-verified `5073-5082` myself against the
installed 3.6.0 tree: matches. Correcting via a new comment rather than
editing the original is the right call under an append-only logbook
convention.

### On the figure-verifiability question

Worth answering with something built, not just discussed. `tesseract` is
already installed on this machine (5.5.3, Homebrew) — I ran it against both
regenerated PNGs rather than reasoning about it abstractly:

```
$ tesseract rotate-token.png stdout
...
Run switch-

mempalace-

http.sh: mints
new token
...
```

OCR recovers every label in both images, including the exact wrapped script
name — de-hyphenate-and-join across the line breaks reconstructs
`switch-mempalace-http.sh: mints new token` byte-for-byte. So an in-figure
string **is** machine-checkable today, with a tool already present in this
environment, without migrating away from the Nano-Banana/raster-PNG path
R14 already sanctions. Concretely buildable as its own script, in the same
family as the `pr-reviewer` skill's bundled linters (degrade gracefully when
`tesseract` is absent, don't hard-fail the check):

`scripts/check-figure-labels.sh` — for every `docs/assets/**/*.png` with a
sibling `.prompt.md`: OCR the PNG, extract the sidecar's own `Box text: "..."`
lines (already written, ordered, one per numbered step — no new authoring
format needed), normalize whitespace and de-hyphenate line-wrapped words, and
assert each expected label appears as a substring in OCR order. That closes
exactly the class of defect blocking-finding-1 was, going forward, and would
have caught the original convert-machine.png regression the moment CI ran on
it rather than waiting for a human (or a reviewer) to eyeball it. I'd file it
as its own ticket rather than fold it into 0136 — it's tooling infrastructure
for the *next* figure, not documentation for this one, and 0136's own
`## Out of scope` already draws that kind of line.

The more fundamental alternative — Mermaid/SVG instead of raster for exactly
the *procedure* diagrams (ordered boxes + arrows are pure text+layout, no
photorealistic content to lose) while keeping Nano-Banana for genuinely
illustrative art like the topology figure — trades the current polish for
diagrams that are directly `grep`-able as source. That's a bigger call
(new diagram convention, not a checker script) and belongs in whoever owns
the diagramming convention's lap, not this review.

### Summary

Both blocking findings are fixed and I independently re-verified the fix
against the code and against the rendered images, not just the sidecar prose.
The corrected 5-step rotation procedure has no ordering hazard I could find
after specifically hunting for one. The non-blocking citation was corrected
properly. CI is green end to end, including the long-running
`check-components` job (waited for it rather than approving on a pending
check). No new findings. **APPROVE.**

Per `AGENTS.md` → *Branching Strategy*: merge authorization is the user's,
not mine or the team-lead's — this verdict clears the review gate, it isn't a
merge instruction.

Issue #880, opened 2026-08-12
🔐 The documented token rotation never revokes the old token


hcross:
The documented token-rotation procedure does not revoke the token. An operator who
follows it because they suspect a leak ends up believing the credential is
replaced while the running daemon keeps honouring the old one.

## The false claim, and where it originates

`scripts/uninstall-mcp-daemon.sh:64` prints the rotation procedure and describes
step 2 like this:

```sh
echo "  2. bash scripts/switch-mempalace-http.sh     # mints a new one, restarts,"
echo "                                               # and re-registers every CLI"
```

**`restarts` is false.** Traced end to end:

| Step | Code | Behaviour |
|---|---|---|
| `switch-mempalace-http.sh:21` | calls `install_mcp_daemon` and nothing else | no stop, no restart, no `launchctl`, no `systemctl` |
| `install_mcp_daemon` (`common.sh:812-830`) | mints the token, installs the launcher, calls `install_daemon_supervisor` | — |
| `install_daemon_supervisor` (`common.sh:602-608`, macOS) | `if launchctl list \| grep -q "$label"; then echo "already loaded — skipping load."` | **explicitly skips** when the daemon is already loaded |
| same (Linux, `620-623`) | `systemctl --user enable --now` | no-op on an already-active unit |
| `mcp-daemon-launcher.sh:187` | reads the token file, then `exec`s the server | the token is fixed for the process's lifetime |

There is no `WatchPaths`, no reload hook, and nothing that re-reads the token file
anywhere in the chain. So on the path that actually applies to a rotation — the
daemon is running, which is the premise of rotating its token — **the old token
stays honoured until something restarts the process.**

`scripts/stop-mcp-server.sh:23-35` is the piece that would do it (`launchctl stop`
on macOS, `systemctl --user restart` on Linux; under a supervisor a stop *is* a
restart request). The rotation procedure never invokes it.

## Impact

This is a security-documentation defect rather than a vulnerability: no privilege
is gained, but the procedure's entire purpose is to make a suspected-leaked
credential stop working, and it does not do that. The operator is told to run it
"if you have any reason to think it leaked".

It also propagated. Spec 0136's runbook (PR #877, in review) documents rotation
faithfully **from this script** — which is what a documentation ticket is supposed
to do — and inherited the false guarantee along with the steps, including into a
generated figure that asserts step 2 is where the superseded token stops being
honoured. The PR review caught it there; this ticket is the upstream cause.

## What to fix

[...]

Issue #4, opened 2026-05-14
✨ GitHub Actions skill, CI configurator agent, and CI debugger agent


hcross:
## Summary

Introduce a new `github-actions` community component bundle covering three related capabilities:
a skill for day-to-day GitHub Actions work, an agent that configures new pipelines from scratch,
and an agent that diagnoses and fixes failing pipelines.

### 1. Skill — `github-actions`

**Path:** `community-config/skills/github-actions/`

A reusable skill that gives any agent deep, up-to-date knowledge of GitHub Actions:
syntax rules, runner behaviour, caching strategies, secret handling, and common pitfalls.

#### `reference/` subdirectory

Curated, version-pinned documentation embedded directly in the skill bundle so agents
work offline and without hallucinating stale API details:

| File | Content |
|------|---------|
| `workflow-syntax.md` | Full annotated `on:` / `jobs:` / `steps:` reference |
| `expressions.md` | Context objects, functions, operators |
| `runners.md` | GitHub-hosted vs self-hosted, OS matrix, resource limits |
| `caching.md` | `actions/cache` patterns, key strategies, size limits |
| `secrets-and-vars.md` | Secrets vs variables, scoping, OIDC patterns |
| `permissions.md` | `GITHUB_TOKEN` permission model, least-privilege patterns |
| `reusable-workflows.md` | `workflow_call`, `workflow_dispatch`, input/output contracts |

#### Scripts

Validation and simulation tools bundled under `scripts/`:

| Script | Purpose |
|--------|---------|
| `lint-workflow.sh` | Validate YAML syntax + schema via `actionlint` |
| `check-pinned-actions.sh` | Flag unpinned action refs (`uses: owner/action@v1` → must be SHA) |
| `check-secrets-exposure.sh` | Grep for patterns that may leak secrets into logs |
| `simulate-job.sh` | Dry-run a single job locally via `act` (Docker-based runner simulation) |
| `simulate-workflow.sh` | Dry-run a full workflow via `act`, with event payload injection |

The `act`-based simulation scripts require Docker and `act` to be installed locally.
They are opt-in; the skill degrades gracefully when `act` is absent.

### 2. Agent — `ci-configurator`

**Path:** `community-config/agents/ci-configurator/`

A specialist agent whose sole mandate is bootstrapping a new GitHub Actions pipeline
for a given project. It interviews the user (language, test framework, target environments,
deployment target, secret inventory) and produces a ready-to-commit workflow file,
[...]

Issue #748, opened 2026-08-08
🐛 The MCP daemon's default port collides on a real machine, and the failure loops instead of diagnosing


hcross:
First real-world run of `task mempalace:switch-http` failed. Two defects, one of which is the more important.

## What happened

```
2026-08-08T20:55:16+0200 starting MemPalace MCP HTTP daemon on 127.0.0.1:8021/mcp
Failed to start MCP HTTP server on 127.0.0.1:8021: [Errno 48] Address already in use
```

…repeated **172 times**, once every ~11 s, until the operator noticed. `KeepAlive` plus `ThrottleInterval=10` is doing exactly what it was configured to do.

Port 8021 was already held on that machine, by `launchd:1` on behalf of a system service — visible via `netstat -anv` but invisible to `lsof` without elevation, so the obvious diagnostic returns nothing. It is also listed in `/etc/services` as `intu-ec-client`.

The transaction behaved correctly: `install_mcp_daemon` refused, no assistant was switched, all four stayed on `stdio`. The all-or-nothing contract held. **This is a usability and diagnosis defect, not a data-safety one.**

## Defect 1 — a poor default

8021 is not a good default. It has an `/etc/services` assignment and it collided on the very first machine the code ran on. The default should be something with no registered claim.

## Defect 2 — the one that matters

The daemon **loops on an unrecoverable condition instead of diagnosing it**.

A port collision is not transient: retrying every 11 seconds for hours cannot resolve it. The launcher already probes ChromaDB on a deadline before handing off — it should probe its own bind the same way, and when the port is taken it should say so once, name what holds it, and exit in a way the supervisor does not immediately retry.

What the operator saw was a task that failed with a log path. What the log gave them was 172 identical lines. The information that would have helped — *this port is held by someone else, here is how to find out who, here is how to choose another* — was never printed, and the plan's "bind failure is loud" mitigation is only true in the sense that it writes to a file nobody was told to read.

`MEMPALACE_MCP_PORT` already exists as an override, which is the correct escape hatch. It is undiscoverable at the moment of failure.

## Suggested

1. Move the default to a port with no registered claim.
2. Pre-flight the bind in the launcher, before the ChromaDB wait. On failure: one message naming the port, stating that `lsof` may show nothing because the holder can be a system service under `launchd`, pointing at `netstat -anv | grep <port>`, and naming `MEMPALACE_MCP_PORT`.
[...]

PR #897, opened 2026-08-13
🔐 Make token rotation actually revoke the old token (spec 0139)

hcross:
Iteration 3: The daemon process replacement contract works end to end. After the printed rotation procedure, the MemPalace MCP daemon accepts the newly minted token and rejects the superseded one. The core mechanism is a single positive accept predicate (2xx on POST /mcp with the current token file value) probed via stdin-fed curl config so the credential never enters the process table. Four iteration-3 test fixes pin the defect-free path.

The operator's printed procedure at `scripts/uninstall-mcp-daemon.sh` now states truthfully that step 2 "mints a new one, replaces the daemon process, and re-registers every CLI" and that the new token only takes effect once the daemon process is replaced (spec 0139 R3).

## How to read this PR?

1. **`scripts/lib/common.sh` — `_mcp_daemon_probe_accepts` function (lines 833–884)** — Accepts three parameters: `host`, `port`, `token`. Issues POST `/mcp` with a single non-notification method (`tools/list`). The bearer token reaches curl via stdin-fed config (`printf 'header = "Authorization: Bearer %s"\n' "$token" | curl -K - …`) rather than `-H` argv, matching the pattern documented at `register_mempalace_mcp` (lines 1039–1044) — on Linux `/proc/<pid>/cmdline` is world-readable, so any local uid sampling the process table harvests argv-borne credentials. Single positive accept predicate: the case statement at line 880–883 returns success only on a 2xx response code. The comment (lines 833–846) documents verification against mempalace's own handler (`mcp_server.py`: a missing or incorrect bearer sends exactly 401; an authenticated non-notification method is answered 200; `/healthz` is `require_auth=False` and would be satisfied by a stale process). No `|| echo "000"` fallback: curl's own `-w '%{http_code}'` already writes literal "000" on a dead connection, and appending a fallback double-writes on that exact path (verified). The token charset claim requires qualification: `mcp_token_read_or_create`'s mint path produces 48 characters of [A-Za-z0-9] (so the config's double-quoted value needs no escaping for minted tokens), but the read path returns whatever a pre-existing token file holds, minus whitespace — an operator-supplied token can carry quotes or backslashes. This is a malformed-header risk, not an injection one: the same whitespace-stripping and config format constraints force a fail-visible outcome (non-2xx), never a silent accept (lines 864–874 document both paths and the risk model).

2. **`scripts/lib/common.sh` — `mcp_daemon_replace_process` function (lines 886–974)** — Takes no parameters. Reads configuration from environment defaults: `MEMPALACE_MCP_LABEL`, `MEMPALACE_MCP_UNIT`, `MEMPALACE_MCP_HOST`, `MEMPALACE_MCP_PORT` (all with fallbacks to `MCP_DAEMON_*_DEFAULT`). Probes with `_mcp_daemon_probe_accepts`; if the current token is already accepted (2xx), returns 0 immediately (no window opened, R5's carve-out per delta-01). Otherwise, prints R5 disclosure (rationale lines 919–933, printed text lines 934–940) at stdout in the voice of the post-uninstall WARNING, then calls the supervisor's restart via `launchctl stop` (macOS, line 945) or `systemctl --user restart` (Linux, line 952) — the same primitives `stop-mcp-server.sh` documents as "a stop IS a restart request" under KeepAlive / Restart=always configuration. The function then re-probes on a deadline: `deadline_s="${MCP_DAEMON_REPLACE_DEADLINE:-15}"` (default 15 seconds, line 913, same budget as `install_daemon_supervisor`'s own health poll at line 637). If the probe accepts within the deadline, returns 0; if the deadline expires, prints error diagnostics and returns 1.

3. **`scripts/tests/test-mcp-daemon.sh` — Section 3 reintroduction guard (lines 268–296)** — New assertion that neither `scripts/lib/common.sh` nor this test file passes a bearer token through an `-H` or `--header` argv flag. The pattern `argv_bearer="Bea""rer"` (line 289) assembles the last syllable at runtime so the assertion line itself cannot satisfy the search it performs. The grep pattern (line 290) admits six forms: `-H` or `--header` flag, space or `=` after the flag, optional opening quote, `Authorization:`, and anything between the colon and the bearer scheme name — a cold review verified all six variant spellings (`-H`, `-H=`, `--header`, `--header=`, optional quote, no separator) were unmatched by earlier forms. Comments are filtered by matching grep's own `<line>:` prefix (line 293), not by leading whitespace. Scope is limited to the two files this spec governs; `scripts/tests/test-setup-org-mcp.sh:199` (org-MCP registration's `--header` shape) is deliberately out of scope per delta-01.

4. **`scripts/tests/test-mcp-daemon.sh` — Section 16 hermetic half (lines 709–777)** — Provisions a fake MCP server on a computed port `HERMETIC_PORT=$((MEMPALACE_MCP_PORT + 1))` (line 738) to avoid TIME_WAIT socket collisions with the behavioural half's real daemon. The comment at lines 728–737 documents: curl POSTs against fake-mcp.py leave TIME_WAIT residue (~20s observed); `scripts/lib/mcp-daemon-launcher.sh`'s port preflight (`:118–143`) does not set SO_REUSEADDR and sees that residue as "already in use", killing the daemon before it ever binds. Runs two fixtures: 200 (accepts immediately; asserts that the early-return path is silent and returns 0) and 401 (never accepts; asserts failure path and deadline diagnostic). Asserts R5 disclosure substring (`*"receives the newly minted token"*`, line 775) as its only executable witness — a substring so copy edits do not break the suite. The test runs under a randomised label with no supervisor unit loaded, so it takes the Darwin `else` branch (line 946) and Linux's inactive-unit branch (line 953) — no process is actually replaced, but the disclosure is printed on all paths that attempt it, by design.

5. **`scripts/tests/test-mcp-daemon.sh` — Section 16 behavioural half (lines 779–947)** — Probed skip if mempalace is not importable (lines 807–810). Defines `_probe_code` (lines 831–837, comment lines 817–830) as a second, independent implementation of the probe pattern used by the helper — deliberately separate so a bug in the lib's curl shape cannot hide behind the assertions meant to catch it. Defines `_wait_bindable` (lines 852–870, comment lines 839–851) as a Python `socket.bind` probe that blocks until the port is actually bindable, matching the same technique `scripts/lib/mcp-daemon-launcher.sh` uses in its preflight (`:118–143`). Installs the launcher, creates token A, starts the daemon, waits for health, probes with A (baseline, non-vacuous assertion). Rotates the token file (creates token B), probes with B (401, the shipped defect: the old process still runs). Calls `mcp_daemon_replace_process` which fails visibly since no supervisor unit is loaded. Manually kill-and-relaunch the daemon, wait for bindability again (accounts for fresh TIME_WAIT from the pre-replace probes), wait for health, probe with B (2xx, the fix: the replaced process accepts it), probe with A (401, R1: the replaced process refuses the superseded token).

6. **`scripts/tests/test-mcp-daemon.sh` — Port guard with both constants and the REAL_MCP_PORT clause** — The LIVE collision guard is at lines 105–110: `|| [ "${MEMPALACE_MCP_PORT}" = "${REAL_MCP_PORT}" ]` (line 109), with capture at line 69 and dedicated diagnostic at lines 113–116. This clause is reachable in practice and is the only one that actively stops a collision. The two production constants — `PROD_PORT="8021"` (line 39) and `PROD_PORT_CURRENT="41893"` (line 51) — are checked at lines 107–108 and serve as unreachable tripwires on the override itself (lines 40–50 document both as inert guards). The override at the suite's top sets the port unconditionally to a random value in 18000–18899, making the constant-value clauses structurally unreachable as the code stands. What actually protects a pre-#748 machine is the override; the two constants are kept as a guard on the override itself — if a future change weakens it to `${VAR:-default}` or drops it, an operator who exports a production port hits the guard already in place.

7. **`scripts/switch-mempalace-http.sh` — Token rotation integration (lines 34–42)** — Calls `mcp_daemon_replace_process` (line 36) after `install_mcp_daemon` (line 21) and before `switch_assistants_to_http` (line 46). A failed replacement exits 1 before any assistant is re-registered, ensuring no assistant is ever switched over a stale credential (spec 0139 R2).

8. **`scripts/uninstall-mcp-daemon.sh` — Printed procedure (lines 61–71)** — Step 2 now reads "mints a new one, replaces the daemon process, and re-registers every CLI". Added explanation that the new token takes effect once step 2 has replaced the daemon process and fails visibly rather than report a completed switch while the daemon still honours the old one (spec 0139).

## How to test this PR?

```bash
# Expected: 73 passed, 0 failed

task spec:lint
# Expected: Linting passed!
```

**Iteration 3 test fixes (developer commit 7ccb691):**

1. **REAL_MCP_PORT operator-port collision clause — the live guard** — Added `REAL_MCP_PORT` capture (line 69: `REAL_MCP_PORT="${MEMPALACE_MCP_PORT}"`), checked in the guard clause (line 109: `|| [ "${MEMPALACE_MCP_PORT}" = "${REAL_MCP_PORT}" ]`), with dedicated diagnostic (lines 113–116). This is the only reachable clause and actively stops an operator-triggered collision: if `MEMPALACE_MCP_PORT` is exported to a production value before running the suite, this clause refuses the run. The two production constants (`PROD_PORT="8021"`, `PROD_PORT_CURRENT="41893"`) are unreachable tripwires on the suite's own override (unconditional 18000–18899 binding at the top, documented at lines 40–50), kept in place so a future weakening of that override is caught immediately.

2. **Chroma-aware skip gate** — Behavioural half probes for mempalace importability (lines 807–810: two-condition gate checking both `import mempalace` and a `curl` GET on Chroma's `/api/v2/heartbeat` at `127.0.0.1:8001`) before starting a real daemon. CI has no mempalace venv and no Chroma, so the skip is clean and quoted rather than silent. Prevents silent non-execution without explanation.

3. **Six-shape argv pattern in reintroduction guard** — Section 3's anti-argv assertion uses a runtime-assembled pattern that admits all six flag/spacing/quote forms: `-H` or `--header`, space or `=` after the flag, optional opening quote, `Authorization:`, and anything between the colon and the bearer scheme name. Pattern last syllable (`"Bea""rer"`) is concatenated at runtime so the assertion line cannot satisfy itself. Comments filtered by grep's own `<line>:` prefix.

4. **Corrected charset comment** — The original charset comment stated tokens are 48 chars `[A-Za-z0-9]` only when minted (which is true), but the READ path of `mcp_token_read_or_create` returns whatever a pre-existing token file holds, minus whitespace — an operator-supplied token can carry quotes or backslashes. This is a malformed-header risk, not an injection one: the same whitespace-stripping and config format constraints mean a malformed header fails visibly (non-2xx), never a silent accept. Lines 864–874 now document both paths and the risk model.

**Spec 0139 status transitions (both files):**
- `specs/0139-token-rotation-revocation.md`: status `approved` → `implemented`
- `specs/0139-token-rotation-revocation.delta-01.md`: status `approved` → `implemented`

Each file is metadata-only; version fields unchanged (spec 0139 at 1.0.0, delta-01 at 1.1.0).

**No version bump or CLI-matrix trigger:**
- `scripts/` paths only; no `artifacts/`, `extensions/`, `config/`, `hooks/`, or CLI-specific paths touched.
- No new provenance carriers.

**Full suite verification:**
73 passed, 0 failed (developer commit 7ccb691 + independent re-run verified).

**No close keywords:**
This PR contains NO close/fix/resolve keywords for #880.


hcross:
## Verdict: APPROVE

Independent cold-start review (PR #897, head `9dbcd3d`). No blocking findings; three non-blocking notes below, two of them PR-body-only (no code change needed).

### CI status

All 13 required checks pass as of this review (`gh pr checks 897`): `build`, `check-agents-size`, `check-ci-parity`, `check-components` (13m15s — consistent with the pre-existing baseline on `main`, where the same `test-mcp-daemon.sh` step already takes ~3m; not a regression introduced here), `check-extension-version-bump`, `check-feedback-routing`, `check-skill-versions`, `check-spec-id-reserved`, `gitlab-ci-check`, `grep-anti-patterns`, `lint-markdown`, `lint-specs`, `test-harness-curate`. `closingIssuesReferences` is empty, matching the PR body's claim of no auto-close keywords.

### Correctness

The diff delivers what spec 0139 requires:

- `scripts/lib/common.sh:847-931` adds `_mcp_daemon_probe_accepts` (positive 2xx-on-`/mcp` accept predicate) and `mcp_daemon_replace_process`. The already-loaded-unit branch it targets is real: `install_daemon_supervisor` (common.sh:602-608) does skip the `launchctl load` when a unit is already running, so without this function the old process would keep answering after a "switch." The 15s default deadline (`deadline_s="${MCP_DAEMON_REPLACE_DEADLINE:-15}"`, line 893) correctly mirrors the budget `install_daemon_supervisor`'s own health poll uses at line 637, exactly as the surrounding comment claims.
- `scripts/switch-mempalace-http.sh:34-42` wires the call between `install_mcp_daemon` and `switch_assistants_to_http`, failing the whole script (exit 1) before any assistant is registered if the process cannot be replaced — satisfies R2's "fail visibly, never report success over a stale credential."
- `scripts/uninstall-mcp-daemon.sh:64-69` — the printed rotation procedure now states when the new token takes effect and that failure is visible, matching the actual script behaviour (R3).
- `scripts/tests/test-mcp-daemon.sh` section 16 (~200 new lines) exercises both the "never accepts → deadline → fail visibly" path and the "already accepting → no restart" path hermetically via `fake-mcp.py`, plus a behavioural half against a real daemon that is correctly guarded to skip cleanly when `mempalace` isn't importable (confirmed: no `pip install mempalace` step exists anywhere in `check-components` before this test runs, so only the hermetic half executes in CI — consistent with the precedent section 13 already established for the same reason).

Two inaccuracies in the **PR body itself** (not the code):

1. The body states the deadline's "production default is 30 seconds" (twice: "How to read this PR" §1 and the `common.sh` bullet in "Detailed description"). The shipped default is **15**, not 30 (`common.sh:893`, matching the cross-referenced `common.sh:637` budget the code comment itself cites). Worth editing the PR body before merge — it's a trivial edit that doesn't touch code or reopen CI.
2. The body attributes the "double-`000` curl fallback removed... at both sites" fix to `scripts/switch-mempalace-http.sh`. That file contains no `curl` calls in this diff at all — the "no `|| echo \"000\"`" reasoning actually lives in `scripts/lib/common.sh:849-854` (`_mcp_daemon_probe_accepts`) and is mirrored in `scripts/tests/test-mcp-daemon.sh` (`_probe_code` helper, ~line 738-745). Likely just a stale file reference from an earlier iteration of the branch.

Given spec 0139 is *specifically* about the printed procedure being truthful, it's worth tightening the PR body to the same standard before it lands as the permanent record.

### Convention compliance

- Gitmoji titles on both commits (`🔐`, `🔖`); no conventional-commit prefixes. ✅
- Branch `feat/0880-token-rotation-revocation` follows `<prefix>/<NNNN>-<slug>` against the tracking issue (#880). ✅
- Spec-PR ordering respected: spec 0139 merged separately first (`c0453e3`, #894) before this implementation branch. ✅
- Status flip (`approved` → `implemented`) rides the implementation PR itself, per `docs/spec-format.md`. ✅
- No version-bump trigger: none of the changed paths (`scripts/**`, `specs/**`) fall under the skill/agent/extension paths enumerated in `docs/version-bump-convention.md`. ✅
- No CLI-matrix trigger: `switch-mempalace-http.sh` and `uninstall-mcp-daemon.sh` don't match the `scripts/{build,install,setup,import,manage}-*.sh` glob (prefixes are `switch-` / `uninstall-`, not `install-`). ✅

### Test coverage

Solid, and matched to the spec's own scenarios (R1-R4 traceability table in the PR body checks out against the diff). One non-blocking observation, security-flavored rather than a coverage gap:

`_mcp_daemon_probe_accepts` (`common.sh:855-859`) and the test's mirrored `_probe_code` (`test-mcp-daemon.sh:~741-745`) pass the live bearer token as a literal `curl -H "Authorization: Bearer ${token}"` argument. That token sits in the probing curl process's argv — world-readable via `/proc/<pid>/cmdline` on Linux for the process's lifetime — and this call repeats every 0.3s while `mcp_daemon_replace_process` polls to its deadline, not just once. This is exactly the exposure vector `common.sh:996-1000` documents and deliberately avoids for `claude mcp add --header` ("on Linux /proc/<pid>/cmdline is world-readable, so any local uid sampling the process table during setup harvests the credential"). It isn't a new class of risk this PR introduces, though: the other three CLIs' registration path (`write_json_config_secure`'s `jq --arg auth "Bearer ${token}" ...`, e.g. `common.sh:1013`, `1021`) already puts the same token in `jq`'s argv today, pre-existing and unrelated to this diff. Not blocking here, but worth a follow-up ticket to harden all these call sites together (e.g. a `curl -K` config passed via process substitution) rather than adding one more instance ad hoc next time.

### Linter findings

`lint-shell.sh` on the four changed shell files reports `SC1091`, `SC2015`, `SC2016`, `SC2034`, `SC2088`, plus "not executable" for `common.sh` (mode `100644`, expected `100755`) and "missing `set -e`" for `common.sh`. Diffed against the same linter run against `main`'s versions of these same files: **identical category set**, and `common.sh`'s `100644` mode is unchanged from `main` (`git ls-files --stage` confirms). No new lint categories — matches the PR body's own claim exactly.

`lint-markdown.sh specs/0139-token-rotation-revocation.md` — exit 0, no findings.

hcross:
## Verdict: REQUEST CHANGES

Cold review, iteration 2. No authoring context used — the diff, the files at
`headRef` (`7ca9587`), the specs and the test run below are the only inputs.

**The shipped code is sound and I found nothing blocking in it.** Both
iteration-1 concerns are genuinely resolved, and I verified them by mutation and
by running the suite, not by reading. What blocks is the PR body: the rewrite
that was supposed to eliminate iteration 1's wrong mechanism claims still
carries six of them, two of which name identifiers that do not exist in this
repository. `## Detailed description (for agents)` is a durable artifact for
downstream readers; shipping it wrong is the same defect iteration 1 raised.

---

### CI status

All 13 required checks **pass** at head `7ca9587`
(`gh pr checks 897 --repo crewrig/crewrig`): `build`, `check-agents-size`,
`check-ci-parity`, `check-components`, `check-extension-version-bump`,
`check-feedback-routing`, `check-skill-versions`, `check-spec-id-reserved`,
`gitlab-ci-check`, `grep-anti-patterns`, `lint-markdown`, `lint-specs`,
`test-harness-curate`. `check-components` was `in_progress` when I began and
resolved to `pass` before this verdict was written. `mergeStateStatus` is
`UNSTABLE` only because of that in-flight job.

---

### Correctness

**Iteration-1 concern 1 — token out of the process table: resolved, verified.**
Both probe sites feed the bearer through a stdin curl config (`-K -`):
`common.sh:868-872` and `test:783-788`. `printf` is a bash builtin, so the token
never reaches an `execve` argv. The token does travel as a positional argument
to the *shell function* (`common.sh:848`, `common.sh:908`), which is in-process
and not observable in `/proc/<pid>/cmdline` — the security property holds.

I mutation-tested the reintroduction guard rather than reading it. Restoring
`-H "Authorization: Bearer $token"` inside the probe at `common.sh:868-872`
makes `test:213-220` fire (`a bearer token reached argv via -H:
…/common.sh:870`); the unmutated baseline is silent, so the assertion is not
vacuous. The comment filter is load-bearing as documented: it correctly spares
the prose at `common.sh:1033`, which quotes the very flag being banned.

**Iteration-1 concern 2 — R5 disclosure: resolved, conformant.** The
disclosure at `common.sh:927-933` sits in straight-line code between the
accept-immediately return (`common.sh:908-910`) and the `case`
(`common.sh:935`), so it precedes every replacement attempt and is silent only
on the path delta-01 exempts. It names the port, the window, and the
consequence — matching R5's "whoever holds the port receives the newly minted
token" (`specs/0139-token-rotation-revocation.delta-01.md:15-24`).

**The fresh-install path does not regress.** `switch-mempalace-http.sh:21`
runs `install_mcp_daemon` first and aborts on failure, so by the time
`mcp_daemon_replace_process` runs at `:36` a daemon is already serving. On a
fresh load it started on the current token, the probe accepts, and no window is
opened. The only path that issues a stop is the already-loaded one — which is
exactly the shipped defect this ticket exists to fix.

**Test suite: 73 passed, 0 failed — reproduced.** I ran
`bash scripts/tests/test-mcp-daemon.sh` against a clean clone of
`feat/0880-token-rotation-revocation` at `7ca9587`. Both halves of section 16
executed (this box has `MEMPALACE_PYTHON` pointing at a venv where `mempalace`
imports), including the five behavioural assertions. The body's count is
accurate.

---

### Blocking — the rewritten body still misdescribes the shipped code

All six share one root: the body describes an *intended* design rather than the
code that shipped. Fixes are mechanical; no code change is needed.

1. **`class: tech` — the probe helper's signature.** Body item 1 says "The probe
   helper **takes no arguments and reads the token from the current token
   file**". It takes three: `common.sh:833` documents `<host> <port> <token>`
   and `common.sh:848` is `local host="$1" port="$2" token="$3"`. The token is
   read by the *caller* at `common.sh:905` and passed positionally at
   `common.sh:908` and `common.sh:958`.

2. **`class: tech` — the probe helper is not called from the switch script.**
   Body item 1 cites "(call at `switch-mempalace-http.sh:36`)". Line 36 calls
   `mcp_daemon_replace_process`. `_mcp_daemon_probe_accepts` is called only from
   `common.sh:908` and `common.sh:958`.

3. **`class: tech` — the supervisor identifiers do not exist.** Body item 2 says
   the helper "calls `launchctl stop mempalace-mcp-daemon` (macOS) or
   `systemctl --user restart mempalace-mcp-daemon` (Linux)". There is no label
   or unit named `mempalace-mcp-daemon` anywhere in the repo — the string
   appears only inside spec *slugs*. The shipped defaults are
   `MCP_DAEMON_LABEL_DEFAULT="com.mempalace.mcp-server"` (`common.sh:725`) and
   `MCP_DAEMON_UNIT_DEFAULT="mempalace-mcp-server"` (`common.sh:726`), matching
   `config/launchd/com.mempalace.mcp-server.plist` and
   `config/systemd/mempalace-mcp-server.service`. A reader following item 2
   verbatim runs a command that acts on nothing.

4. **`class: tech` — the hermetic half does not rotate anything.** Body item 3
   says it "rotates the token, and verifies the old token fails (401) while the
   new token succeeds (2xx)". The hermetic half (`test:709-748`) writes one
   fixed 48×`a` token at `test:710` and runs `fake-mcp.py` in 200-mode then
   401-mode. Its four assertions are: return 0 plus silence on the
   already-accepting path (`test:717-722`), non-zero on the never-accepting
   path (`test:729-731`), the deadline diagnostic (`test:732-735`), and the R5
   substring (`test:745-748`). No second token is ever minted here, and no 401
   is ever asserted *of a superseded token* — `fake-mcp.py` is simply told to
   answer 401 (`test:724`). The old-vs-new rotation assertions are in the
   *behavioural* half, at `test:845-848` and `test:870-879`.

5. **`class: tech` — the behavioural half deliberately does *not* follow the
   rotation procedure.** Body item 3 says it "follows the rotation procedure".
   It asserts the opposite: `test:850-854` requires
   `mcp_daemon_replace_process` to **fail** ("fails visibly when it cannot
   replace the process"), then replaces the process by hand with `pkill`
   (`test:859`) and a direct relaunch (`test:865`). The test's own comment says
   so at `test:856-858`. What it covers is the replacement *effect*, not the
   supervisor path — a distinction the body erases.

6. **`class: tech` — `_wait_bindable` line range.** Body item 3 cites
   `test:760-778`. The function is at `test:803-821`; `test:760-778` is comment
   prose plus the `_probe_code` header.

---

### Convention compliance

- **Gitmoji** — PR title and all four authored commits conform (`AGENTS.md` →
  *Naming Convention*).
- **PR body sections** — all four required sections present
  (`docs/pr-format.md`).
- **Close-keyword trap** — `gh pr view 897 --json closingIssuesReferences`
  returns `[]`. The body's claim is verified; #880 stays open for the
  post-merge flow.
- **Logbook (Rule B)** — #880 carries 8 comments, the last at
  `2026-08-14T08:39:56Z`. Updated incrementally, not at the end.
- **Version bump** — no skill or agent source touched; `check-skill-versions`
  passes. The body's claim is correct.
- **CLI matrix** — I checked the four changed script paths against the
  authoritative trigger list in `docs/cli-matrix-maintenance.md:10-16`. None
  match `scripts/{build,install,setup,import,manage}-*.sh` or any other trigger
  (`uninstall-mcp-daemon.sh` is not `install-*.sh`). No `docs/cli-matrix.md`
  update owed. The body's claim is correct.
- **Branch naming** — `feat/0880-token-rotation-revocation` conforms.
- **Spec status** — both `0139` and `0139.delta-01` flip `approved` →
  `implemented`; `lint-specs` passes. The body's *Status transition* section
  mentions only delta-01, though the diff flips both — worth a sentence when
  fixing the items above.

---

### Test coverage

Coverage is genuinely strong, and unusually honest about its own limits
(`test:750-763` names the untestable supervisor precondition instead of faking
it). Four non-blocking observations:

7. **`class: tech` — the new production-port guard clause cannot fire.**
   `test:91` adds `|| [ "${MEMPALACE_MCP_PORT}" = "${PROD_PORT_CURRENT}" ]`, but
   `test:57` has already run
   `export MEMPALACE_MCP_PORT="$((18000 + (RANDOM % 900)))"` unconditionally.
   The comparison therefore tests the script's own random 18000–18899 assignment
   against 41893 and is unreachable. The comment at `test:40-45` claims it
   "would silently stop protecting pre-#748 machines" otherwise — it protects
   nothing as written. (The pre-existing clauses at `test:88-90` share the
   defect, so this PR extends a dead guard rather than introducing one.) Fix:
   `export MEMPALACE_MCP_PORT="${MEMPALACE_MCP_PORT:-$((18000 + (RANDOM % 900)))}"`,
   or drop the port clauses and say why they are redundant.

8. **`class: tech` — the prerequisite skip covers only half the prerequisites.**
   `test:765` gates the behavioural half on `import mempalace`, but the launcher
   also needs a reachable ChromaDB: `mcp-daemon-launcher.sh:154-160` dies when
   the port is unreachable within `MEMPALACE_MCP_CHROMA_WAIT`, which this suite
   pins to 2s at `test:67`, and `test:766` points it at the operator's real
   `8001`. A developer with `mempalace` installed but Chroma stopped gets
   `FAIL the daemon never became healthy` (`test:881`) and exit 1 — a red suite
   for a missing prerequisite rather than the clean skip the body claims under
   R4. Probing `127.0.0.1:8001` alongside the import check would restore it.

9. **`class: tech` — the anti-argv pattern is narrower than its claim.** It
   catches the canonical shape (mutation-verified above), but I ran four
   variants through `test:214` and all four pass unnoticed:
   `--header="Authorization: Bearer $t"`, `-H'Authorization: Bearer $t'`,
   `-HAuthorization:\ Bearer\ $t`, and a double space after the colon. Cheap
   widening: allow `[= ]`, an optional quote, and `[[:space:]]+` after the
   colon. Not blocking — the guard demonstrably catches the shape that was
   actually removed.

10. **`class: tech` — the charset invariant holds only for minted tokens.**
    `common.sh:864-866` justifies leaving the config's double-quoted value
    unescaped with "the token is 48 characters of `[A-Za-z0-9]` (see
    `mcp_token_read_or_create`)". That is guaranteed for tokens that function
    *mints* (`tr -dc 'A-Za-z0-9' | head -c 48`); on the read path it returns any
    pre-existing file's contents with whitespace stripped. Config-line injection
    is impossible — newlines are stripped by that same `tr -d '[:space:]'` — so
    the worst case is a malformed header and a fail-visible replacement. One
    clause in the comment ("for tokens this function minted") would close the
    gap between what is claimed and what is enforced.

---

### Linter findings

`bash .claude/skills/pr-reviewer/scripts/lint-shell.sh` on the four changed
files at `headRef`, compared against the same command on `crewrig/main`:

```
base (main):  42 SC2015    head:  48 SC2015
all other shellcheck codes: identical counts
```

The six new `SC2015` instances are all the file's established
`[ cond ] && ok "…" || nope "…"` assertion idiom (42 pre-existing occurrences).
`ok()` at `test:33` always returns 0, so the `A && B || C` hazard does not
apply. No new finding class is introduced; nothing to fix.

`lint-shell.sh` also reports `scripts/lib/common.sh: missing 'set -e'` — expected
and correct for a sourced library, unchanged from `main`.

---

### What unblocks this

Fix items 1–6 in the PR body (and optionally the *Status transition* sentence).
No code change is required; items 7–10 are observations for the author's
judgement, not conditions of merge. One more consideration for whoever picks
this up: items 1, 4 and 5 all describe a design where the probe owns the token
read and the hermetic half owns the rotation. If that design was real at some
point, the body may be describing a draft that the shipped code moved past —
worth checking that the fix rewrites from `7ca9587`, not from the earlier text.

hcross:
## Verdict: REQUEST CHANGES

Cold review at `7ccb6911ea1010859c1709b642628e1fe92acf2a`, PR body observed at `updatedAt=2026-08-14T09:21:28Z`.

**The code is clean and I verified it by running and mutating it — all four iteration-3 test fixes are genuinely resolved.** The block is again editorial: the body carries one retracted mechanism claim, omits one mechanism it is named after, and its line anchors point 17–52 lines away from the code they name.

## CI status

All 13 required checks **pass** at this HEAD, including `check-components`, which was still pending at the start of this review and has since completed (13m26s).

`build`, `check-agents-size`, `check-ci-parity`, `check-extension-version-bump`, `check-feedback-routing`, `check-skill-versions`, `check-spec-id-reserved`, `lint-markdown`, `lint-specs`, `test-harness-curate`, `check-components`, `gitlab-ci-check`, `grep-anti-patterns` — no failures, none pending.

## Correctness

I ran the suite rather than reading it. `bash scripts/tests/test-mcp-daemon.sh` at this HEAD: **73 passed, 0 failed**, exit 0 — the body's claim reproduces. The behavioural half genuinely executed (`MEMPALACE_PYTHON` resolved to a venv with `mempalace` importable and ChromaDB reachable at `127.0.0.1:8001`), so the five rotation assertions are real, not skipped.

### The four iteration-3 test fixes — all confirmed resolved

1. **Port-collision guard clause** — `REAL_MCP_PORT` captured at `scripts/tests/test-mcp-daemon.sh:69`, checked at `:109`, diagnostic at `:113-116`; `PROD_PORT_CURRENT="41893"` at `:51`, checked at `:108`. Resolved.
2. **Chroma-aware skip gate** — two-condition gate at `:807-810`. I exercised **both** arms: with a python lacking `mempalace` the suite printed `skip mempalace is not importable …` and finished 67/0; with `mempalace` present and the heartbeat pointed at a dead port the gate selected the ChromaDB arm. Resolved.
3. **Six-shape argv pattern** — the shipped pattern at `:290` catches **6/6** shapes I threw at it (`-H "…"`, `-H '…'`, `-H=…`, `--header "…"`, `--header=…`, `-HAuthorization:…`). The comment filter at `:293` is correct in both directions: a real code violation is flagged, a commented one is filtered. **Mutation-tested against the real suite**: injecting `curl -H "Authorization: Bearer $tok"` into an inert function in `scripts/lib/common.sh` flips the assertion to `FAIL a bearer token reached argv via -H:` and the suite to exit 1. The guard is non-vacuous. Resolved.
4. **Corrected charset comment** — `scripts/lib/common.sh:864-874` now documents both the mint path and the read path and states the malformed-header-not-injection risk model. Resolved **in the code**; see F1 for the body.

### Security

No concerns. The bearer reaches curl through a stdin config (`-K -`) on both the library probe (`common.sh:875-879`) and the test's deliberately independent `_probe_code` (`:831-837`) — never argv. Ordering in `scripts/switch-mempalace-http.sh` is right: `install_mcp_daemon` (`:21`) guarantees a serving daemon before `mcp_daemon_replace_process` (`:36`), and a failed replacement exits 1 **before** `switch_assistants_to_http` (`:46`), so no assistant is ever registered over a stale credential. The R5 window disclosure prints on every path that attempts a replacement (`common.sh:934-940`). `scripts/lib/mcp-daemon-launcher.sh:118-143` confirmed to bind without `SO_REUSEADDR`, which is exactly the TIME_WAIT rationale the tests cite.

I left no residue: the production daemon on `:41893` still answers `/healthz`, and no orphan test daemon or `fake-mcp.py` process remains.

## Blocking findings

### F1 — §1 restates the claim iteration 3 retracted (`class: tech`)

*How to read this PR?* item 1 still says:

> Token charset is 48 characters of [A-Za-z0-9] by construction (from `mcp_token_read_or_create`), so the config's double-quoted value needs no escaping.

That is the unqualified claim fix #4 exists to correct, and the body contradicts itself: *Detailed description* item 4 says the read path "returns whatever a pre-existing token file holds, minus whitespace — an operator-supplied token can carry quotes or backslashes," matching `common.sh:864-874`. Item 1 asserts a security property (no escaping needed) that this PR retracted. Qualify it to the mint path and point at the read path's risk model.

### F2 — the "operator-port collision clause" is described without the collision clause (`class: tech`)

*Detailed description* item 1 is titled **"Port guard operator-port collision clause"** but its prose covers only `PROD_PORT_CURRENT`, and §6 enumerates the guard as "Both constants are checked in the guard clause." The guard at `:105-110` has **four** port clauses, and the one the fix is named after — `|| [ "${MEMPALACE_MCP_PORT}" = "${REAL_MCP_PORT}" ]` at `:109`, with its capture at `:69` and its dedicated diagnostic at `:113-116` — is never mentioned. That clause is the only one that is reachable in practice; the two production constants are, as the code comment at `:40-50` itself says, unreachable tripwires on the override. The body describes the inert half and omits the live half.

### F3 — the line anchors do not point at the code they name (`class: tech`)

The body states it was "rewritten exclusively from the diff at HEAD." Fourteen anchors do not resolve at HEAD; nine misdirect a reader into unrelated sections. They match neither HEAD nor the previous commit `7ca9587` consistently, so they appear to be approximate rather than read off any one revision. Corrected table — this should be a single mechanical edit:

| Body location | Claimed | Actual at `7ccb691` |
|---|---|---|
| §1 `register_mempalace_mcp` argv prose | `common.sh:996-1001` | `1039-1044` (function at `1033`); `996-1001` is `mcp_assistant_config_path`/`mcp_daemon_url` |
| §2 `mcp_daemon_replace_process` span | `886-979` | `886-974` |
| §2 R5 disclosure | `919-930` | rationale `919-933`, printed text `934-940` |
| §2 `launchctl stop` | `946` | `945` (`946` is the `else`) |
| §2 `systemctl --user restart` | `952` | `952` — correct |
| §4 hermetic half span | `697-752` | `709-777` |
| §4 `HERMETIC_PORT=` | `686` | `738` |
| §4 TIME_WAIT comment | `705-718` | `728-737` |
| §4 R5 substring assertion | `750` | `775` |
| §4 Darwin `else` branch | `947` | `946` (`947` is the warning echo) |
| §4 Linux inactive-unit branch | `953` | `953` — correct |
| §5 behavioural half span | `754-896` | `779-947` |
| §5 import/Chroma gate | `758` | `807-810` |
| §5 `_probe_code` | `766-780` | `831-837` (comment `817-830`) |
| §5 `_wait_bindable` | `782-798` | `852-870` (comment `839-851`) |
| §6 port guard clause | `88-90`, line `90` | `105-110`; `PROD_PORT_CURRENT` clause at `108` |
| §3 all anchors (`268-296`, `289`, `290`, `293`) | — | correct |
| §3 `test-setup-org-mcp.sh:199` | — | correct, exact |
| §5 launcher preflight `:118-143` | — | correct, exact |
| §1 probe fn `833-884`, case `880-883` | — | correct |

## Non-blocking observations

- **NB1** (`class: tech`) — §3 says "all four variant spellings" and then lists five items. The pattern's `[[:space:]=]*` is zero-or-more, so it admits a sixth no-separator form the same paragraph describes elsewhere. Say six, or drop the count.
- **NB2** (`class: tech`) — §4 and §5 cite `mcp-daemon-launcher.sh`; the path is `scripts/lib/mcp-daemon-launcher.sh`.
- **NB3** (`class: tech`) — the off-by-one anchors (§2 `946`, §4 `947`) are folded into F3's table but are cosmetic on their own.
- **NB4** (`class: tech`) — `shellcheck` reports 2 warnings across the changed files: `common.sh:1554` (SC2088, tilde in quotes) and `test-mcp-daemon.sh:121` (SC2034, `CREWRIG_REPO_DIR` unused). **Both are outside this diff's hunks and pre-existing** — no action required here. The remaining 55 findings are all `info`-level (SC2015/SC2016/SC1091), idiomatic for this suite's `&& ok || nope` style.

## Convention compliance

- **Gitmoji** — title `🔐 Make token rotation actually revoke the old token (spec 0139)` conforms; all six commits carry gitmoji.
- **Branch naming** — `feat/0880-token-rotation-revocation` matches `<prefix>/<NNNN>-<slug>`.
- **PR body template** — all four required sections present.
- **Close keywords** — `closingIssuesReferences` is empty. The body's claim is verified against the API, not just read.
- **Version bump** — no `artifacts/`, `extensions/`, or provenance carriers touched; `check-skill-versions` passes.
- **CLI matrix** — no triggering path in the diff. `scripts/lib/common.sh` is not in the `AGENTS.md` trigger list, and neither changed top-level script matches `scripts/{build,install,setup,import,manage}-*.sh`. The body's claim holds.
- **Spec status** — both `specs/0139-*.md` files flip `approved` → `implemented`; `lint-specs` passes.
- **Logbook** — issue #880 open, correctly not auto-closed by this PR.

## Test coverage

Strong, and I confirmed it is not coverage theater. Section 16 (`:709-947`) covers the ticket's actual defect: the baseline assertion (`:883-887`) is non-vacuous, the shipped defect is asserted live at `:894-897` (401 for the new token while the old process runs), and R1 revocation is asserted at `:925-928` (the replaced process refuses the superseded token). The hermetic half pins both the early-return silence and the deadline-expiry diagnostic. Section 3's reintroduction guard is mutation-proven above.

## Linter findings

`lint-markdown.sh` on both changed spec files: exit 0, no output.

`lint-shell.sh` on all four changed shell files: exit 1, driven entirely by info-level notes plus the two pre-existing warnings in NB4. The one substantive note is on an unchanged line:

```
/tmp/pr897work/scripts/lib/common.sh: missing 'set -e' (or 'set -euo pipefail') near the top
```

`common.sh` is a sourced library, not an entry point; `set -e` there would change the failure semantics of every caller. Not a finding against this PR.

---

**To unblock:** fix F1 (qualify the charset claim), F2 (describe the `REAL_MCP_PORT` clause), and F3 (apply the corrected anchor table). No code change is required — the implementation and its tests are verified correct at this HEAD.

hcross:
## Verdict: REQUEST CHANGES

Cold review, iteration 4, at head `7ccb6911ea1010859c1709b642628e1fe92acf2a`;
PR body observed at `updatedAt=2026-08-14T09:45:00Z`. Inputs: the diff, the six
changed files plus four referenced ones read at `headRef` through the contents
API, the two specs, and a local run of the suite.

**All three iteration-3 findings and all four nits are genuinely resolved — I
re-derived every one of them against the files rather than against the corrected
table.** The code remains clean and I found nothing blocking in it.

What blocks is narrower than last round but it is the same artifact and the same
class: `## How to read this PR?` never mentions `scripts/switch-mempalace-http.sh`
— the file that actually invokes the helper this PR exists to add — and the
agents-facing section names a mechanism the code does not use.

---

## CI status

All 13 required checks **pass** at this head (`gh pr checks 897`), none pending,
none failing: `build`, `check-agents-size`, `check-ci-parity`, `check-components`
(13m26s), `check-extension-version-bump`, `check-feedback-routing`,
`check-skill-versions`, `check-spec-id-reserved`, `gitlab-ci-check`,
`grep-anti-patterns`, `lint-markdown`, `lint-specs`, `test-harness-curate`.

`gh pr view 897 --json closingIssuesReferences` returns `[]` — empty, matching
the body's "No close keywords" claim, so merging will not auto-close #880.
`mergeStateStatus` is `CLEAN`, `mergeable` is `MERGEABLE`.

---

## Iteration-3 findings — all three resolved

**F1 (retracted charset claim) — resolved.** *How to read this PR?* §1 now reads
"The token charset claim requires qualification: `mcp_token_read_or_create`'s
mint path produces 48 characters of [A-Za-z0-9] … but the read path returns
whatever a pre-existing token file holds, minus whitespace". I checked that
against the function itself rather than against the comment: the mint path is
`LC_ALL=C tr -dc 'A-Za-z0-9' < /dev/urandom | head -c 48` at
`scripts/lib/common.sh:1491`, and the read path is
`existing="$(tr -d '[:space:]' < "$token_file")"` at `common.sh:1476`. Both
halves of the body's sentence hold, the malformed-header-not-injection framing
matches `common.sh:864-874`, and §1 no longer asserts the unqualified
no-escaping-needed security property. The body no longer contradicts itself.

**F2 (live collision clause omitted) — resolved.** §6 is now titled "Port guard
with both constants and the REAL_MCP_PORT clause" and its first sentence leads
with the live half: "The LIVE collision guard is at lines 105–110:
`|| [ "${MEMPALACE_MCP_PORT}" = "${REAL_MCP_PORT}" ]` (line 109), with capture at
line 69 and dedicated diagnostic at lines 113–116." *Detailed description* item 1
leads the same way. Verified at head: capture at `test-mcp-daemon.sh:69`, clause
at `:109`, diagnostic at `:113-116`, and the inertness of the two constants is
exactly what the code comment at `:40-50` says, given the unconditional override
at `:74` (`export MEMPALACE_MCP_PORT="$((18000 + (RANDOM % 900)))"`). The body
now describes the live half first and the inert half as inert.

**F3 (stale anchors) — resolved.** I re-verified **all 37** anchors in the body
against the files at this head, not the 10 asked for. **36 resolve exactly**,
including every row the corrected table touched:

- `common.sh` — `833-884` (probe fn), `833-846` (comment), `864-874` (charset),
  `880-883` (case), `886-974` (replace fn), `913` (deadline), `919-933`
  (rationale), `934-940` (printed text), `945` (`launchctl stop`), `946`
  (Darwin `else`), `952` (`systemctl restart`), `953` (Linux `else`), `637`
  (health-poll budget), `1039-1044` (`register_mempalace_mcp` argv prose).
- `test-mcp-daemon.sh` — `39`, `40-50`, `51`, `69`, `105-110`, `109`, `113-116`,
  `268-296`, `289`, `290`, `293`, `709-777`, `728-737`, `738`, `775`, `779-947`,
  `807-810`, `817-830`, `831-837`, `839-851`, `852-870`.
- `scripts/lib/mcp-daemon-launcher.sh:118-143` (port preflight, `socket.bind`, no
  `SO_REUSEADDR`) — exact. `scripts/tests/test-setup-org-mcp.sh:199` — exact.
  `scripts/uninstall-mcp-daemon.sh:61-71` — matches the diff hunk `@@ -61,8 +61,12 @@`.

The one that does not resolve exactly is cosmetic and folded into F4's edit below.

## The four nits — all resolved

- **NB1** — §3 now says "six forms" and enumerates six (`-H`, `-H=`, `--header`,
  `--header=`, optional quote, no separator). Count and list agree.
- **NB2** — §4 and §5 both now write the full path
  `scripts/lib/mcp-daemon-launcher.sh`.
- **NB3** — the two off-by-ones are corrected: `945`/`946` and `946`/`953` all
  land on the lines the body names.
- **NB4** — `shellcheck -S warning` over the four changed scripts still reports
  exactly the two pre-existing warnings, both outside this diff's hunks:
  `common.sh:1554` (SC2088) and `test-mcp-daemon.sh:122` (SC2034). The only
  `warning`-level output; no action here, as agreed.

---

## Findings

### F4 — `How to read this PR?` omits the file that wires the feature (`class: tech`)

`scripts/switch-mempalace-http.sh` is one of the six changed files and the only
one that makes this PR take effect in production:

```
@@ -31,6 +31,16 @@
+if ! mcp_daemon_replace_process; then
+  echo "ERROR: the daemon process could not be replaced — it may still be"
+  echo "       honouring a superseded token. No assistant has been switched over"
+  echo "       a stale credential."
+  exit 1
+fi
```

The body never mentions it. `grep -ni "switch-mempalace-http" ` over the body
returns nothing; the only occurrence of the word "switch" is inside §7's quotation
of the *uninstall* script's printed text. §1–§2 describe the helper,
§3–§6 describe the tests, §7 describes the printed procedure — the reading guide
covers five of six files and skips the call site.

This is not a cosmetic gap. The body's opening sentence asserts the end-to-end
outcome ("After the printed rotation procedure, the MemPalace MCP daemon accepts
the newly minted token and rejects the superseded one") without ever naming the
code that produces it, so a reader of the body alone cannot tell that
`mcp_daemon_replace_process` is invoked anywhere at all — the helper reads as
wired to nothing. It also drops R2's fail-visibly guarantee, which lives entirely
in that hunk: the script aborts *before* `switch_assistants_to_http`, so no
assistant is ever registered over a stale credential. That ordering is the
requirement, and it is undocumented.

Corroboration that the omission actually costs readers: both prior cold reviews
had to locate and cite `scripts/switch-mempalace-http.sh:34-42` from the diff
themselves — it appears in the Correctness section of the iteration-2 and
iteration-3 verdicts, but in neither case because the body pointed there.

`docs/pr-format.md` asks *How to read this PR?* for "a reading guide to help
reviewers navigate the changeset — highlight key files", and *Detailed
description (for agents)* for "every change made in this PR … explicit about what
was added, modified, or removed and why". Neither section covers this file.

**To fix:** add one item to *How to read this PR?* for
`scripts/switch-mempalace-http.sh` (lines 34–42 at head) stating that it calls
`mcp_daemon_replace_process` between `install_mcp_daemon` (`:21`) and
`switch_assistants_to_http`, and that a failed replacement exits 1 before any
assistant is re-registered (R2).

### F5 — the Chroma skip gate is described as a mechanism it does not use (`class: tech`)

*Detailed description* item 2 says the gate is:

> two-condition gate checking both `import mempalace` and `socket.connect` to Chroma at `127.0.0.1:8001`

The second condition is not a socket connect. At
`scripts/tests/test-mcp-daemon.sh:807-810` it is an HTTP GET through curl against
Chroma's v2 heartbeat endpoint:

```sh
if "${MEMPALACE_PYTHON}" -c 'import mempalace' >/dev/null 2>&1 \
   && curl -sf --max-time 2 \
        "http://${behav_chroma_host}:${behav_chroma_port}/api/v2/heartbeat" \
        >/dev/null 2>&1; then
```

`grep -n "socket" ` over `:800-816` returns nothing. The distinction is not
pedantic in this file: a `socket` probe *does* exist 40 lines later —
`_wait_bindable` at `:852-870` uses `socket.bind` — so "socket.connect" points a
reader at the wrong one of two adjacent probes. It is also a weaker check than
what ships: a connect proves something is listening, the heartbeat proves Chroma
is serving its v2 API.

This is the same class as F1 — a mechanism claim in the agents-facing section
that the code does not support. It is milder, because the anchor beside it
(`807-810`) is correct and lands the reader on the truth in four lines.

**To fix:** replace `socket.connect` with the heartbeat probe, e.g. "a
`curl` GET on Chroma's `/api/v2/heartbeat` at `127.0.0.1:8001`".

---

## Non-blocking observation

- **NB5** (`class: tech`) — §6 says the two production constants "are checked at
  line 108". Line `108` checks `PROD_PORT_CURRENT` only; `PROD_PORT` is checked
  one line earlier at `:107`. This is the corrected table's §6 row
  ("`PROD_PORT_CURRENT` clause at `108`") applied to both constants rather than
  one. Purely cosmetic — it sits inside the `105-110` span the same sentence
  already cites correctly — so fold `107–108` into F4's edit rather than treating
  it as its own round.

---

## Correctness

Not re-audited from scratch — the implementation is twice-verified and I sampled
rather than repeated that work. What I did check independently:

- **The suite reproduces.** `bash scripts/tests/test-mcp-daemon.sh` at this head
  in a detached worktree: **73 passed, 0 failed**, exit 0, and `grep -n skip` over
  the output returns nothing — so the behavioural half genuinely executed here
  rather than skipping, and the five rotation assertions are real, including
  "the replaced process refuses the superseded token (R1: rotation revokes it)".
  The body's "73 passed, 0 failed" claim and its *How to test this PR?* command
  are accurate as written.
- **The 15s deadline cross-reference holds.**
  `deadline_s="${MCP_DAEMON_REPLACE_DEADLINE:-15}"` at `common.sh:913` and
  `local deadline=$((SECONDS + 15))` at `common.sh:637` — the body's "same budget"
  claim is exact.
- **The R5 disclosure placement is what §2 claims.** The `echo` block at
  `common.sh:934-940` sits in straight-line code between the accept-immediately
  return (`:915-917`) and the `case "$(uname -s)"` (`:942`), so it precedes every
  path that proceeds past the early return, and is silent only on the path
  delta-01 exempts.
- **The argv-reintroduction guard is scoped as §3 says.** The grep at `:290-292`
  covers `common.sh` and this test file only; `test-setup-org-mcp.sh:199` does
  carry the `--header Authorization: Bearer` argv shape and is deliberately out of
  scope per delta-01.

## Convention compliance

- **Title** — `🔐 Make token rotation actually revoke the old token (spec 0139)`:
  Gitmoji, no conventional-commit prefix. Conforms to *Naming Convention*.
- **Branch** — `feat/0880-token-rotation-revocation`, `<prefix>/<NNNN>-<slug>`.
  Conforms.
- **PR body sections** — all four present in order. §*How to read* and §*Detailed
  description* are incomplete per F4; the structure itself is right.
- **Version bump** — no `artifacts/`, `extensions/`, `config/`, or `hooks/` paths
  touched, so `docs/version-bump-convention.md` does not fire. `check-skill-versions`
  passes. The body's claim is correct.
- **CLI matrix** — `scripts/` paths only, none matching the
  `scripts/{build,install,setup,import,manage}-*.sh` trigger list, so
  `docs/cli-matrix-maintenance.md` does not require a `docs/cli-matrix.md` update.
  Correct.
- **Spec-PR ordering** — specs 0139 and its delta-01 are already on `main`; this
  PR only flips `status: approved` → `implemented` on both (one line each), which
  is the DEV-stage close-out, not a spec change. Frontmatter otherwise intact
  (`complexity: small`, `interaction-mode: MINIMAL`, `version` unchanged at 1.0.0
  and 1.1.0). `lint-specs` passes.
- **Close keywords** — none; `closingIssuesReferences` is `[]`. Conforms to the
  logbook rule that #880 is closed after merge, not by it.

## Test coverage

Adequate and non-vacuous. Section 16 covers both halves of the contract: the
hermetic half (`:709-777`) pins the already-accepting early return (returns 0,
prints nothing) and the never-accepting deadline path (non-zero, diagnostic),
and the behavioural half (`:779-947`) pins the real rotation — probe with A
accepts, rotate to B and probe B (401, the shipped defect), replace, probe B
(2xx) and probe A (401, R1). The R5 disclosure is asserted by substring at `:775`
so copy edits do not break the suite. The `_probe_code` helper at `:831-837` is
deliberately a second implementation of the curl shape rather than a call into
the library's, which is the right call — it keeps a bug in the lib's probe from
hiding behind the assertions meant to catch it. Section 3's reintroduction guard
(`:268-296`) was mutation-tested in iteration 3 and I did not repeat that.

## Linter findings

`lint-shell.sh` over the four changed shell scripts — exit 0. Only `info`-level
output (SC2015 on the suite's `&& ok || nope` idiom, SC2016, SC1091), plus the
two pre-existing `warning`s recorded under NB4. Nothing introduced by this diff.

`lint-markdown.sh` is not applicable to the changed set beyond the two spec files,
whose one-line frontmatter edits are covered by the passing `lint-markdown` and
`lint-specs` CI jobs.

---

**To unblock:** one mechanical body edit — add the
`scripts/switch-mempalace-http.sh` item (F4), replace `socket.connect` with the
heartbeat probe (F5), and correct `108` to `107–108` (NB5). No code change is
required; the implementation and its tests are correct at this head and CI is
green.

hcross:
## Verdict: APPROVE

Iteration 4, continued — closing the thread opened at
[#issuecomment-5291978661](https://github.com/crewrig/crewrig/pull/897#issuecomment-5291978661).
Same head `7ccb6911ea1010859c1709b642628e1fe92acf2a`; body re-observed at
`updatedAt=2026-08-14T09:56:31Z` (was `09:45:00Z` when I wrote the block).
Body-only edit, no code touched, so this remains one review pass rather than a
new iteration.

**All three findings are resolved in-pass. I approve.**

I diffed the new body against the one I blocked on rather than re-reading it
whole: the edit touches exactly three places and nothing else.

### F4 — resolved

*How to read this PR?* has a new item 7 for `scripts/switch-mempalace-http.sh`,
and the former uninstall item is renumbered to 8. The reading guide now covers
all six changed files.

I verified every anchor in the new item against the file at head, not against
the wording I proposed:

- `lines 34–42` — the added block runs from `echo ""` at `:34` to `fi` at `:42`.
- `mcp_daemon_replace_process` at `:36` — exact.
- `install_mcp_daemon` at `:21` — exact.
- `switch_assistants_to_http` at `:46` — exact.
- "A failed replacement exits 1 before any assistant is re-registered" — the
  `exit 1` is at `:41`, inside the `if !` block that closes at `:42`, and the
  registration call is at `:46`. The ordering claim holds.

The R2 attribution is correct and worth stating plainly, because it is the
reason this item mattered: `specs/0139-token-rotation-revocation.md:28-33` names
`scripts/switch-mempalace-http.sh` explicitly and requires that "when the daemon
process cannot be replaced, the script SHALL fail visibly rather than report
success over a stale credential." That requirement lives entirely in this hunk,
and the body now points at it.

### F5 — resolved

*Detailed description* item 2 now reads "a `curl` GET on Chroma's
`/api/v2/heartbeat` at `127.0.0.1:8001`", matching
`scripts/tests/test-mcp-daemon.sh:807-810` exactly. `grep -n "socket.connect"`
over the body returns nothing. The remaining `socket` probe in the file
(`_wait_bindable`, `:852-870`) is no longer shadowed by a wrong description of
its neighbour.

### NB5 — resolved

§6 now says the two production constants are "checked at lines 107–108".
`PROD_PORT` is at `:107`, `PROD_PORT_CURRENT` at `:108`. Correct.

### Nothing else degraded

- No stale cross-references survived the renumbering: `grep -n "item 7\|§7\|item 8\|§8"`
  over the body returns nothing, so no prose points at the old numbering.
- Item 8's own anchors (`61–71`) are unchanged and still match the diff hunk
  `@@ -61,8 +61,12 @@`.
- Every other anchor in the body is untouched by this edit and was verified
  exhaustively in the previous comment — 37 of 37 now resolve at this head.
- `closingIssuesReferences` is still `[]`, `state` is `OPEN`, head is unchanged,
  so CI remains green at 13/13 from the run recorded in the previous comment.

---

## Standing summary for the merge decision

- **Code** — clean. Twice independently verified across iterations 2 and 3, and
  sampled again here; I ran the suite myself at this head in a detached worktree:
  **73 passed, 0 failed**, exit 0, zero skips, so the behavioural half genuinely
  executed and the R1 assertion "the replaced process refuses the superseded
  token" is not vacuous.
- **CI** — 13/13 required checks pass, none pending, none failing.
- **Conventions** — Gitmoji title, `feat/0880-token-rotation-revocation` branch
  shape, all four PR-body sections present and now complete, no version-bump or
  CLI-matrix trigger, spec 0139 and delta-01 frontmatter intact at
  `status: implemented`.
- **Auto-close** — `closingIssuesReferences` is empty, so merging will not close
  #880; close it manually after merge per the logbook rule.

No blocking findings and no non-blocking observations remain.

Issue #913, opened 2026-08-14
🔐 Harden the daemon credential paths — socket handoff, argv sweep, guard fixes


hcross:
Consolidated security-hardening follow-up from the #880 lifecycle (spec 0139 + delta-01, merged in PR #897). Four items on the same credential surface, batched per the cost discipline of issue #882 instead of one ticket each.

## 1. Eliminate the replacement window (structural)

`mcp_daemon_replace_process` (spec 0139 R5) discloses the residual risk instead of eliminating it: between the supervisor stop and the relaunch, the TCP port is fully released, and a local process winning the bind race receives the newly minted token from the first probe and can fake a 2xx rotation success. The structural fix is holding the socket across the replacement — a systemd socket-activation unit (and the launchd equivalent) so no gap exists for another process to claim. Spec 0139 delta-01 explicitly routes this here. Origin: security pass on PR #897, iteration 1, HIGH finding 2.

## 2. Sweep the remaining bearer-in-argv call sites

PR #897 moved the rotation probes' Authorization header out of curl argv (stdin config, `curl -K -`). One deliberate exception remains repo-wide: `scripts/tests/test-setup-org-mcp.sh:199` asserts the org-MCP config shape on purpose. Audit any other credential-bearing command construction (`ps`-visible argv, `/proc/<pid>/cmdline`) across `scripts/` and align on the stdin-config pattern. Origin: security pass on PR #897, HIGH finding 1 (resolved for the rotation path; the sweep is the residue).

## 3. Two LOW hardening fixes on the anti-argv test guard

From the security re-check at 7ccb691, both proven escapable by mutation, deferred here by user decision at the merge gate (four lines total):

- The guard pattern (`scripts/tests/test-mcp-daemon.sh:290`) is case-sensitive; `-H "authorization: bearer $token"` (valid for curl, headers are case-insensitive) escapes it. Fix: `grep -i`.
- The comment filter (`:293`) is unanchored: `grep -v ':[[:space:]]*#'` drops a genuine violation line that carries a trailing comment containing `: #`. Fix: anchor to the line-number prefix, `grep -v '^[0-9]*:[[:space:]]*#'`.

## 4. One INFO copy improvement on the R5 disclosure

[...]

PR #948, opened 2026-08-14
🔐 Harden the anti-argv guard and disclose the rotation recovery (spec 0149)

hcross:
The anti-argv guard in the daemon suite was a static grep whose passing state looked identical whether or not it could still catch a violation, and spec 0149 names two mutations it let through. This PR closes those, closes a third vacuity the review found underneath them — a sweep pointed at a path that no longer exists — widens the guard to the credential flags `curl` supplies itself, and adds the recovery action to the rotation-window disclosure.

Spec: [`specs/0149-daemon-credential-hardening.md`](https://github.com/crewrig/crewrig/blob/main/specs/0149-daemon-credential-hardening.md) · Logbook: issue #913 · Interaction mode: MINIMAL · Complexity: small · **Review iteration 2**

Branch is up to date with `main` at `bfb320e`. All line anchors below are at `9d7ee5a`, re-derived after the refresh — `main`'s `7deb998` inserted an assertion at `:142` of the same suite, so every anchor in this file moved by +10 since iteration 1.

## How to read this PR?

Four commits, four files. Read in this order.

**1. `specs/0149-daemon-credential-hardening.md`** — the contract everything traces to (R1–R5). The only change is the frontmatter `status: approved` → `status: implemented`, in its own commit (`87dd7b3`); no body line, no `id`, no `slug`, no `version` (`docs/spec-format.md` → *Recording a status transition*).

**2. `scripts/tests/test-mcp-daemon.sh` §3 (`:258-415`)** — the load-bearing file, and the only one iteration 2 changed. Four design decisions, in the order they appear:

- **A missing swept path is a violation, not a silent skip** (`:347-350`). This is the review's blocking finding and the reason for iteration 2. `grep` on a missing file yields no hits and `2>/dev/null` ate the diagnostic, so renaming either swept file left every assertion green while the repository was searched for nothing — byte-identical to the shipped green state. The check lives **inside** `_argv_bearer_hits`, on exactly the paths it is about to search, rather than in a second list at the call site: a second list can drift from the first as the swept set grows, this cannot. A missing path is recorded as a violation-shaped entry (`<path>:0:MISSING …`), so no caller can skip past it, and the call site asserts the absence of that marker at `:409-412` **separately from and before** the sweep verdict at `:413-415`. That separation is the point: a clean sweep and a sweep that read nothing are the same empty string, so emptiness carries information only once the paths are known to have been read.
- **Three argv shapes, one alternation** (`:331-333`). The header shape is the one the daemon probes could regress into; the other two are the likelier *innocent* reintroduction, a contributor reaching for curl's own credential flags, which put the raw token in argv just as plainly as a hand-built header does — `--oauth2-bearer <token>` (shape B, `:332`) and `curl … -u/--user "<user>:<pass>"` (shape C, `:333`).
- **Shape C alone is anchored** on `curl` appearing on the same line and on the `user:password` colon, because bare `-u` is common in ordinary shell. That anchor is *defensive rather than currently load-bearing*, and the distinction is measured, not assumed: removing it leaves this suite green at `81/0`, because neither swept file contains a matching line. Its value shows up one directory wider — over all of `scripts/`, the unanchored form produces exactly one false positive, `scripts/worktree-claim.sh:493` (`now_iso() { date -u +%Y-%m-%dT%H:%M:%SZ; }`), while the anchored form produces none. The anchor is what would let this shape survive a widened sweep; today it costs nothing and buys that.
- **The guard proves it can still catch a violation** (`:364-401`). Five fixture lines are written under `${TEST_HOME}` (`:371-377`, removed by the existing `cleanup()` trap at `:102-103`) and swept through the *same* helper as the real files: the lower-case shape R1 names, the trailing-comment shape R2 names, a pure comment that must stay excluded, and one line per credential-flag shape. Both halves of the banned string are still assembled at runtime (`:316-319`) — once R1 makes the match `-i`, any line here that spells the header name out self-trips the pattern. The hit-count assertion at `:399-401` (now `-eq 4`) remains the one that kills the degenerate "exclude nothing" filter, which would satisfy every named requirement while leaving the guard useless.

  One diagnostic fix rides along (`:354-359`): hits are prefixed with their file **per line** rather than per blob, so the second and later violations of a file no longer carry `grep`'s bare `<line>:` for the reader to attribute by inference.

**3. `scripts/lib/common.sh:940`** — one added `echo` (from `a1a0600`, untouched by iteration 2), inside the existing unconditional stdout block of `mcp_daemon_replace_process`, in the surrounding WARNING's voice and column alignment. It satisfies spec 0149 R3, which makes a disclosure that names the risk without naming a recovery non-conformant. **The security pass then found that the recovery it names does not actually recover** — rotating again re-runs the same vulnerable path and hands the squatter the new token too. That is a behaviour change (identify the process holding the port, evict it, verify the port is free), not a copy edit, so it is routed to **#951** by user decision at the REVIEW gate rather than fixed here. This line ships as the literal R3 requirement; #951 tracks making it true.

**4. `scripts/tests/test-mcp-daemon.sh` §16 (`:902`)** — the witness that makes the line above falsifiable. `:894` already asserted the risk half; `:902` adds the recovery half. `case` is case-sensitive, so the asserted substring tracks `common.sh:940` verbatim.

**5. `scripts/tests/test-setup-org-mcp.sh:199-207`** — the R4 deliberate-exception marker, immediately above the assertion it explains at `:208`. Note what the comment deliberately does *not* do: it never quotes the banned argv shape, because a third occurrence would break R4's own "no other occurrence in `scripts/`" clause that the sweep below verifies.

## How to test this PR?

Prerequisites: `bash`, `git`, and `task` (for `spec:lint`). No MemPalace venv is needed — the daemon suite skips the probes that require one.

**1. The two suites this diff touches.**

```sh
bash scripts/tests/test-mcp-daemon.sh       # exit 0 — "passed: 81   failed: 0"
bash scripts/tests/test-setup-org-mcp.sh    # exit 0 — "RESULT: 41 passed, 0 failed"
```

`81`, not the `80` measured before the branch refresh: `main`'s `7deb998` (spec 0159, symlinked palace token path) added one assertion to this same suite, which the merge brought in.

Seven assertions in §3 belong to this guard:

```
  ok   the guard catches a violation spelled in lower case (R1)
  ok   the guard catches a violation carrying a trailing comment (R2)
  ok   the guard catches curl's own token flag (shape B)
  ok   the guard catches basic auth passed through argv (shape C)
  ok   the anchored filter excludes the comment line and nothing else (R2)
  ok   every swept path exists, so a clean sweep below means something
  ok   no credential is passed through an argv flag (R8)
```

plus `ok   the disclosure names the recovery, not only the risk (0149 R3)` in §16.

**2. Spec lint and Bash 3.2 portability.**

```sh
task spec:lint                          # exit 0 — "Linting passed!" (212 files)
bash scripts/check-bash32-portability.sh
# exit 0 — "OK: no forbidden Bash 4+ construct and no unguarded array value
# expansion in scripts hooks (3 declared construct(s), 171 file(s) scanned,
# 157 .sh file(s) array-scanned)."
```

**3. R4's sweep clause — exactly two matches under `scripts/`, and exactly the two named.**

```sh
B="Bea""rer"; grep -rniE -- "-(H|-header)[[:space:]=]*[\"']?Authorization:.*${B}" scripts/
```

```
scripts/tests/test-setup-org-mcp.sh:208:[[ "$HTTP_STR" == *"--header Authorization: Bearer"* ]] && ok "http argv: --header \"K: V\"" || bad …
scripts/lib/common.sh:1040:      # NOT `claude mcp add --header "Authorization: Bearer $token"`: on Linux
```

The first is the deliberate exception this PR marks (a fixture-driven *shape* assertion, whose token is the literal placeholder `${ATLASSIAN_TOKEN}` from the `ORG_HTTP` fixture at `:58`); the second is prose explaining the hazard.

**Failure modes — mutate the guard and watch it redden.** This is the check worth running, because a green state that carries no information is the exact defect this PR exists to remove. Every row below was produced at `9d7ee5a` by mutating `scripts/tests/test-mcp-daemon.sh` in place and running the **whole** suite; each restores with `git checkout -- scripts/tests/test-mcp-daemon.sh`.

| Mutation | Suite | Exit | Assertions that redden |
|---|---|---|---|
| **as shipped** | `81 / 0` | 0 | — |
| `:404` swept path renamed to a nonexistent file | `79 / 2` | 1 | `every swept path exists`; the sweep verdict (`a credential reached argv: …common-RENAMED.sh:0:MISSING`) |
| `:351` `grep -niE` → `grep -nE` (drop `-i`) | `79 / 2` | 1 | R1 lower-case; hit count (`got 3`) |
| `:332` shape B removed from the alternation | `79 / 2` | 1 | shape B; hit count (`got 3`) |
| `:333` shape C removed from the alternation | `79 / 2` | 1 | shape C; hit count (`got 3`) |
| `:333` shape C's `curl` anchor removed | `81 / 0` | 0 | none — see the anchoring note in *How to read this PR?* |

Row 2 is the one iteration 2 exists for: before this change that same mutation left the suite at `77 / 0`, exit 0, indistinguishable from a clean run.

### Requirement traceability

| Req | Change | Executable witness |
|---|---|---|
| **R1** — match case-insensitively | `:351` carries `grep -niE`; `argv_hdr` (`:317`) joins `argv_scheme` (`:316`) as a runtime-assembled half, because `-i` makes any line spelling the header name self-trip the pattern | `:380-383` — fixture line 1, lower-case `authorization: bearer`; killed by dropping `-i` |
| **R2** — comment filter excludes only comments | `:352` anchors on `^[0-9]+:`; the sweep is per-file (`:344-362`) so the anchor can match at all | `:384-387` — fixture line 2, a violation carrying `# ref: #913`; plus the hit-count assertion at `:399-401` |
| **R3** — disclosure names the recovery | `common.sh:940` — `Rotate the token again if the window may have been claimed.` | `:902`, paired with the pre-existing risk witness at `:894`. **Sufficiency of the named recovery is tracked in #951**, not resolved here |
| **R4** — deliberate occurrence marked at its site | `test-setup-org-mcp.sh:199-207` — 9 comment lines above `:208`, naming it deliberate, citing spec 0149 R4, recording that the asserted string is the org-MCP argv *shape*, not a live credential | The sweep in *How to test* §3: exactly 2 matches, exactly the two named |
| **R5** — structural fix requested upstream | No code change — the replacement-window bind race lives inside the upstream `mempalace` package, so no local change can hold the socket across a process replacement | **MemPalace/mempalace#2257** — recorded on #913 in [comment 5294787144](https://github.com/crewrig/crewrig/issues/913#issuecomment-5294787144) before the issue closes, per R5's scenario |

R5's `SHALL` is discharged by publication, not by this diff. Until upstream ships the capability, spec 0139 delta-01's R5 disclosure remains the in-force local treatment and no local proxy component is introduced (spec 0149 → *Out of scope*).

### Review findings from iteration 1

| Finding | Class | Status |
|---|---|---|
| **F1** — the sweep passed vacuously when a swept path did not exist | `tech` (blocking) | **Fixed** in `e38180e`, by construction rather than at the call site. The shipped comment claiming "a mutation cannot satisfy the witness while leaving the repository unguarded" was false as written when the review found it; it is true now, and row 2 of the failure-mode table is its witness |
| **LOW** (security pass) — the guard was `-H`-shaped only, so curl's own credential flags escaped it | `tech` | **Fixed** in `e38180e` — two shapes added to the alternation, each with its own fixture line and its own mutation-locked assertion |
| **F2** — the failure message attributed only a file's first hit to that file | `tech` (non-blocking) | **Fixed** in `e38180e` — per-line prefixing at `:354-359` |
| **F3** — two shipped comments assert a repository-wide property nothing verifies | `spec` (non-blocking) | **Open, accepted.** `class: spec`, not routed to this pass. The claim is true today (the sweep above returns exactly the two named matches) but is unqualified present tense, and nothing in CI reads a comment |
| Recovery advice does not evict the squatter | security | **Routed to #951** by user decision at the REVIEW gate — a behaviour change, not a copy edit |

### Deviation from the approved plan

**One, flagged since iteration 1.** Plan step 2 specified `tr 'A-Z' 'a-z'`; the shipped code at `:318-319` uses `tr '[:upper:]' '[:lower:]'`. Verified rather than asserted: `shellcheck` emits `SC2019`/`SC2018` on the bracket-less form, and neither code appears in this repository's baseline for these three files. Behaviour is identical for the ASCII-only literals being folded.

### Shell-lint delta

This repository runs no `shellcheck` CI job; the available linter is `.github/skills/pr-reviewer/scripts/lint-shell.sh`, which has pre-existing findings on all three touched scripts, so it was run on both sides and compared rather than read as a gate. Counting **every** `SCnnnn (level)` occurrence — including the two-code `^---^` form that a `grep -cE -- '-- SC[0-9]+'` one-liner misses:

- Baseline (the three files at `crewrig/main` = `bfb320e`): **89** findings — `SC1091` ×2, `SC2015` ×77, `SC2016` ×8, `SC2034` ×1, `SC2088` ×1.
- At `9d7ee5a`: **96** findings, same five codes. No new category.
- 8 new, 1 gone; the one that disappeared is the R8 assertion, reworded from `no bearer token is passed through an -H argv flag` to `no credential is passed through an argv flag`, so it is the same finding on new text.
- Net `+1 SC2015` (the hit-count assertion) and `+6 SC2016`, all `(info)`. The `SC2016`s are the five fixture `printf` lines and the shape-C `case` pattern: single quotes are load-bearing there, since the `$t` must reach the fixture unexpanded — it stands in for a token variable in the shape being matched.

(Iteration 1's body reported `83 → 87` against the old merge-base using the narrower `grep -cE` method the reviewer also used; the same method gives `84 → 89` here. The figures above use the fuller count and the current merge-base.)

### Convention triggers — none fires

- **Version bump — not triggered.** `docs/version-bump-convention.md`'s affected-path list covers component sources under `artifacts/**` and `extensions/**`; none of the four touched paths matches, and no `scripts/**` file carries a `provenance.version`.
- **CLI matrix — not triggered.** The trigger surface names `scripts/build-components.sh` and `scripts/{build,install,setup,import,manage}-*.sh`. None of the three touched scripts matches — `test-setup-org-mcp.sh` sits under `scripts/tests/`, not `scripts/setup-*.sh`. The change is one shared helper plus its tests, uniform across all four CLIs, so no parity cell shifts.
- **Test wiring — not triggered.** No test file is added or removed.
- **Exec bits preserved.** `100755` on both test scripts, `100644` on `common.sh` — identical to `crewrig/main`.

### Lifecycle state

SPECS → spec-PR #916 merged. PLAN → posted on #913, cold `architect` review APPROVE with zero findings; both non-blocking observations applied. DEV → `a1a0600`, `87dd7b3`. REVIEW iteration 1 → REQUEST CHANGES (F1 blocking, F2/F3 non-blocking) plus a security pass. DEV iteration 2 → `e38180e`. Branch refreshed onto `main` at `bfb320e` via `9d7ee5a`. REVIEW iteration 2 → next.

This PR carries **no close keyword** for #913: R5 makes the recorded upstream URL a precondition for closing it, and the issue is the ticket's logbook (`docs/logbook-issues.md` → Rule A), closed under Rule C after merge.

**Writer:** `pr-logbook` owns this PR body for ticket #913.

### DIFF
--- a/scripts/lib/common.sh
+++ b/scripts/lib/common.sh
@@ -937,6 +937,7 @@ mcp_daemon_replace_process() {
   echo "           receives the newly minted token in the very next probe — this"
   echo "           one — and can then answer your agents with fabricated memory."
   echo "           Nothing below can tell that process from the real daemon."
+  echo "           Rotate the token again if the window may have been claimed."
   echo ""
 
   case "$(uname -s)" in
--- a/scripts/tests/test-mcp-daemon.sh
+++ b/scripts/tests/test-mcp-daemon.sh
@@ -270,27 +270,85 @@ grep -q "${tok}" "${launcher}" 2>/dev/null \
 # sampling the process table while a probe runs harvests the credential — the
 # reason both probes feed the header through a stdin curl config instead
 # (scripts/lib/common.sh `_mcp_daemon_probe_accepts`, and `_probe_code` in
-# section 16 below). The pattern's last syllable is concatenated at runtime so
-# THIS assertion cannot satisfy the search it performs. It deliberately admits
-# every spelling of the flag rather than the canonical one only: `--header` as
-# well as `-H`, `=` or nothing in place of the space (`--header=`, `-H'…'`,
-# `-HAuthorization:…`), an optional opening quote, and anything at all between
-# the colon and the scheme name — a cold review put all four variants past the
-# earlier `-(H|-header) ` form untouched. Over-matching is the right failure
-# direction here: a false positive costs a reader one glance, a false negative
-# ships the credential back into the process table. Comments are stripped
-# by matching grep -n's own `<line>:` prefix, NOT by leading whitespace —
-# the trap section 4 below documents; the filter is load-bearing here, since
-# common.sh's `register_mempalace_mcp` explains the hazard in prose that
-# quotes the very flag being banned. Scoped to the two files this spec
-# governs: scripts/tests/test-setup-org-mcp.sh asserts the org-MCP `--header`
-# argv shape on purpose, and sweeping that call site is the follow-up
-# specs/0139-token-rotation-revocation.delta-01.md defers.
-argv_bearer="Bea""rer"
-argv_bearer_hits="$(grep -nE -- "-(H|-header)[[:space:]=]*[\"']?Authorization:.*${argv_bearer}" \
+# section 16 below). BOTH halves of the banned string — the header name and
+# the scheme word — are concatenated at runtime (`argv_hdr`, `argv_scheme`) so
+# THIS assertion cannot satisfy the search it performs. Splitting the scheme
+# alone sufficed while the search was case-sensitive; it does not since spec
+# 0149 R1 made the match case-insensitive, because any line in this block that
+# spells the header name out then self-trips the pattern. It deliberately
+# admits every spelling of the flag rather than the canonical one only:
+# `--header` as well as `-H`, `=` or nothing in place of the space
+# (`--header=`, `-H'…'`, `-HAuthorization:…`), an optional opening quote, and
+# anything at all between the colon and the scheme name — a cold review put
+# all four variants past the earlier `-(H|-header) ` form untouched.
+# Over-matching is the right failure direction here: a false positive costs a
+# reader one glance, a false negative ships the credential back into the
+# process table.
+#
+# Comments are stripped by matching grep -n's own `<line>:` prefix ANCHORED
+# (spec 0149 R2), NOT by leading whitespace — the trap section 4 below
+# documents — and NOT unanchored, which used to swallow any genuine violation
+# whose line merely carried a trailing comment. The anchor is why the sweep
+# runs ONE FILE AT A TIME: hand `grep -n` two paths and every hit gains a
+# `<path>:` prefix, so `^[0-9]` never matches and the filter degenerates into
+# excluding nothing. The filter is load-bearing either way, since common.sh's
+# `register_mempalace_mcp` explains the hazard in prose that quotes the very
+# flag being banned.
+#
+# Scope: the two files this spec governs, swept through the same helper as the
+# fixture below — the executable witness that both mutations spec 0149 names
+# are caught and that a pure comment is still excluded. The one deliberate
+# occurrence in the repository, scripts/tests/test-setup-org-mcp.sh, asserts
+# the org-MCP argv shape on purpose and carries a comment saying so at its own
+# call site (spec 0149 R4); no other occurrence exists under scripts/.
+argv_scheme="Bea""rer"
+argv_hdr="Authoriz""ation"
+argv_pat="-(H|-header)[[:space:]=]*[\"']?${argv_hdr}:.*${argv_scheme}"
+# ONE pattern and ONE filter, shared by the self-test and the real sweep, so a
+# mutation cannot satisfy the witness while leaving the repository unguarded.
+_argv_bearer_hits() {
+  local f hits all=""
+  for f in "$@"; do
+    hits="$(grep -niE -- "${argv_pat}" "${f}" 2>/dev/null \
+      | grep -vE '^[0-9]+:[[:space:]]*#' || true)"
+    [ -n "${hits}" ] && all="${all}${f}:${hits}"$'\n'
+  done
+  printf '%s' "${all}"
+}
+
+# Self-test BEFORE the sweep: this guard is a static grep whose passing state
+# looks identical whether or not it can still catch a violation, so it has to
+# prove it can. Three fixture lines under ${TEST_HOME} (the cleanup trap at
+# :102-103 removes them): the lower-case shape R1 names, the trailing-comment
+# shape R2 names, and a pure comment that must stay excluded. The two words
+# are handed to printf at runtime for the same reason the pattern splits them.
+argv_fixture="${TEST_HOME}/argv-guard-fixture.txt"
+argv_hdr_lc="$(printf '%s' "${argv_hdr}" | tr '[:upper:]' '[:lower:]')"
+argv_scheme_lc="$(printf '%s' "${argv_scheme}" | tr '[:upper:]' '[:lower:]')"
+{ printf 'curl -H "%s: %s $t" http://x\n'              "${argv_hdr_lc}" "${argv_scheme_lc}"
+  printf 'curl -H "%s: %s $t" http://x  # ref: #913\n' "${argv_hdr}"    "${argv_scheme}"
+  printf '  # prose: NOT curl -H "%s: %s $t"\n'        "${argv_hdr}"    "${argv_scheme}"
+} > "${argv_fixture}"
+argv_self_hits="$(_argv_bearer_hits "${argv_fixture}")"
+argv_self_n="$(printf '%s' "${argv_self_hits}" | grep -c . || true)"
+case "${argv_self_hits}" in
+  *"${argv_hdr_lc}: ${argv_scheme_lc}"*) ok "the guard catches a violation spelled in lower case (R1)" ;;
+  *) nope "the guard missed the lower-case violation: ${argv_self_hits}" ;;
+esac
+case "${argv_self_hits}" in
+  *"# ref: #913"*) ok "the guard catches a violation carrying a trailing comment (R2)" ;;
+  *) nope "the comment filter swallowed a violation carrying a trailing comment: ${argv_self_hits}" ;;
+esac
+# The assertion that kills the degenerate "exclude nothing" filter, which would
+# satisfy both cases above while making the guard useless: exactly the two
+# violation lines are reported, and the pure comment is not.
+[ "${argv_self_n}" -eq 2 ] \
+  && ok "the anchored filter excludes the comment line and nothing else (R2)" \
+  || nope "expected exactly 2 fixture violations, got ${argv_self_n}: ${argv_self_hits}"
+
+argv_bearer_hits="$(_argv_bearer_hits \
   "${REPO_DIR}/scripts/lib/common.sh" \
-  "${REPO_DIR}/scripts/tests/test-mcp-daemon.sh" 2>/dev/null \
-  | grep -v ':[[:space:]]*#' || true)"
+  "${REPO_DIR}/scripts/tests/test-mcp-daemon.sh")"
 [ -z "${argv_bearer_hits}" ] \
   && ok "no bearer token is passed through an -H argv flag (R8)" \
   || nope "a bearer token reached argv via -H: ${argv_bearer_hits}"
@@ -774,6 +832,14 @@ esac
 case "${out}" in
   *"receives the newly minted token"*) ok "the replacement window is disclosed before the stop is issued (R5)" ;;
   *) nope "no replacement-window disclosure: ${out}" ;;
+esac
+# Second half of the same witness: spec 0149 R3 makes a disclosure that names
+# the risk without naming the recovery non-conformant, so the risk substring
+# above cannot be the only thing asserted. `case` is case-sensitive, so this
+# substring tracks the copy in common.sh verbatim.
+case "${out}" in
+  *"Rotate the token again"*) ok "the disclosure names the recovery, not only the risk (0149 R3)" ;;
+  *) nope "the disclosure names the risk without the recovery action: ${out}" ;;
 esac
 
 # Behavioural half: probed skip on the prerequisites the launcher actually
--- a/scripts/tests/test-mcp-daemon.sh
+++ b/scripts/tests/test-mcp-daemon.sh
@@ -265,10 +265,11 @@ grep -q "${tok}" "${launcher}" 2>/dev/null \
   && nope "the token VALUE was substituted into the launcher" \
   || ok "no token value is baked into the launcher"
 
-# Reintroduction guard: no bearer may travel through an `-H`/`--header` argv
-# flag either. /proc/<pid>/cmdline is world-readable on Linux, so a local uid
-# sampling the process table while a probe runs harvests the credential — the
-# reason both probes feed the header through a stdin curl config instead
+# Reintroduction guard: no credential may travel through an argv flag either —
+# not the hand-built header, and not curl's own credential flags, which end up
+# in the same place. /proc/<pid>/cmdline is world-readable on Linux, so a local
+# uid sampling the process table while a probe runs harvests the credential —
+# the reason both probes feed the header through a stdin curl config instead
 # (scripts/lib/common.sh `_mcp_daemon_probe_accepts`, and `_probe_code` in
 # section 16 below). BOTH halves of the banned string — the header name and
 # the scheme word — are concatenated at runtime (`argv_hdr`, `argv_scheme`) so
@@ -296,38 +297,73 @@ grep -q "${tok}" "${launcher}" 2>/dev/null \
 # flag being banned.
 #
 # Scope: the two files this spec governs, swept through the same helper as the
-# fixture below — the executable witness that both mutations spec 0149 names
-# are caught and that a pure comment is still excluded. The one deliberate
-# occurrence in the repository, scripts/tests/test-setup-org-mcp.sh, asserts
-# the org-MCP argv shape on purpose and carries a comment saying so at its own
-# call site (spec 0149 R4); no other occurrence exists under scripts/.
+# fixture below — the executable witness that every shape the pattern claims to
+# catch is caught, that a pure comment is still excluded, and that the swept
+# paths were actually read. The one deliberate occurrence in the repository,
+# scripts/tests/test-setup-org-mcp.sh, asserts the org-MCP argv shape on
+# purpose and carries a comment saying so at its own call site (spec 0149 R4);
+# no other occurrence exists under scripts/.
 argv_scheme="Bea""rer"
 argv_hdr="Authoriz""ation"
+argv_hdr_lc="$(printf '%s' "${argv_hdr}" | tr '[:upper:]' '[:lower:]')"
+argv_scheme_lc="$(printf '%s' "${argv_scheme}" | tr '[:upper:]' '[:lower:]')"
+# Three shapes, one alternation. The header shape is the one the daemon probes
+# could regress into; the other two are the likelier INNOCENT reintroduction —
+# a contributor reaching for curl's own credential flags, which put the raw
+# token in argv just as plainly as a hand-built header does:
+#   A  -H/--header "<hdr>: <scheme> …"   the header shape
+#   B  --oauth2-bearer <token>           curl sends the same header from it
+#   C  curl … -u/--user "<user>:<pass>"  basic auth, credential in argv
+# C is anchored on `curl` on the same line and on the `user:password` colon,
+# unlike A and B: bare `-u` is far too common to match on its own (`shopt -u`,
+# `systemctl --user`, `date -u '+%H:%M'` — the last would match the colon rule
+# without the anchor). A and B need no anchor: both flags are unambiguous.
 argv_pat="-(H|-header)[[:space:]=]*[\"']?${argv_hdr}:.*${argv_scheme}"
-# ONE pattern and ONE filter, shared by the self-test and the real sweep, so a
-# mutation cannot satisfy the witness while leaving the repository unguarded.
+argv_pat="${argv_pat}|--oauth2-${argv_scheme_lc}[[:space:]=]"
+argv_pat="${argv_pat}|curl.*[[:space:]]-(u|-user)[[:space:]=]+[\"']?[^\"'[:space:]]*:"
+# ONE pattern and ONE filter, shared by the self-test and the real sweep.
+#
+# A path that does not exist is recorded as a violation, not skipped. This is
+# the difference between a guard and a guard-shaped no-op: `grep` on a missing
+# file yields no hits and `2>/dev/null` eats the diagnostic, so renaming either
+# swept file — `git mv`, a refactor — would otherwise disarm the sweep and
+# leave every assertion in this block green while the repository is searched
+# for nothing. Checking inside the helper, on exactly the paths it is about to
+# search, is what makes the property hold BY CONSTRUCTION: a second list at the
+# call site could drift from the first, this cannot.
 _argv_bearer_hits() {
-  local f hits all=""
+  local f line hits all=""
   for f in "$@"; do
+    if [ ! -f "${f}" ]; then
+      all="${all}${f}:0:MISSING — the guard swept a path that does not exist"$'\n'
+      continue
+    fi
     hits="$(grep -niE -- "${argv_pat}" "${f}" 2>/dev/null \
       | grep -vE '^[0-9]+:[[:space:]]*#' || true)"
-    [ -n "${hits}" ] && all="${all}${f}:${hits}"$'\n'
+    [ -n "${hits}" ] || continue
+    # Per line, not per blob: prefixing the file to the whole blob left the
+    # second and later hits of a file carrying grep's bare `<line>:`, so a
+    # reader had to infer block boundaries to attribute them.
+    while IFS= read -r line; do
+      [ -n "${line}" ] && all="${all}${f}:${line}"$'\n'
+    done <<< "${hits}"
   done
   printf '%s' "${all}"
 }
 
 # Self-test BEFORE the sweep: this guard is a static grep whose passing state
 # looks identical whether or not it can still catch a violation, so it has to
-# prove it can. Three fixture lines under ${TEST_HOME} (the cleanup trap at
+# prove it can. Five fixture lines under ${TEST_HOME} (the cleanup trap at
 # :102-103 removes them): the lower-case shape R1 names, the trailing-comment
-# shape R2 names, and a pure comment that must stay excluded. The two words
-# are handed to printf at runtime for the same reason the pattern splits them.
+# shape R2 names, a pure comment that must stay excluded, and one line per
+# credential-flag shape B and C. Every word the pattern searches for is handed
+# to printf at runtime, for the same reason the pattern itself splits them.
 argv_fixture="${TEST_HOME}/argv-guard-fixture.txt"
-argv_hdr_lc="$(printf '%s' "${argv_hdr}" | tr '[:upper:]' '[:lower:]')"
-argv_scheme_lc="$(printf '%s' "${argv_scheme}" | tr '[:upper:]' '[:lower:]')"
 { printf 'curl -H "%s: %s $t" http://x\n'              "${argv_hdr_lc}" "${argv_scheme_lc}"
   printf 'curl -H "%s: %s $t" http://x  # ref: #913\n' "${argv_hdr}"    "${argv_scheme}"
   printf '  # prose: NOT curl -H "%s: %s $t"\n'        "${argv_hdr}"    "${argv_scheme}"
+  printf 'curl --oauth2-%s "$t" http://x\n'            "${argv_scheme_lc}"
+  printf 'curl -sS -u "%s" http://x\n'                 'svc:$t'
 } > "${argv_fixture}"
 argv_self_hits="$(_argv_bearer_hits "${argv_fixture}")"
 argv_self_n="$(printf '%s' "${argv_self_hits}" | grep -c . || true)"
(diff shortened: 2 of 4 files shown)


hcross:
## Verdict: REQUEST CHANGES

Cold review of #948 at `87dd7b3` (`updatedAt=2026-08-14T15:27:06Z`, unchanged across this
review). Every number below was re-derived at the moment of writing; the commands are quoted
so each claim is re-runnable. The diff was audited against `specs/0149-daemon-credential-hardening.md`
read at the PR's head ref, not from the local tree.

One blocking finding, two non-blocking. The blocking one is a residual vacuity in the very
property this PR exists to establish, and it is a one-line fix.

---

### CI status

**30 / 30 checks pass** at `87dd7b3`. The two jobs that were `pending` when this review
opened have since completed:

```
chroma-mcp    pass   3m20s
test-wiring   pass   8m18s
```

`gh pr checks 948 | awk '{print $2}' | sort | uniq -c` → `30 pass`. No failing or pending
required check; the preflight clears and the editorial review below is not moot.
`gh pr view 948 --json closingIssuesReferences` → `[]`, so the merge closes no issue — correct
per R5, which makes the recorded upstream URL a precondition for closing #913.

---

### Correctness

**Reproduced locally at `87dd7b3`** (scratch tree from `git archive 87dd7b3`, so nothing was
run against a live worktree):

| Command | Result | PR-body claim |
|---|---|---|
| `bash scripts/tests/test-mcp-daemon.sh` | exit 0 — `passed: 77   failed: 0` | matches |
| `bash scripts/tests/test-setup-org-mcp.sh` | exit 0 — `RESULT: 41 passed, 0 failed` | matches |
| `bash scripts/check-bash32-portability.sh` | exit 0 — `3 declared construct(s), 168 file(s) scanned, 155 .sh file(s) array-scanned` | matches |
| R4 sweep (§3 of *How to test*) | exactly 2 matches: `test-setup-org-mcp.sh:208`, `common.sh:1040` | matches |

**The failure-mode table reproduces in full — all four rows.** Each mutation was applied to a
copy of the head tree and the whole suite re-run:

| Mutation | Suite | Assertions that redden |
|---|---|---|
| `:312` `grep -niE` → `grep -nE` | `75 / 2` | R1 lower-case; count (`got 1`) |
| `:313` filter → unanchored `':[[:space:]]*#'` | `75 / 2` | R2 trailing-comment; count (`got 1`) |
| `:313` filter → `\| cat` (degenerate "exclude nothing") | `75 / 2` | count (`got 3`); real sweep — `a bearer token reached argv via -H: …/scripts/lib/common.sh:1040` |

The third row is the load-bearing one and it behaves exactly as the body claims: the degenerate
filter satisfies both named requirements and is killed only by the hit-count assertion at
`:345-347` and by the real sweep. The fixture at `:328-330` is well built for this — line 2's
`# ref: #913` contains the `: #` sequence that the unanchored filter matches, which is precisely
why row 2 reddens.

**Beyond what the body claims, I injected a genuine reintroduction** — a non-comment
`curl -sS -H "Authorization: Bearer ${token}"` line with a trailing comment, at
`scripts/lib/common.sh:848` inside `_mcp_daemon_probe_accepts()`. The suite went to
`76 / 1`, exit 1:

```
  FAIL a bearer token reached argv via -H: …/scripts/lib/common.sh:848:  curl -sS -H "Authorization: Bearer ${token}" … # reintroduced
```

So the guard works end-to-end on the real swept files, not only on its fixture. The per-file
iteration at `:309-317` is genuinely load-bearing: the `^[0-9]+:` anchor cannot match a
multi-path `grep -n` whose hits carry a `<path>:` prefix, and the block's own comment at
`:288-294` states that correctly.

**R3** — `scripts/lib/common.sh:940` sits inside the existing unconditional stdout block
(`:934-941`), in the surrounding WARNING's voice and at its column, and `:841` asserts the
recovery half verbatim (`case` is case-sensitive, so it tracks the copy). **R5** — verified
discharged by publication: comment `5294787144` on #913 records
`https://github.com/MemPalace/mempalace/issues/2257`, and #913 is `OPEN`, so the URL is on
record before the issue closes, as the R5 scenario requires.

**Line anchors.** All 20 anchors cited in *How to read this PR?* are accurate at `87dd7b3`
(`:248`, `:268`, `:303-306`, `:309`, `:317`, `:319`, `:325`, `:331`, `:334-341`, `:345-347`,
`:349`, `:354`, `common.sh:940`, `:833`, `:841`, `test-setup-org-mcp.sh:199-208`,
`common.sh:1040`). So are both line references shipped **inside** the new comments: the cleanup
trap is at `test-mcp-daemon.sh:102-103` and the `ORG_HTTP` fixture is at
`test-setup-org-mcp.sh:58`, whose token is indeed the literal `${ATLASSIAN_TOKEN}`.

---

#### F1 — the sweep still passes vacuously when a swept path does not exist (blocking)

`class: tech` · `scripts/tests/test-mcp-daemon.sh:309-317`, `:349-354`

`_argv_bearer_hits` runs `grep -niE -- "${argv_pat}" "${f}" 2>/dev/null` (`:312`). The
`2>/dev/null` swallows `No such file or directory`, so a swept path that no longer exists
contributes no hits and the R8 assertion at `:352-354` reports clean. I mutated **both** paths
at `:350-351` to non-existent names (`comon.sh`, `test-mcp-daemon-renamed.sh`) and re-ran the
whole suite:

```
  ok   the guard catches a violation spelled in lower case (R1)
  ok   the guard catches a violation carrying a trailing comment (R2)
  ok   the anchored filter excludes the comment line and nothing else (R2)
  ok   no bearer token is passed through an -H argv flag (R8)
  passed: 77   failed: 0        (exit 0)
```

That is **byte-identical to the shipped green state** while the repository is swept for nothing
at all. It also falsifies, as written, the claim shipped in the code at `:307-308` and repeated
in the PR body:

> ONE pattern and ONE filter, shared by the self-test and the real sweep, so a mutation cannot
> satisfy the witness while leaving the repository unguarded.

The mutation above satisfies every witness — including the new hit-count assertion — and leaves
the repository unguarded. The self-test closes the *helper-logic* half of the "green whether or
not it can catch anything" problem; the *swept-target* half stays open, and a plain rename of
either file (`git mv`, a future refactor) disarms the guard with no signal.

I am flagging this as blocking rather than as a nit for two reasons: this vacuity class is
exactly what spec 0149 exists to eliminate — a static grep guard whose passing state carries no
information — and the diff ships a comment asserting the property it does not have.

**Verified fix** (not asserted — I ran it). Assert the swept paths at the call site, outside the
command substitution, immediately before `:349`:

```sh
for argv_f in "${REPO_DIR}/scripts/lib/common.sh" \
              "${REPO_DIR}/scripts/tests/test-mcp-daemon.sh"; do
  [ -f "${argv_f}" ] \
    && ok "the swept path exists: ${argv_f##*/}" \
    || nope "the guard sweeps a path that does not exist: ${argv_f}"
done
```

Against the two-typo mutant that shipped green above, this gives `passed: 77   failed: 2`,
exit 1, with the R8 diagnostic left intact. Putting the same `[ -f … ] || nope …` *inside* the
helper loop also reddens, but `nope` writes to stdout and the helper runs in a command
substitution, so its output is captured into `argv_bearer_hits` and garbles the R8 failure
message — hence the call-site placement.

Narrowing the `:307-308` comment instead of adding the assertion would also resolve the
inaccuracy, but it leaves the guard disarmable by a rename, so the assertion is the better
trade at one line.

---

### Convention compliance

Verified against the rules as written, not from memory:

- **Spec status transition** — `docs/spec-format.md:299-313` (*Recording the `approved` →
  `implemented` transition*) requires the edit to ride the implementation PR's own **new
  commit** and to touch "no body line, no `id`, no `slug`, no `version`". Commit `87dd7b3` is
  exactly that: `1 file changed, 1 insertion(+), 1 deletion(-)`, `status: approved` →
  `status: implemented`. Conformant.
- **CLI matrix — correctly not triggered.** The authoritative trigger surface is
  `docs/cli-matrix-maintenance.md:11-16`: `scripts/build-components.sh` and
  `scripts/{build,install,setup,import,manage}-*.sh`. None of the four touched paths matches —
  `scripts/lib/common.sh` is a shared library, and `scripts/tests/test-setup-org-mcp.sh` lives
  under `scripts/tests/`, not `scripts/setup-*.sh`. The body's reasoning holds.
- **Version bump — correctly not triggered.** No path under `artifacts/**` or `extensions/**`;
  no touched file carries a `provenance.version`.
- **Gitmoji** — PR title `🔐 …` and both commits (`🔐 …`, `🔖 …`) conform; no
  conventional-commit prefix anywhere.
- **Branch naming** — `feat/0913-daemon-credential-hardening` matches `<prefix>/<NNNN>-<slug>`.
- **PR format** — all four required sections present per `docs/pr-format.md:18,23,28`.
- **Language** — English throughout, including the new in-file comments.
- **Logbook** — #913 is the ticket's own logbook (Rule A) and stays open; no close keyword.

#### F2 — the guard's failure message attributes only its first line to a file (non-blocking)

`class: tech` · `scripts/tests/test-mcp-daemon.sh:314`

`all="${all}${f}:${hits}"$'\n'` prefixes `${f}:` to the **blob**, so the second and later
violation lines of the same file carry only `grep -n`'s bare `<line>:`. Visible in the injected
run above and harmless with one hit; with hits in both swept files a reader has to infer block
boundaries to attribute the unprefixed lines. `printf '%s:%s\n' "${f}" "${line}"` per line
would remove the inference. Diagnostic quality only — the pass/fail decision is unaffected.

#### F3 — two shipped comments assert a repository-wide property nothing verifies (non-blocking)

`class: spec` · `scripts/tests/test-setup-org-mcp.sh:199`, `scripts/tests/test-mcp-daemon.sh:303`

Both claim the marked site is the only bearer-in-argv occurrence under `scripts/` ("and the only
one under scripts/", "no other occurrence exists under scripts/"). **True today** — my sweep
returns exactly the two named matches — and the executable guard covers only `common.sh` and
`test-mcp-daemon.sh`, because spec 0149 → *Out of scope* deliberately declined a repo-wide CI
lint. That trade is the spec's to make and I am not relitigating it; the observation is about the
wording: an unqualified present-tense claim about all of `scripts/` will silently rot as the tree
grows, and nothing in CI reads a comment. Scoping the phrasing ("the only occurrence as of spec
0149") or extending the sweep to `scripts/` with this one site excluded would both close it.

---

### Test coverage

Adequate and, unusually, **demonstrably non-vacuous** — the three new assertions at `:334-347`
were each shown to redden under a targeted mutation (table above), which is the standard this
guard failed before. Every requirement that admits an executable witness has one:

- **R1** → `:334-337`, fixture line 1 (`authorization: bearer`), killed by dropping `-i`.
- **R2** → `:338-341` plus the hit-count assertion at `:345-347`, killed by unanchoring the
  filter and by removing it.
- **R3** → `:841`, paired with the pre-existing risk half at `:833`.
- **R4** → no in-suite witness; verified by the manual sweep (2 matches, as claimed). See F3.
- **R5** → discharged by publication; verified on #913, comment `5294787144`.

The one coverage hole is F1: no assertion proves the sweep is pointed at files that exist.

The plan deviation the body flags (`tr '[:upper:]' '[:lower:]'` instead of `tr 'A-Z' 'a-z'`, at
`:326-327`) is the right call and introduces no new finding category — `SC2018`/`SC2019` are
absent from the head-side linter output, confirming the substituted form is the one shellcheck
asks for.

---

### Linter findings

`.claude/skills/pr-reviewer/scripts/lint-markdown.sh specs/0149-daemon-credential-hardening.md`
→ **exit 0**, clean.

`.claude/skills/pr-reviewer/scripts/lint-shell.sh scripts/lib/common.sh scripts/tests/test-mcp-daemon.sh scripts/tests/test-setup-org-mcp.sh`
→ exit 1, with pre-existing findings on all three files, so it was run on both sides and
compared. `shellcheck` was present; no script degraded.

Counting findings as `grep -cE '\-\- SC[0-9]+'`, the body's figures reproduce exactly:
**83 at the merge-base (`e0a60df`) → 87 at `87dd7b3`**, same five codes
(`SC1091`, `SC2015`, `SC2016`, `SC2034`, `SC2088`), none disappeared. The delta is `+1 SC2015`,
`+3 SC2016`, all `(info)`, quoted verbatim:

```
In scripts/tests/test-mcp-daemon.sh line 328:
{ printf 'curl -H "%s: %s $t" http://x\n'              "${argv_hdr_lc}" "${argv_scheme_lc}"
         ^-- SC2016 (info): Expressions don't expand in single quotes, use double quotes for that.

In scripts/tests/test-mcp-daemon.sh line 329:
  printf 'curl -H "%s: %s $t" http://x  # ref: #913\n' "${argv_hdr}"    "${argv_scheme}"
         ^-- SC2016 (info): Expressions don't expand in single quotes, use double quotes for that.

In scripts/tests/test-mcp-daemon.sh line 330:
  printf '  # prose: NOT curl -H "%s: %s $t"\n'        "${argv_hdr}"    "${argv_scheme}"
         ^-- SC2016 (info): Expressions don't expand in single quotes, use double quotes for that.

In scripts/tests/test-mcp-daemon.sh line 346:
  && ok "the anchored filter excludes the comment line and nothing else (R2)" \
  ^-- SC2015 (info): Note that A && B || C is not if-then-else. C may run when A is true.
```

All four are correctly characterised in the body and none is actionable: the single quotes at
`:328-330` are load-bearing (`$t` must reach the fixture unexpanded, since it stands in for the
token variable in the shape being matched), and the `A && B || C` at `:346` is the shape this
suite uses throughout. The pre-existing `scripts/lib/common.sh: missing 'set -e'` note is
present on both sides.

---

### Summary

Careful, well-witnessed work: every figure in the PR body reproduced, all 20 line anchors are
accurate, the mutation table holds in full, and the guard catches a genuine reintroduction I
injected myself. **F1 blocks**: the sweep can be silently disarmed by renaming either swept
file — the suite stays at `77 / 0`, exit 0 — and the diff ships a comment claiming that cannot
happen. One `for` loop at the `:349` call site closes it, verified. F2 and F3 are non-blocking
and can ride the same commit.

*Reviewed cold by `pr-reviewer` — no authoring context. Head `87dd7b3`, `updatedAt` unchanged
at post time.*

hcross:
## Verdict: APPROVE

Cold-start review, iteration 2, at head `9d7ee5a`. Zero blocking findings. The
iteration-1 blocker (F1) and the security LOW are both closed **and
mutation-locked**; I reproduced every row of the body's failure-mode table
independently and added two mutations of my own that the body does not claim.

### CI status

All **30/30** required checks pass at `9d7ee5a`
(`gh pr checks 948`); `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.
No failing or pending job. Notable passes for this diff: `test-wiring` (8m7s),
`lint-specs`, `check-skill-versions`, `check-extension-version-bump`,
`ci-parity`, `check-ci-parity`.

### Correctness

**F1 — the vacuous sweep. Resolved by construction.** The path-existence check
lives inside `_argv_bearer_hits` (`scripts/tests/test-mcp-daemon.sh:347-350`),
on exactly the paths it is about to search, and records a missing path as a
violation-shaped `<path>:0:MISSING` entry. The call site asserts the absence of
that marker at `:409-412` *before and separately from* the sweep verdict at
`:413-415`. Placing the check inside the helper rather than in a second list at
the call site is the right call: a second list is a new thing that can drift
from the first, and this one cannot.

I mutation-tested it rather than reading it. Renaming the swept path at `:404`
to a nonexistent file:

```
FAIL the sweep is vacuous — a swept path does not exist:
     .../scripts/lib/common-RENAMED.sh:0:MISSING — the guard swept a path that does not exist
FAIL a credential reached argv:
     .../scripts/lib/common-RENAMED.sh:0:MISSING — the guard swept a path that does not exist
  passed: 79   failed: 2     (exit 1)
```

Two assertions redden, not one — the marker fails the existence check *and*
makes the sweep non-empty. The body states this same mutation left iteration 1
at `77 / 0`, exit 0. The hole is closed.

**The security LOW — the guard was `-H`-shaped only. Resolved.** Three shapes
now share one alternation (`:331-333`), and each is independently
mutation-locked by two assertions (its own `case` plus the hit-count):

| Mutation | Result | Assertions reddened |
|---|---|---|
| `:332` shape B removed | `79 / 2`, exit 1 | `shape B`; hit count (`got 3`) |
| `:333` shape C removed | `79 / 2`, exit 1 | `shape C`; hit count (`got 3`) |
| `:351` `grep -niE` → `-nE` | `79 / 2`, exit 1 | R1 lower-case; hit count (`got 3`) |
| `:333` shape C `curl` anchor removed | `81 / 0`, exit 0 | none |

Every row matches the body's table exactly, including the last one — the body
is candid that the anchor is defensive rather than currently load-bearing, and
that candour is verified, not asserted.

**Two mutations of my own, beyond the body's table.**

1. *The degenerate "exclude-nothing" filter.* Replacing the comment filter at
   `:352` with a passthrough gives `79 / 2`, exit 1 — killed twice, by the
   hit-count assertion (`got 5`) and by the real sweep catching
   `scripts/lib/common.sh:1040`, the prose comment that quotes the banned flag.
   The body's claim that the hit-count assertion kills this degenerate case
   holds.

2. *A genuine reintroduction into a really-swept file.* The fixture self-test
   proves the **helper** works; it does not by itself prove the **sweep** is
   wired to real files. So I appended a real violation to `scripts/lib/common.sh`
   (as a never-called function body, so it does not execute at source time):

   ```
   FAIL a credential reached argv:
        .../scripts/lib/common.sh:2183:_unused_regression_probe() { curl -H "Authorization: Bearer ${tok}…
     passed: 80   failed: 1     (exit 1)
   ```

   Caught, with the path-existence assertion correctly staying green. The sweep
   reads the real files and catches a live violation, not only fixtures.

**Shape coverage against real violation shapes.** I checked the shipped
alternation against six spellings — quoted and unquoted, `-u` and `--user`,
spaced and unspaced header — and all six match under `grep -niE`. No gap
between what the fixture asserts and what a contributor would plausibly write.

### Convention compliance

- **Title** — `🔐 Harden the anti-argv guard …`: Gitmoji, no conventional-commit
  prefix (`AGENTS.md` → *Naming Convention*). All four commit subjects likewise
  (`🔐`, `🔖`, `✅`, plus the merge commit).
- **Branch** — `feat/0913-daemon-credential-hardening`, matching
  `<prefix>/<NNNN>-<slug>`.
- **Spec status transition** — `approved` → `implemented` recorded in the
  implementation PR's own commit (`87dd7b3`), which is exactly what
  `docs/spec-format.md` → *Recording a status transition* prescribes for this
  transition; `id`/`slug`/body untouched.
- **Version bump — correctly not triggered.** `docs/version-bump-convention.md`
  lists only `artifacts/**` and `extensions/**` `SKILL.md`/`AGENT.md` paths.
  None of the four touched paths matches. (`common.sh` mentions `provenance` ten
  times, but all in code that *reads other components'* provenance —
  `_antigravity_has_provenance` — it carries none of its own.)
- **CLI matrix — correctly not triggered.** The glob is `scripts/setup-*.sh`;
  `scripts/tests/test-setup-org-mcp.sh` sits under `scripts/tests/`. Neither
  `scripts/lib/common.sh` nor the test files match any trigger path.
- **Exec bits preserved** — `100644` `common.sh`, `100755` both test scripts,
  identical to `main` (`git ls-tree`).
- **Close keywords** — `closingIssuesReferences` is `[]`. Correct: R5 makes the
  upstream URL a precondition for closing #913, and #913 is the logbook (Rule A),
  closed under Rule C after merge.
- **Logbook** — #913 open, updated incrementally at each stage (DEV, REVIEW
  iteration 1, DEV iteration 2), satisfying Rule B.
- **R5 traceability** — MemPalace/mempalace#2257 is recorded on #913 in
  comment 5294787144, verified present.
- **#951** — open, titled for evicting the squatter. The R3 recovery-sufficiency
  concern is genuinely tracked, and correctly out of scope here.

### Test coverage

Both suites pass at head, matching the body exactly:

- `scripts/tests/test-mcp-daemon.sh` — `passed: 81   failed: 0`, exit 0.
- `scripts/tests/test-setup-org-mcp.sh` — `RESULT: 41 passed, 0 failed`, exit 0.

All seven §3 assertions the body enumerates are present, in the listed order,
plus the §16 R3 witness (`the disclosure names the recovery, not only the risk`)
paired with the pre-existing risk witness at `:894`. The new §16 assertion
tracks `common.sh:940` verbatim, and `case` is case-sensitive, so it cannot
drift silently.

The fixture (`:371-377`) is written under `${TEST_HOME}` and removed by the
existing `cleanup()` trap at `:102-103` — verified, no stray files.

### Linter findings

- **`lint-shell.sh`** — exit 1 on all three scripts, all pre-existing. Counting
  every `SCnnnn (level)` occurrence on both sides:

  - Baseline (three files at `main`): **89** — `SC2015` ×77, `SC2016` ×8,
    `SC1091` ×2, `SC2034` ×1, `SC2088` ×1.
  - At `9d7ee5a`: **96** — `SC2015` ×78, `SC2016` ×14, `SC1091` ×2,
    `SC2034` ×1, `SC2088` ×1.

  Net `+1 SC2015`, `+6 SC2016`, all `(info)`, no new category. The body's
  figures and per-code breakdown reproduce **exactly**. `SC2019`/`SC2018` appear
  zero times on both sides, which is the evidence the body offers for the
  `tr '[:upper:]' '[:lower:]'` plan deviation — confirmed.

- **`lint-markdown.sh`** on the spec — exit 0, clean.
- `task spec:lint` — `Linting passed!`, 212 files.
- `scripts/check-bash32-portability.sh` — exit 0, `171 file(s) scanned,
  157 .sh file(s) array-scanned`. The new helper uses only Bash 3.2-safe
  constructs (`local`, `<<<`, `$'\n'`).

**Body-vs-code audit.** I spot-checked ~20 line anchors against the shipped file
at `9d7ee5a` — `:102-103`, `:258`, `:316-319`, `:331-333`, `:344-362`,
`:347-350`, `:351`, `:352`, `:354-359`, `:371-377`, `:380-383`, `:384-387`,
`:399-401`, `:404`, `:409-412`, `:413-415`, `:894`, `:902`, `common.sh:940`,
`common.sh:1040`, `test-setup-org-mcp.sh:58` and `:199-208`. **Every one is
exact.** §3 does span `:258-415` (section 4 opens at `:417`). The R4 sweep
returns exactly two matches, exactly the two named. The `+10` re-derivation
after the refresh was done correctly.

### Non-blocking observations

**N1 — `[ ! -f ]` admits an unreadable file.** `class: tech`

`scripts/tests/test-mcp-daemon.sh:347` tests existence, not readability. A file
that exists but cannot be read passes `[ -f ]`, and `grep` then exits 2 with no
output — which `2>/dev/null` and `|| true` both swallow, reproducing precisely
the vacuous green F1 was opened for. Verified directly:

```
[ -f secret.sh ] => TRUE (passes existence check)
grep exit=2, no output
```

Not reachable from a normal git checkout — git never produces a mode-000 file,
so this needs a deliberate `chmod`, unlike the rename/refactor vector that F1
was actually about. That vector is closed and locked, so this does not block.
`[ ! -r "${f}" ]` would close both cases in one token (it is false for missing
*and* unreadable); capturing `grep`'s exit 2 as a violation would be the fuller
form. Worth a later pass, not this one.

**N2 — F3 stands, as accepted.** `class: spec`

The "only occurrence under `scripts/`" claims in
`test-setup-org-mcp.sh:199-207` and the §3 header are true today — I re-ran the
sweep and got exactly the two named matches — but they are unqualified present
tense and nothing in CI reads a comment. Already accepted as non-blocking and
correctly not routed to this pass. Noted, not re-raised.

**N3 — advisory for the merge step, not a finding.** `main` advanced to
`3b6f932` (#949, `test-wiring` optimization) after this body was written, so the
body's "up to date with `main` at `bfb320e`" is now one commit stale. No
re-verification is needed: `git diff bfb320e..main` over the three touched
scripts is empty, so every measurement above stands unchanged. GitHub reported
`MERGEABLE` / `CLEAN` throughout this review and, at the moment of posting, is
recomputing (`UNKNOWN`) because #949 has just landed — expected, transient, and
not a signal about this branch. Per `AGENTS.md`, the up-to-date precondition is
verified immediately before `gh pr merge` regardless; re-check it there.

**N4 — reproduction hazard for whoever re-runs the §3 sweep locally.** Not a
defect in this PR. If your `grep` is `ugrep` (a common drop-in on macOS), the
body's sweep command returns **zero** matches under `-i`, because ugrep 7.5.0
does not match `[^"'[:space:]]*` the way GNU and BSD grep do. I hit this and
nearly filed a false finding against the body's false-positive count. Under real
BSD grep the body is right on the nose: the unanchored shape C yields exactly
one false positive, `scripts/worktree-claim.sh:493`
(`now_iso() { date -u +%Y-%m-%dT%H:%M:%SZ; }`), and the anchored form yields
none. Use `/usr/bin/grep` to reproduce. The shipped guard is unaffected — CI runs
GNU grep, and all six real violation shapes match under both real greps.

### Summary

The change does what it claims, and — unusually — the PR body's every checkable
number is correct: `81/0`, `41/0`, all six failure-mode rows, `89 → 96` with the
per-code split, `212` files, the R4 sweep count, the anchor false-positive
count, and ~20 line anchors re-derived after a `+10` shift. The one claim I
could not reproduce turned out to be my own toolchain lying to me, not the body.
The guard now fails loudly in every way it is supposed to, and I confirmed that
by breaking it seven different ways rather than by reading it.

Approving. N1 is worth a cheap follow-up; N2 is already tracked; N3 is for the
merger.

---

*Cold-start independent review. Verdict posted as a plain comment: the
authenticated identity (`hcross`) is the PR author, so GitHub rejects a formal
review event on one's own PR (`pr-reviewer` skill → step 6, rung 2). This
comment is the binding verdict artifact.*

Issue #745, opened 2026-08-08
🔍 Verify the MCP daemon listener's owner in status-mcp-server.sh


hcross:
The daemon binds a fixed loopback port. Nothing authenticates the daemon to its clients, so any local process that binds that port first becomes the memory server for all four assistants — and receives the real bearer token in the first request header. It can then answer with fabricated memory into the context of agents holding Bash and git push.

Two windows exist: at boot, since launchd has no ordering between user agents; and after `uninstall-mcp-daemon.sh`, which frees the port while the registrations still point at it. The uninstall now warns explicitly, which covers the operator-error path.

What is not covered is detection: `status-mcp-server.sh` cannot tell a squatter from a healthy daemon, since a squatter answering 200 on /healthz and 401 on an unauthenticated POST is indistinguishable. Verifying the listener's pid against the supervisor's would close that.

Raised by the security review of #742 as MED, explicitly not a merge blocker: it defends against someone who already has local code execution.

Refs #739, ADR 0016.

hcross:
## PLAN — issue #745 (spec 0158)

### Approach

The status probe `scripts/status-mcp-server.sh` gains a fourth check: identify the PID of the process listening on the MCP port and compare it against the PID the OS supervisor (launchd `com.mempalace.mcp-server` on macOS, systemd `mempalace-mcp-server.service` on Linux) reports for the daemon unit. A mismatch is reported as a usurped listener, naming both PIDs and the recovery action (rotate the bearer token), with a non-zero exit. When either PID cannot be determined, the owner is reported as unverifiable with a non-zero exit. The expected PID is overridable via environment so the hermetic suite can fix it without a real supervisor.

### Steps

1. Edit `scripts/lib/common.sh` — add `mcp_listener_pid <port>` (lsof on macOS, ss on Linux) and `mcp_supervisor_pid` (launchctl print `gui/<uid>/<label>` on macOS, `systemctl --user show -p MainPID` on Linux), honouring the existing `MEMPALACE_MCP_LABEL` / `MEMPALACE_MCP_UNIT` overrides and a new `MEMPALACE_MCP_EXPECTED_PID` test override.
2. Edit `scripts/status-mcp-server.sh` — add an owner-check section after the auth check: compare listener PID vs supervisor PID; on mismatch report a usurped listener naming both PIDs plus the token-rotation recovery (mirroring the uninstall disclosure), on undeterminable PID report the owner as unverifiable; set `rc=1` in both cases.
[...]

PR #952, opened 2026-08-14
✨ Verify the MCP daemon listener's owner (spec 0158)

hcross:
The status probe `scripts/status-mcp-server.sh` now verifies the process listening on the MCP port against the process the OS supervisor runs for the daemon, reporting a usurped listener or an unverifiable owner with a non-zero exit instead of silent health. The hermetic suite exercises every served-port branch with the branch under test as the only difference from a healthy run.

## How to read this PR?

Read in this order:

1. `scripts/lib/common.sh:736-793` — the two new helpers. `mcp_listener_pid <port>` resolves the PID listening on the port (lsof on macOS, ss on Linux); `mcp_supervisor_pid` resolves the PID the OS supervisor reports for the daemon unit (launchctl print `gui/<uid>/<label>` on macOS, `systemctl --user show -p MainPID` on Linux). Both honour the existing `MEMPALACE_MCP_LABEL` / `MEMPALACE_MCP_UNIT` overrides and add two test overrides, `MEMPALACE_MCP_EXPECTED_PID` and `MEMPALACE_MCP_LISTENER_PID`. The set-but-empty contract (`${VAR+x}`) is load-bearing: an explicitly empty value forces the lookup to report undeterminable, while an unset one falls through to the real lookup.
2. `scripts/status-mcp-server.sh:72-96` — the new section 3 "Listener owner (spec 0158)". It runs only in the serving state (`rc -eq 0`, after liveness and auth) and BEFORE the launcher drift check, so a drifted launcher cannot hide a usurped listener. Three outcomes: VERIFIED (exit unchanged), USURPED (names both PIDs plus the token-rotation recovery, `rc=1`), UNVERIFIABLE (either side undeterminable, `rc=1`).
3. `scripts/tests/test-mcp-daemon.sh` — section 13d, four hermetic cases fixing both sides via the overrides.

## How to test this PR?

Run the hermetic suite:

```sh
```

Expected: `passed: 84   failed: 0`, exit 0. Section 13d covers the four served-port branches: healthy match (exit 0, VERIFIED), usurped mismatch (non-zero, USURPED, both PIDs named, recovery action named), unverifiable-no-supervisor (non-zero, UNVERIFIABLE), unverifiable-no-listener (non-zero, UNVERIFIABLE, forced by an empty `MEMPALACE_MCP_LISTENER_PID`).

Failure mode to try: with a serving daemon on the MCP port, run the probe with `MEMPALACE_MCP_EXPECTED_PID=1234 MEMPALACE_MCP_LISTENER_PID=5678 bash scripts/status-mcp-server.sh` — it must exit non-zero and report `*** USURPED LISTENER ***` naming both PIDs.

- **Owner check (spec 0158 R1, R4).** When the probe is in the serving state, it compares the listener PID against the supervisor PID. A mismatch reports a usurped listener naming both PIDs and the token-rotation recovery (`rm -f $(mcp_token_path)`, then re-run `bash scripts/switch-mempalace-http.sh`), mirroring the uninstall disclosure (R5). When either side cannot be determined, the owner is reported as unverifiable. Both cases exit non-zero.
- **Expected process from the supervisor, not a file (R2).** The expected PID comes from launchd/systemd, never from a file a same-uid local process could write.
- **Hermetic reachability (R3).** `MEMPALACE_MCP_EXPECTED_PID` pins the supervisor side, `MEMPALACE_MCP_LISTENER_PID` pins the listener side; set-but-empty forces undeterminable. Each served-port branch differs from a healthy run by exactly one override.
- **v2-F1 (launcher-sync precondition).** Section 13d re-installs the launcher first because section 13c left its hash corrupted to `deadbeef`; the healthy-match case asserts exit 0, which requires every probe check to pass, including the drift check.
- **v2-F2 (downstream consumer).** `scripts/switch-mempalace-http.sh:62` consumes the probe's exit code as the transaction's final gate. A usurped or unverifiable owner now fails that transaction after registration — spec-aligned and desirable, since the switch installs the supervisor unit first and the owner check passes in the normal flow.
- **Spec status transition.** The PR also flips spec 0158 `status: approved` → `status: implemented` (metadata-only, per `docs/spec-format.md` → *Recording a status transition*), so the spec lands on `main` already carrying `implemented` at the moment its implementation lands.

### DIFF
--- a/scripts/lib/common.sh
+++ b/scripts/lib/common.sh
@@ -725,6 +725,72 @@ MCP_DAEMON_PORT_DEFAULT="41893"
 MCP_DAEMON_LABEL_DEFAULT="com.mempalace.mcp-server"
 MCP_DAEMON_UNIT_DEFAULT="mempalace-mcp-server"
 
+# mcp_listener_pid <port> — the PID of the process listening on <port>, or
+# empty when none can be identified. lsof on macOS, ss on Linux.
+#
+# MEMPALACE_MCP_LISTENER_PID fixes the listener side for a hermetic suite
+# (spec 0158 R3). The set-but-empty contract is load-bearing: an explicitly
+# empty value forces the lookup to report undeterminable, while an unset one
+# falls through to the real lookup. `${VAR+x}` is what distinguishes the two —
+# `:-` would collapse both to the fallback.
+mcp_listener_pid() {
+  local port="$1"
+  if [ -n "${MEMPALACE_MCP_LISTENER_PID+x}" ]; then
+    printf '%s\n' "${MEMPALACE_MCP_LISTENER_PID}"
+    return 0
+  fi
+  local pid=""
+  case "$(uname -s)" in
+    Darwin)
+      pid="$(lsof -nP -iTCP:"${port}" -sTCP:LISTEN -t 2>/dev/null | head -n1)"
+      ;;
+    Linux)
+      pid="$(ss -H -tlnp "sport = :${port}" 2>/dev/null \
+        | sed -n 's/.*pid=\([0-9][0-9]*\).*/\1/p' | head -n1)"
+      ;;
+    *)
+      pid=""
+      ;;
+  esac
+  printf '%s\n' "${pid}"
+}
+
+# mcp_supervisor_pid — the PID the OS supervisor reports for the daemon unit,
+# or empty when the unit is not loaded or reports no running process. The
+# expected process comes from the OS supervisor — launchctl print
+# gui/<uid>/<label> on macOS, systemctl --user show -p MainPID on Linux — never
+# from a file a same-uid local process could write (spec 0158 R2).
+#
+# MEMPALACE_MCP_EXPECTED_PID fixes the supervisor side for a hermetic suite
+# (spec 0158 R3); set-but-empty forces undeterminable, unset falls through to
+# the real supervisor lookup.
+mcp_supervisor_pid() {
+  if [ -n "${MEMPALACE_MCP_EXPECTED_PID+x}" ]; then
+    printf '%s\n' "${MEMPALACE_MCP_EXPECTED_PID}"
+    return 0
+  fi
+  local label unit pid=""
+  label="${MEMPALACE_MCP_LABEL:-$MCP_DAEMON_LABEL_DEFAULT}"
+  unit="${MEMPALACE_MCP_UNIT:-$MCP_DAEMON_UNIT_DEFAULT}"
+  case "$(uname -s)" in
+    Darwin)
+      pid="$(launchctl print "gui/$(id -u)/${label}" 2>/dev/null \
+        | sed -n 's/^[[:space:]]*pid = \([0-9][0-9]*\)[[:space:]]*$/\1/p' \
+        | head -n1)"
+      ;;
+    Linux)
+      pid="$(systemctl --user show -p MainPID --value "${unit}" 2>/dev/null)"
+      # systemctl reports MainPID=0 when the unit is loaded but no process
+      # runs; that is undeterminable, not a PID.
+      [ "${pid}" = "0" ] && pid=""
+      ;;
+    *)
+      pid=""
+      ;;
+  esac
+  printf '%s\n' "${pid}"
+}
+
 mcp_launcher_installed_path() {
   printf '%s\n' "${MEMPALACE_MCP_LAUNCHER_PATH:-$HOME/.crewrig/mcp-daemon-launcher.sh}"
 }
--- a/scripts/status-mcp-server.sh
+++ b/scripts/status-mcp-server.sh
@@ -69,7 +69,33 @@ if [ "${rc}" -eq 0 ]; then
   fi
 fi
 
-# --- 3. Launcher drift -------------------------------------------------------
+# --- 3. Listener owner (spec 0158) -------------------------------------------
+# Only meaningful while serving: a dead daemon has no listener to verify, and
+# the NOT SERVING verdict already carries the failure. Runs BEFORE the launcher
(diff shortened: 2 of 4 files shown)


hcross:
## Verdict: APPROVE

### CI status

All required checks pass. Three non-required checks are pending (`chroma-mcp`, `changeset-coverage`, `test-wiring`); `mergeStateStatus` is `UNSTABLE` (non-required pending), so the preflight clears. No failing required check.

### Correctness

The owner check (spec 0158) is implemented at `scripts/status-mcp-server.sh:72-96` and behaves as the spec requires on every branch:

- **R1 (verify listener vs supervisor, non-zero naming both PIDs on mismatch).** Gated on `rc -eq 0` (liveness + auth), the check compares `mcp_listener_pid` against `mcp_supervisor_pid`; a mismatch prints `*** USURPED LISTENER ***` naming both PIDs and sets `rc=1`. Verified live: with `MEMPALACE_MCP_EXPECTED_PID=1234 MEMPALACE_MCP_LISTENER_PID=5678` the probe prints `PID 5678 is answering on 127.0.0.1:41893, but the supervisor runs PID 1234`.
- **R2 (expected process from the OS supervisor, not a falsifiable file).** `mcp_supervisor_pid` (`scripts/lib/common.sh:767-800`) reads `launchctl print gui/<uid>/<label>` on macOS and `systemctl --user show -p MainPID` on Linux — never a file. Verified live: on this machine launchctl reports pid 791, the sed extraction returns 791, and `lsof` on port 41893 returns 791.
- **R4 (unverifiable → non-zero).** When either side is empty, the probe prints `UNVERIFIABLE (listener PID …, expected PID …)` and sets `rc=1`. Both sub-cases verified live.
- **R5 (recovery action named).** The usurped report names the token rotation (`rm -f $(mcp_token_path)`, then re-run `bash scripts/switch-mempalace-http.sh`), mirroring the uninstall disclosure at `scripts/uninstall-mcp-daemon.sh:62-67`.
- **Gating and ordering.** The `rc -eq 0` gate means the owner check runs in the "serving and healthy" state — the tiebreaker when every other check passes; when auth is broken the probe already exits non-zero, so no failure is masked. Running the owner check before the launcher-drift check (which itself runs unconditionally) ensures a drifted launcher cannot hide a usurped listener and both problems are reported.
- **Set-but-empty contract.** Both helpers use `[ -n "${VAR+x}" ]` to distinguish set-but-empty (forces undeterminable) from unset (falls through to the real lookup); safe under `set -u`.

### Convention compliance

- Gitmoji titles/commits (`✨`, `📝`); no conventional-commit prefixes.
- No close keywords: `closingIssuesReferences` is empty; the second commit's "(#745)" carries no close keyword.
- English-only project content throughout.
- Version-bump rule not triggered: no skill/agent source touched (only `scripts/lib/common.sh`, `scripts/status-mcp-server.sh`, `scripts/tests/test-mcp-daemon.sh`, `specs/0158-…`); `check-skill-versions` passes.
- CLI-matrix rule not triggered: the status probe is not a CLI integration point; `check-ci-parity`/`ci-parity` pass.
- Spec status transition `approved → implemented` is recorded in the implementation PR's own commit (metadata-only, one line), per `docs/spec-format.md` → *Recording a status transition* (second case).
- PR body follows the required format (purpose, How to read, How to test, Detailed description).

### Test coverage

Section 13d (`scripts/tests/test-mcp-daemon.sh:669-739`) adds four hermetic cases covering the four served-port branches, each differing from a healthy run by exactly one override (R3): healthy match (exit 0, VERIFIED), usurped mismatch (non-zero, USURPED, both PIDs named, recovery named), unverifiable-no-supervisor (non-zero, UNVERIFIABLE), unverifiable-no-listener (non-zero, UNVERIFIABLE, forced by set-but-empty `MEMPALACE_MCP_LISTENER_PID`). The healthy-match case asserts exit 0, which requires every probe check to pass — the launcher is re-installed first because 13c corrupted its hash (v2-F1).

Test run in the worktree: `bash scripts/tests/test-mcp-daemon.sh` → `passed: 84   failed: 0`, exit 0.

### Linter findings

- `lint-shell.sh`: findings are all pre-existing — `scripts/lib/common.sh` is non-executable and lacks `set -e` on `main` too (it is a sourced library, so both are correct as-is); the shellcheck hits are in code untouched by this PR. The new code (`common.sh:728-800`, `status-mcp-server.sh:72-96`, `test-mcp-daemon.sh:669-739`) is shellcheck-clean; the SC2015 info notes in the new assertions follow the file's established `A && B || C` idiom.
- `lint-markdown.sh`: clean (exit 0).

### Blast radius

`scripts/switch-mempalace-http.sh:62` is the only consumer of the probe's exit code; the PR body acknowledges it. Verified: the switch installs the supervisor (`install_mcp_daemon`) before the probe gate, so the owner check passes in the normal flow. `Taskfile.yml:45` (`mempalace:status`) runs the probe as a status report — the exit code is the report itself, not a transaction gate. `doctor-mempalace.sh` references the probe only in comments. The two new helpers have no consumers other than `status-mcp-server.sh`.

### Findings

- **Non-blocking (class: tech)** — macOS `launchctl print` `pid = 0` asymmetry: Linux explicitly treats `MainPID=0` as undeterminable (`common.sh:783-784`), but macOS would treat a `pid = 0` line as a real PID, yielding USURPED rather than UNVERIFIABLE. Unreachable in the serving state in practice (a launchd service with `pid = 0` is not running, so healthz would fail unless another process serves — in which case USURPED is arguably the right call). Cosmetic asymmetry only.
- **Non-blocking (class: tech)** — the usurped report's `$(mcp_token_path)` may create `~/.mempalace/` if absent (the helper `mkdir -p`s the palace parent). A status probe creating a directory is a minor impurity; harmless since it is the daemon's own home.

Issue #961, opened 2026-08-16
📋 Findings ledger — deferred non-blocking findings


hcross:
> [!IMPORTANT]
> **PERMANENT SINGLETON INFRASTRUCTURE — DO NOT CLOSE**
>
> This issue is the permanent **Findings Ledger** for the crewrig project (introduced by **spec 0162**, issue #885).
> It serves as the single persistent sink for all non-blocking findings deferred during REVIEW passes across all tickets.
> Under the framework contract, **this issue SHALL NEVER be closed**.

---

## 📋 Purpose & Role

During the **REVIEW** stage of the development lifecycle, non-blocking findings that do not block a PR merge can be routed to this ledger rather than looping endlessly or being silently dropped (*"deferred ≠ dropped"*).

This ledger provides durable visibility, ownership, and periodic triage for deferred technical debt.

## 🔄 Drain Protocol

To trigger a drain pass, the project maintainer posts a `DRAIN` comment on this issue. Each open entry is evaluated by the maintainer against one of three dispositions:

| Disposition | Action |
|---|---|
| **Promote** | The finding is material; a dedicated ticket is created to resolve it. |
| **Accept** | The finding is noted but deemed non-actionable; the entry is closed with a recorded rationale. |
| **Carry** | The finding remains valid but is deferred to the next drain pass. |

> [!NOTE]
> Only the human maintainer can dispose of ledger entries. Agents may draft recommendations but cannot auto-dispose entries.

## 🛡️ Growth Guardrails

- **Warning threshold:** An automated warning is posted when the open-entry count exceeds **10**.
- **Hard cap & lock:** When the open-entry count exceeds **20**, the user is paged and all further ledger-route operations across the project are **blocked** until a `DRAIN` pass is executed and at least one entry is disposed.

hcross:
## Ledger entries — PR #981 / ticket #970 (spec 0166, reviewer seat)

Four non-blocking findings from the `review/970` seat's pass 3, routed here
because **the user authorized the merge of #981 as it stood**, declining the
orchestrator's recommendation of a fourth iteration. All four were `APPROVE`-
compatible; none blocked. Merged at `82c4df5`.

Per spec 0162 R2, each entry records source PR, source ticket, class, summary,
date routed, and routing actor.

| Entry | Source PR | Ticket | Class | Summary | Routed | Actor |
|---|---|---|---|---|---|---|
[...]

PR #981, opened 2026-08-20
📝 Seat one continuous reviewer per review surface (spec 0166)

### LINKED ISSUE #970: ⚡ Keep the same reviewer across the review iterations of a ticket
## Problem

Every review pass in the lifecycle spawns a **fresh** reviewer:

- `docs/plan-review-protocol.md` — the plan is reviewed by a second `architect`
  spawned cold, and a revised plan "is posted and re-reviewed cold".
- `docs/retroactive-loop.md` — the REVIEW loop re-spawns `pr-reviewer` cold on
  every iteration; lower-precedence findings are "re-tagged onto the next
  iteration's verdict by the next reviewer spawned cold".
- `docs/agent-team-protocol.md` — spec-PRs are cold-reviewed too.

The cold-start contract exists to guarantee **independence from the author**
(`artifacts/core/agents/pr-reviewer/AGENT.md` → *Cold start contract*). It does
not, on its own, require a *different* reviewer on each iteration — yet that is
what the current wording mandates.

## Consequence

Each new reviewer arrives with no memory of what the previous pass already
examined and accepted. It re-audits the whole artifact and surfaces findings on
surfaces that were untouched since the previous verdict. The observable effect
is a review loop that rarely converges in two passes, and a class of findings
that is not a defect signal but a reviewer-turnover artifact.

The lifecycle already has cost-bounding levers — `max-iterations` (ADR-0010,
default 5), the per-tier PLAN-revision cap (spec 0138) — but those *truncate*
the loop; they do not make it converge.

## Intent

Make the reviewer for a given (ticket, stage) **stable across iterations**: the
same reviewer that raised the findings is the one that checks whether they were
addressed. Independence from the author is preserved; only the reviewer-to-
reviewer turnover is removed.

Scope covers all three review surfaces named above: spec review, plan review,
and PR review.

## Open question for the SPECS stage

Whether "the same reviewer" means a persistent agent process (Claude Code
`SendMessage`) or a stable reviewer *seat* re-instantiated each pass with its
own prior verdicts as input. The second reading is the only one reachable on
CLIs with no coordination bus (`docs/agent-team-protocol.md` → *On CLIs with no
multi-agent coordination surface*), where parity is reached through sequential
`Agent` spawns that inherit no context. The spec must resolve this before any
plan is written.

## Related

- #882 / #883 / #884 — lifecycle cost bounding (truncation levers, complementary
  to this ticket).
- Prior art inside the project: `docs/plan-review-protocol.md` already mandates a
  **Finding Traceability Audit** on every revised plan. The REVIEW loop has no
  equivalent obligation.

hcross:
Implements spec 0166: every review pass in the lifecycle is now attributed to
one **reviewer seat** keyed `<surface>/<ticket>`, so the reviewer that raised a
finding is the reviewer that judges whether it was addressed. Documentation-only
— the contract lands as one new normative page plus pointers at the exact
decision points, and nothing executes it mechanically.

Closes #970

## How to read this PR?

Read `docs/reviewer-seat.md` first and in full. It is the whole contract — new
in this PR, and every other edit here is a pointer into it. Its load-bearing
sections, in the order a reviewer needs them:

| Section | What it settles |
|---|---|
| *Seat identity* | the key `<surface>/<ticket>[#<generation>]`, one live seat per surface |
| *The seat line, and where it goes* | where the line goes, per verdict transport |
| *Seat dossier* + *Reconstructing a dossier* | what a fresh pass rebuilds, and the commands that rebuild it |
| *What counts as a dossier entry* | which posted verdicts count, and how a voided one is excluded |
| *Instantiating a seated pass* | the references-only brief, and what makes a pass discardable |
| *Bounded scope from the second pass* | what pass two onward must read, and what it may skip |
| *Nothing mechanical observes this* | the enforcement gap, stated in the artifact itself |

Then read the six protocol-document edits, each of which is a pointer placed
where its reader already stands — `AGENTS.md`, `docs/retroactive-loop.md`,
`docs/plan-review-protocol.md`, `docs/agent-team-protocol.md`,
`docs/plan-format.md`, `docs/cli-matrix.md`. Then the four
`artifacts/core/` component sources, which gain only the obligations a reviewer
discharges while composing a verdict. The 16 files under `.claude/`, `.gemini/`,
`.github/` and `.agents/` are generated by `scripts/build-components.sh` and
carry no independent content — skip them.

**Three design decisions worth knowing before you judge the shape:**

1. **One contract document, pointers everywhere else.** The seat contract binds
   three surfaces, two reviewer roles and the orchestrator, so paraphrasing it
   into each protocol document would guarantee drift. The plan cites four
   measured instances of exactly that drift already in the tree (the
   routing-matrix pointers repaired in this diff). The cost is accepted openly:
   `docs/reviewer-seat.md` is a fifth place a reviewer must know to look.
2. **Anything varying per verdict transport is stated once.** Where the `seat:`
   line goes depends on which transport carries the verdict, and every case is
   defined in *The seat line, and where it goes*. No component source re-derives
   it — they defer, in the same words. Deliberately not enumerated here: the set
   grew by one in review, and a second copy of it in this body would be the very
   drift the design decision exists to prevent.
3. **Spec 0166 R20 is a hard no-change list.** Gating, counters, routing
   precedence, termination, the loop caps, and which role occupies which review
   surface are all untouched. Two places where that reads as an omission are
   deliberate — see *Deliberate non-changes* below.

## How to test this PR?

**Nothing in this suite verifies the contract this PR introduces.** Spec 0166 →
*Out of scope* rules out a mechanical checker for the seat line, the finding
identifiers, and the prior-finding audit by name; the contract stays a documented
procedure the orchestrator follows. So no green check can observe whether a pass
actually emits `seat:`, uses `i<N>-F<M>`, opens with a prior-finding audit, or
bounds its reading. The only evidence is a live review loop. Read the suite below
as certifying that the *documents* are well formed and the *build* is consistent
— nothing more.

From a checkout of this branch:

```sh
bash scripts/build-components.sh --target all --check   # 16 outputs match sources
bash scripts/build-docs-index.sh --check                # docs/index.json matches tree
bash scripts/check-agents-size.sh                       # AGENTS.md under byte ceiling
bash scripts/check-markdown-links.sh                    # every link resolves
BASE_REF=crewrig/main bash scripts/check-skill-versions.sh   # 4 version bumps present
node scripts/lib/spec-linter.js specs/0166-stable-reviewer-seat.md
git diff --name-only crewrig/main...HEAD | grep '\.md$' | xargs npx markdownlint
```

Every one exits `0` at the current head, each status captured directly rather
than through a pipeline (`markdownlint` over the changed Markdown files the last
command selects). What each one actually certifies:

- **Well-formed Markdown across every changed Markdown file, and every link
  resolving.**
- **`AGENTS.md` under its ceiling** — 19 740 B → 20 523 B against `THRESHOLD=22000`
  in `scripts/check-agents-size.sh:5`, a 1 477-byte margin. The script fails on
  `-ge` (line 10), so this is a hard failure, not a warning.
- **Four version bumps present**, one per modified component source.
- **16 build outputs byte-identical to their sources** across all four CLIs.
- **`docs/index.json` matching the docs tree** after the new page.

**Failure mode worth exercising** — the byte gate is the check most likely to
bite the next editor of this contract, and it is live rather than vacuous:
padding `AGENTS.md` by 2 000 B makes `scripts/check-agents-size.sh` exit non-zero.
The equivalents hold for the other two checks that were mutation-tested: reverting
any one of the four version bumps fails `check-skill-versions.sh`, and appending a
single byte to any of the 16 build outputs fails
`build-components.sh --target all --check`.

Implements the 20 requirements of `specs/0166-stable-reviewer-seat.md` along the
14 steps of the approved plan v2 (comment `5345066335` on issue #970, APPROVE'd
by the `plan/970` seat in comment `5345181244`). For the diff totals, read the
Files-changed tab rather than a figure quoted here — the forge keeps them
current as the head moves, and the breakdown below is the part that stays true.

Iteration 2 closed three review findings (`i1-F1`–`i1-F3`); their dispositions
are recorded on issue #970.

### Added

- **`docs/reviewer-seat.md`** (new) — the normative contract, covering
  R1–R19. Written forge-neutral with `gh` as the worked example. Its metadata
  block sits immediately after the H1, where `scripts/build-docs-index.sh`
  anchors it; `section=lifecycle`, `nav_order=20` (behind
  `docs/retroactive-loop.md` at 10, the section's only other page).
- **`AGENTS.md` → `## Reviewer seat`** — a pointer stub between
  *Retroactive review loop* and *Pull Request Format*, 783 B including its
  trailing separator.
- **`docs/cli-matrix.md`** — one new row for the seat contract, naming
  `docs/reviewer-seat.md` and the four component sources as source of truth,
  with all four CLI columns satisfied. The obligation fires because the diff
  touches `artifacts/**`; the *Protocol-only exemption* does not apply, since
  that covers diffs whose sole trigger-surface file is a top-level entry point.
  No `Parity gaps` entry is warranted — the contract is discharged with the forge
  CLI and a self-contained brief on all four assistants (R19).
- **`docs/index.json`** — regenerated by `scripts/build-docs-index.sh` for the
  new page.

### Modified — protocol documents (orchestrator obligations)

- **`docs/retroactive-loop.md`** — six anchored touch points on the `review`
  surface: seat instantiation at the REVIEW launch trigger; a retag re-issued
  from the same seat opening no new dossier entry (R18); lower-precedence
  findings re-tagged by the same seat's next pass rather than by a fresh
  reviewer; re-spawn of the same seat on the CI-red path, with R13(c) naming the
  head's CI state as unwaivable by the bound; the R14 cause added to the
  deferred-findings ledger; and the new page added to *Cross-references*.
- **`docs/plan-review-protocol.md`** — the plan is reviewed by the `plan/<ticket>`
  seat; a revised plan is re-reviewed by that same seat. Plus a note recording
  that this surface's existing `v<N>-F<M>` scheme and *Prior-finding
  traceability audit* already discharge R10–R12, so no second disposition record
  is introduced here.
- **`docs/agent-team-protocol.md`** — one new `### Seated review passes`
  subsection under *Standard Team Templates*, stating the orchestrator's
  instantiation duty once. A single subsection rather than a sentence appended to
  each of the three templates: the ordering-constraint paragraph is already
  duplicated verbatim three times in this file, and a fourth triplication would
  guarantee drift. The *Spec-reviewer obligation* paragraph gains the
  `specs/<ticket>` seat and the `s<N>-F<M>` identifier.
- **`docs/plan-format.md`** — one clause repointed; no other line, no
  frontmatter, no section changes.

### Modified — component sources (reviewer obligations), all four version-bumped

| Source | Version | What it gained |
|---|---|---|
| `artifacts/core/agents/pr-reviewer/AGENT.md` | 1.2.0 → 1.3.0 | frontmatter `description`, *Cold start contract*, and `## Activation` |
| `artifacts/core/skills/pr-reviewer/SKILL.md` | 1.3.0 → 1.4.0 | preflight, pass-two bound, seat line, prior-finding audit, identifiers, R15 |
| `artifacts/core/skills/architect/SKILL.md` | 1.2.0 → 1.3.0 | both the plan-reviewer and spec-reviewer branches |
| `artifacts/core/skills/spec-author/SKILL.md` | 1.6.0 → 1.7.0 | the *As reviewer* bullet: seat, seat line, `s<N>-F<M>`, R15 |

Two details a later editor should not undo:

- **`## Activation` in `AGENT.md` had to change, not just the contract section.**
  That section is the copy-paste invocation template, and therefore literally the
  brief an orchestrator sends. Left alone, the contract section would promise a
  seat-keyed brief while the only worked spawn form below it still passed a bare
  PR number — so every real pass would land on the R16 vacancy path and R13's
  bound would never engage on any ticket, invisibly, since no CI job reads a
  brief.
- **`architect/SKILL.md` gained both branches rather than losing one.** Deleting
  the clause that sanctions `architect` on the `specs` surface was the reviewer's
  alternative remedy and is barred: R20 forbids changing which role occupies
  which review surface.

### Modified — generated

16 build outputs under `.claude/`, `.gemini/`, `.github/` and `.agents/`,
regenerated by `scripts/build-components.sh` and staged in the same commit per
`AGENTS.md` → *Agent Team Protocol* → *Built components*.

### Spec status transition

`specs/0166-stable-reviewer-seat.md` moves `status: approved` →
`status: implemented` in its own commit on this branch — one line, no body line,
no `id`, no `slug`, no `version`. Per `docs/spec-format.md` → *Recording the
`approved` → `implemented` transition (in the implementation PR)* this rides this
PR's squash commit, so the spec lands on `main` already carrying
`status: implemented` at the moment its implementation lands. No separate
transition PR is required.

### Deliberate non-changes

Both of these match a grep a reviewer will run, and both are left alone on
purpose:

1. **`docs/retroactive-loop.md` → *REVIEW launch trigger* still names
   `pr-reviewer` in its mode table.** Spec 0166 R20 freezes role-to-surface
   assignment; renaming the occupant would be the one thing R20 names explicitly.
2. **`docs/plan-review-protocol.md` → *Countable invariant* is untouched, even
   though its two worked greps genuinely return 0 against their targets.** This
   is real, pre-existing drift, found while auditing. It sits in a clause spec
   0166 does not amend, and it is a normative verification recipe mandated by
   specs 0004/0138 — repairing it here would change a verification recipe under
   cover of a documentation ticket. The findings ledger (issue #961) is the
   proportionate route if a reviewer wants it tracked.

The plan carries a longer no-change list covering termination conditions, the
`iter:N` counter, routing precedence, the guardrail, the per-tier PLAN cap, every
user gate, and the six unrelated senses of the word "cold" — consult it before
concluding that a surface was missed.

### Rollback

`git revert` of this PR's single squash commit restores every touched file,
including the 16 build outputs and `docs/index.json` (committed artifacts, not
build-time products). The status transition rides the same commit, so the revert
also returns spec 0166 to `status: approved` — the correct post-revert state.
Re-run `bash scripts/build-docs-index.sh --check` afterwards to confirm the index
matches the reverted tree.


hcross:
## Verdict: REQUEST CHANGES

seat: review/970

Cold-start pass. This reviewer was instantiated with a PR number and no
authoring-session context; the diff, metadata, conventions and every file
quoted below were retrieved independently through `gh` and a dedicated
worktree at the head commit.

**Seat vacancy (disclosed, per the contract this PR introduces).** The
brief that instantiated this pass carried no seat key, no last-examined
revision and no prior-verdict location, and no verdict on this PR carries a
`seat:` line. Under `docs/reviewer-seat.md` → *Vacant seat* this pass is
therefore a **vacant seat, cause: ticket predates the contract** — which is
also the path spec 0166 R16 prescribes for a loop already running when the
contract lands. Nothing was bounded; the whole artifact was examined. The
`review/970` key and the `i1-*` identifiers below are adopted voluntarily to
exercise the contract under review — it is not yet in force on `main`.

**Observation on this pass's own instrumentation:** the copy of
`pr-reviewer/SKILL.md` that governed this review resolved from the repository
checkout on the **base** branch, byte-identical to `main`
(`sha256:47770e0e…`), not from the PR head (`sha256:4018c3d4…`). This PR
rewrites that very skill. So this pass was governed by the pre-seat contract
while reviewing the change that introduces the seat contract — worth knowing
when reading the verdict, and not a defect of the diff.

---

### CI status

All **29** required checks **pass** at head `70dba82`; zero failing, zero
pending (`gh pr checks 981`). Preflight clear.

`build`, `changeset-coverage`, `check-agents-size`, `check-ci-parity`,
`check-extension-version-bump`, `check-feedback-routing`,
`check-skill-versions`, `check-spec-id-reserved`, `chroma-mcp`, `ci-parity`,
`component-drift`, `core-paths`, `docs-index`, `e2e`, `extension-install`,
`extension-manifest`, `extension-pivot`, `extension-provenance`,
`figure-labels`, `frontmatter`, `gitlab-ci-check`, `lint-markdown`,
`lint-specs`, `markdown-links`, `mempalace`, `misc`, `setup`,
`test-harness-curate`, `test-wiring`.

Branch is up-to-date with its base: `git merge-base crewrig/main 70dba82`
returns `f8ba80c`, which **is** `crewrig/main`. The *Up-to-date merge
precondition* is satisfied.

Green here certifies less than it appears to, and the PR body says so
plainly: spec 0166 → *Out of scope* rules out a mechanical checker, so no
job observes whether a pass emits a `seat:` line, bounds its reading, or
opens with an audit. **Every finding below is therefore invisible to CI by
construction** — in a contract nothing executes, a documentation defect is
the defect, with no runtime to catch it later.

---

### Findings

#### `i1-F1` — a rung-3 verdict is durable, seat-lined, and unreachable by the prescribed reconstruction

**class: tech**

`docs/reviewer-seat.md:104-107` (*Seat dossier*) admits three dossier
locations for **any** surface — "the spec pull request, **on the logbook
issue**, or the implementation pull request". But *Reconstructing a dossier*
partitions its recipes by surface and queries the logbook issue **only**
under `# plan surface` (`docs/reviewer-seat.md:122-124`). The
pull-request-surface recipe (`docs/reviewer-seat.md:126-128`) enumerates
exactly two locations: `gh pr view <pr> --json comments` and
`gh api …/pulls/<pr>/reviews`.

That third location is not hypothetical — it is a shipped code path in this
same diff. `artifacts/core/skills/pr-reviewer/SKILL.md:203-205`, rung 3 of
the posting ladder, directs a reviewer whose posting is denied to "Return the
full verdict and findings to the orchestrator, **which records them in the
logbook issue**."

So a `review`-surface verdict can be durable, carry its `seat:` line, and sit
in a place the `review`-surface reconstruction never queries. The next pass
reads an empty dossier and declares the seat vacant — the precise failure the
document names and sets out to prevent one paragraph earlier
(`docs/reviewer-seat.md:131-136`: "declares itself vacant for no reason").
It also defeats spec 0166 R7 on that path.

**Failure scenario.** Iteration 1 on a solo-maintainer repo: the reviewer's
`gh pr comment` is refused by a permission classifier, so rung 3 fires and
the orchestrator records the verdict — with `seat: review/970` and findings
`i1-F1…` — on logbook issue #970. Iteration 2 instantiates `review/970`,
follows the pull-request recipe, gets zero seat-lined verdicts from both PR
transports, and reports "vacant seat, cause: prior verdicts absent". The
prior findings are never audited, and R12's audit obligation silently
evaporates even though the record exists and is machine-readable.

**Remedy.** Add the logbook issue to the pull-request-surface enumeration
(the `plan`-surface `--jq` filter already works verbatim), or state at rung 3
that a verdict recorded there is a dossier entry the reconstruction must
include. One line either way.

---

#### `i1-F2` — the dossier is reconstructed from seat lines, but a discarded or retagged verdict carries no discriminator

**class: tech**

The dossier is defined as reconstructible "from the forge alone", enumerated
by the `seat:` line (`docs/reviewer-seat.md:117-119`). Two rules then declare
that certain posted verdicts are *not* dossier entries, without giving a
reconstructing pass any way to tell:

- `docs/reviewer-seat.md:162-165` — a pass whose brief breaches
  references-only "is discarded. Its verdict is not consumed, it is not
  entered in the dossier, and it does not increment the iteration counter."
- `docs/reviewer-seat.md:304-311` (*Retagged verdicts*) — a retagged
  verdict's "corrected form **replaces** the malformed one in the dossier."

Both describe bookkeeping in the orchestrator's head. Neither marks the
superseded artifact on the forge. The discard case is the sharper one: the
reviewer posts its verdict as its own step 6, and only then does the
orchestrator judge the brief non-conforming — so a discarded pass's verdict
is **already public**, well-formed, and byte-indistinguishable from a
consumed one. A later pass enumerating `^seat: ` finds both and cannot tell
which counts.

The generation suffix (`#<generation>`) already solves exactly this problem
for retirement; discard and retag have no equivalent.

**Failure scenario.** The orchestrator briefs `specs/970` with the
developer's change summary attached. The pass posts a verdict raising
`s1-F1` and `s1-F2`, then is discarded under R9. A conforming pass is
re-instantiated, posts its own verdict, and — because the `specs` ordinal
"counts passes, not artifacts" (`docs/reviewer-seat.md:202-204`) — the next
pass reconstructing the dossier sees two seat-lined verdicts, counts two
passes, mints `s3-*` instead of `s2-*`, and runs a prior-finding audit over
`s1-F1`/`s1-F2`: findings from a pass whose verdict R9 says must not be
consumed. The identifier scheme R11 calls "unique and stable" collides, and
the audit is polluted by a discarded pass.

**Remedy.** Give the reconstruction a discriminator — e.g. require the
orchestrator to edit a discarded or superseded verdict to drop or annotate
its `seat:` line, and have *Reconstructing a dossier* say that a verdict
without a live seat line is not a dossier entry.

---

#### `i1-F3` — `architect/SKILL.md` re-derives transport-specific placement, against this PR's own stated invariant

**class: tech**

`artifacts/core/skills/architect/SKILL.md:219-223` tells the spec reviewer
the `seat:` line's placement "follows the pull-request transport rule
(`docs/reviewer-seat.md` → *The seat line, and where it goes*): there is no
`### Verdict:` line to anchor to on a formal review, and the line goes first
in the body."

It points at the right rule and then derives one of that rule's **two**
pull-request branches — and it derives the branch that does not fire on this
project. `docs/reviewer-seat.md:88-90` gives the other: a shared-identity
plain comment "opens with `## Verdict: …`, so the seat line goes immediately
after it." On crewrig the authenticated identity and the PR author are the
same account (verified this pass: both `hcross`), so the shared-identity rung
is the live transport for a spec-PR review, and `plan-review-protocol.md`
already acknowledges that workaround on the sibling surface.

A spec reviewer following `architect/SKILL.md` literally therefore places the
seat line ahead of the `## Verdict: …` header, contradicting the contract.

This is also a self-inconsistency worth naming: the PR body's design
decision #2 states "Anything varying per verdict transport is stated once…
no component source re-derives them." `pr-reviewer/SKILL.md:130-133` honours
that exactly — "take it from `docs/reviewer-seat.md` → *The seat line, and
where it goes* **rather than deriving it here**." `architect/SKILL.md` is the
one source that re-derives it, and got the applicable branch wrong.

**Remedy.** Delete the derived clause and defer, matching
`pr-reviewer/SKILL.md`'s wording.

---

### Correctness

Verified independently, and clean:

- **Cross-references all resolve.** Every section named by
  `docs/reviewer-seat.md` exists at head — including the four that my first
  ATX-heading grep reported missing (*Review rule*, *Reviewer-minted
  identifiers*, *Prior-finding traceability audit*, *Spec-reviewer
  obligation*), which are bold-prose headers; `docs/plan-review-protocol.md`
  carries no ATX headings at all. My instrument was wrong, not the diff.
- **The one verbatim quotation is faithful.** `docs/reviewer-seat.md:262-263`
  quotes "route every non-blocking finding to the **ledger** by default. No
  user gate SHALL fire" and generalizes it to INTERMEDIATE/MINIMAL/AUTO. The
  source is `docs/retroactive-loop.md:305`, the INTERMEDIATE row; the MINIMAL
  and AUTO rows both restate it ("Same as…; no user gate SHALL fire"), so the
  generalization holds.
- **The routing-matrix repoint is a genuine repair.** `AGENTS.md` →
  *Retroactive review loop* is a pointer stub holding no matrix, so the four
  pointers claiming it restates the matrix were false. Both new targets
  exist: `docs/retroactive-loop.md:104` and ADR-0010:210. No stale instance
  survives outside `specs/0011`, where the phrase is incidental.
- **`hcross/crewrig` → `crewrig/crewrig`** in `AGENT.md` → *Activation* is an
  unflagged correctness fix; two forks exist and the canonical one is now
  named.
- **Spec status transition** is exactly the one `status:` line, no `id`, no
  `slug`, no `version`, no body line — as claimed.

**Minor, non-blocking:** the body states "Diff: 28 files, +1136/−82"; the PR
as a whole is **29 files, +1137/−83** (`git diff --shortstat f8ba80c..70dba82`).
The gap is exactly the spec status transition, which the body describes
separately as its own commit — so the figure is the implementation commit's,
not the PR's. Worth one clarifying word only.

### Convention compliance

- **Title** `📝 Seat one continuous reviewer per review surface (spec 0166)`
  — Gitmoji, no conventional-commit prefix. Conforms.
- **PR body** carries *How to read this PR?*, *How to test this PR?* and
  *Detailed description (for agents)*. Conforms to `docs/pr-format.md`.
- **Version bumps** — four component sources modified, four bumps present.
- **CLI-matrix co-maintenance** — diff touches `artifacts/**`, and row 30 is
  added with all four CLI columns populated. Conforms.
- **`docs/index.json`** regenerated for the new page (`nav_order: 20`).
- **Logbook** — `Closes #970`; `iter:1` present.
- **Language** — English throughout.

The body's disclosure of the untouched *Countable invariant* drift in
`docs/plan-review-protocol.md`, and its routing of that to ledger issue #961
rather than repairing a normative verification recipe under a documentation
ticket, is the proportionate call. I am not raising it.

### Test coverage

No test covers the contract, and the PR says so in bold rather than letting a
green suite imply otherwise — the correct disclosure. The eight listed
commands certify document well-formedness and build consistency only, and the
body mutation-tests three of them rather than asserting they are live.

I independently mutation-tested the reviewer's own Markdown linter: injecting
`#Bad heading` and trailing whitespace into a copy of `docs/reviewer-seat.md`
produced `MD018/no-missing-space-atx` and `MD009/no-trailing-spaces` at
exit 1. The clean result at head is a real green, not a vacuous one.

### Linter findings

Bundled scripts selected by extension over the 29 changed files (28 × `.md`,
1 × `.json`), run against the head commit in an isolated worktree:

- `lint-markdown.sh` — 28 files, **exit 0**, no output.
- `lint-json.sh` — `docs/index.json`, **exit 0**, no output.
- `lint-skill.sh`, `lint-shell.sh`, `lint-python.sh` — not applicable (no
  `*.sh`, no `*.py`; `SKILL.md` frontmatter and version bumps are covered by
  the repo's own `check-skill-versions` job, which passes).

One process note: an earlier run of these scripts reported
`docs/index.json : not a regular file` because the shell's working directory
was reset between the path computation and the invocation. Re-run with
verified absolute paths, both exit 0. Recorded because the first result was a
false finding produced by my own harness, not by the diff.

---

### Summary

Unusually strong documentation work. The contract is coherent, states its own
enforcement gap instead of hiding it (*Nothing mechanical observes this*),
argues its one-document-plus-pointers shape from four measured instances of
real drift, repairs that drift in passing, and pre-empts most of the greps a
reviewer would run. The three findings are all narrow under-specifications on
paths the contract itself already worries about, and each is a one-to-two-line
fix in the same documents.

`i1-F1` and `i1-F2` both undercut R7 — "the dossier SHALL be reconstructible
from the forge alone" — on paths this diff ships. Because no CI job observes
any of this and the first live seated loop is the only detector the contract
has, these are worth closing before the contract lands rather than
discovering them on the loop that was supposed to benefit from it.

All three findings are `class: tech` → DEV.

hcross:
## Verdict: REQUEST CHANGES

seat: review/970

Pass 2 of this seat, generation 1. Bounded per `docs/reviewer-seat.md` →
*Bounded scope from the second pass* to the change across
`70dba82..8ba6287` (16 files), the R10 disposition record (comment
`5359055143`), the CI state of head `8ba6287`, and the surfaces that change
reaches. Prior verdict: comment `5358817507`.

---

### Prior-finding audit

| Finding | Recorded disposition | This pass |
|---|---|---|
| `i1-F1` | addressed (`d341301`, `8ba6287`) | **accepted** |
| `i1-F2` | addressed (`d341301`) | **accepted** |
| `i1-F3` | addressed (`d341301`, `8ba6287`) | **accepted** |

**`i1-F1` — accepted, and the declined half was genuinely unnecessary.**
The recipe now queries the logbook issue on **every** surface
(`docs/reviewer-seat.md:133-135`), and the prose names the posting-denied
rung as the shipped path that puts a PR-surface verdict there. I was asked
to judge the declined half rather than take it: I agree it was not needed.
My finding offered the two remedies disjunctively, and the argument for
choosing this one is stronger than the one I gave — R6
(`docs/reviewer-seat.md:111`) lists its locations without partitioning them
by surface, so the recipe was the defective artifact and narrowing the
admissible-locations clause would have contradicted the spec. The second
argument is the decisive one: a rung-3 note informs the *reviewer*, and the
party that needs the information is the *next pass*, which follows the
recipe. `pr-reviewer/SKILL.md` being untouched across the whole iteration is
therefore correct, not an omission — I confirmed it is untouched and I am
not treating that as a gap.

The remedy also went past what I raised, correctly. My finding did not ask
where the seat line goes in a body the orchestrator records rather than the
reviewer posts; `docs/reviewer-seat.md:91-96` adds that fourth transport and
settles it, including that the surface stays `review`/`specs` because the
seat key names the artifact reviewed and not where the verdict landed. That
ambiguity was mine to have missed.

**`i1-F2` — accepted on substance.** My finding was that no discriminator
existed. One now does: *What counts as a dossier entry*
(`docs/reviewer-seat.md:154-187`) defines a dossier entry as a verdict whose
seat line is in R3's exact form, and voiding annotates the line out of that
form. `:218-222` and `:370-373` wire the discard and retag rules into it, so
neither now relies on orchestrator memory. I verified the mechanism rather
than reading it — see *Test coverage*. `i2-F1` below is a defect in how
widely that discriminator is applied, not a re-raise of the finding.

I was asked to judge the two rejections as arguments. Both are right:

- **"drop" rejected in favour of "annotate."** My wording ("drop or
  annotate") was loose. Annotating is strictly better and the stated reason
  is the correct one: it preserves which seat posted the verdict, keeps the
  void human-legible in place, and still leaves R3's exact form. Dropping
  would have destroyed evidence to achieve the same exclusion.
- **`#<generation>` rejected.** This one corrects me. I wrote that
  generation "already solves exactly this problem for retirement," which
  reads as a proposal to reuse it; the rejection is sound and better
  reasoned than my observation. A generation replaces a *seat* and drags two
  consequences with it — the successor inherits the dossier and examines the
  artifact in full (`docs/reviewer-seat.md:350-359`) — neither of which is
  wanted when one *pass* is void and the seat is intact. Overloading it
  would make a seat key stop denoting a seat, which is what R2 fixes.
  Writing the rejection into the contract as *Voiding is not a generation*
  (`:183-187`) so a later pass does not retry it is the right disposal.

**`i1-F3` — accepted.** All three sources now defer instead of deriving:
`architect/SKILL.md:219-223`, `AGENT.md` (*Seat obligations*), and
`spec-author/SKILL.md`. I re-ran my own check across all four component
sources at head: **none** derives a placement branch or states a transport
count. Leaving the plan-surface placement statement alone is correct — that
surface has one transport, so there is no branch to get wrong.

The `8ba6287` sweep is honest and I could not falsify it. The only surviving
cardinality claims in the contract are `:127` ("three admissible locations",
which is R6's own count) and `:137` ("both PR transports", which sits in the
comment directly above the two commands it counts). Both are at their
definition site, which is where the rule the record adopts permits a count.

---

### CI status

**29/29 required checks pass** at head `8ba6287`; zero failing, zero pending
(`gh pr checks 981`). Preflight clear, and never waived: the bound covers
this on every pass.

Base unmoved — `crewrig/main` is still `f8ba80c` and is the merge base of
`8ba6287`, so the branch is 0 behind.

As in pass 1, green certifies almost nothing here. Both findings below sat
under a fully green run, and so did all three of pass 1's — spec 0166 →
*Out of scope* rules out the checker that would see them.

---

### Findings

#### `i2-F1` — the void discriminator is applied to one of the three locations the recipe now enumerates

**class: tech** · blocking

The `i1-F2` remedy makes exclusion-from-dossier depend on an anchored match
against the seat line. The `i1-F1` remedy expands the recipe, on a pull-request surface, from two
locations to three. Those two changes did not meet: the anchored filter is
attached only to the logbook-issue query, and the two pull-request commands
are unfiltered as written.

```bash
# docs/reviewer-seat.md:133-139
gh issue view <ticket> --json comments \
  --jq '[.comments[] | select(.body | test("(?m)^seat: (specs|plan|review)/…$"))]'   # filtered

gh pr view <pr> --json comments            # :138 — no filter
gh api repos/<owner>/<repo>/pulls/<pr>/reviews   # :139 — no filter
```

The normative sentence at `:166-168` ("A dossier entry is a verdict whose
seat line is in the exact form required above") does cover all three
locations, so the **rule** is complete. What is incomplete is the recipe a
pass copies — on the two locations where a `review`- or `specs`-surface
verdict *normally* lands.

This matters more than a docs nit because of the claim the mechanism rests
on. The disposition record states that voiding makes a verdict "drop out of
the dossier **by itself**, with no one having to remember a rule or consult
a second store," and identifies this as the one place in the whole ticket
where a rule is enforced by the command a reader copies. That property holds
on one of three locations. On the other two the discriminator is advisory —
exactly the status the remedy was written to escape — and there is no CI
check anywhere that would notice.

**Failure scenario** (my original `i1-F2` scenario, still open on the PR
surface). The orchestrator briefs `specs/970` with the developer's change
summary attached. The pass posts its verdict on the spec PR raising `s1-F1`
and `s1-F2`, and is discarded under R9. The orchestrator voids it correctly,
annotating the seat line to `seat: specs/970 — VOID (discarded: …)`. Pass 2
reconstructs the dossier, runs the recipe, and reaches the spec PR through
`gh pr view <pr> --json comments` — unfiltered. The voided verdict comes back
and is counted: the `specs` ordinal, which "counts passes, not artifacts"
(`:259-261`), advances to `s3-*` instead of `s2-*`, and the prior-finding
audit runs over `s1-F1`/`s1-F2` — findings that *Instantiating a seated pass*
says must not be consumed.

**Verified, not reasoned.** Against the real PR: the filtered logbook query
and the unfiltered `gh pr view 981 --json comments` both return a
seat-lined body, and a simulated voided verdict is excluded by the anchored
filter (`false`) while the unfiltered command returns it. The two paths
disagree on the same input, which is the defect.

**Remedy.** Apply the same `--jq` test to the PR-comments query and pipe the
reviews query through it — two lines in the same code block. Worth doing in
the block rather than in prose, since the block is what gets copied.

---

#### `i2-F2` — the anchored form silently excludes a valid seat line carrying trailing whitespace

**class: tech** · non-blocking → deferred-findings ledger (#961)

`$`-anchoring the seat key (`docs/reviewer-seat.md:135`) is what makes the
void discriminator work, and I am not asking for it to be loosened — the
disposition record is right that dropping the anchor reopens `i1-F2`. The
finding is that the anchor is also unforgiving of whitespace, and that
failure is silent and asymmetric.

Measured against the shipped pattern:

| Body | Matches | Intended |
|---|---|---|
| `seat: review/970` | true | ✓ |
| `seat: review/970#2` | true | ✓ |
| `seat: review/970 — VOID (…)` | false | ✓ exclusion works |
| `seat: review/970` + 1 trailing space | **false** | ✗ |
| `seat: review/970` + 2 trailing spaces | **false** | ✗ |
| `seat: review/970\r\n` (CRLF body) | **false** | ✗ |

A valid verdict whose seat line picked up a trailing space drops out of the
dossier, and the failure presents as a false vacancy — the `i1-F1` outcome,
reached through a different door. Nothing tells the verdict's author that a
stray space voided their pass, and no check reads it.

Two honest qualifications, which are why this is non-blocking rather than
blocking:

- **Trailing whitespace is not the documented form.** R3 says "exact form,"
  so a whitespace-bearing line is arguably already malformed, and the filter
  is faithful to the rule. The complaint is about the consequence
  asymmetry — a voided verdict failing to match is intended; a valid one
  failing to match loses the whole dossier — not about fidelity.
- **The CRLF half is theoretical on current practice.** I checked: none of
  the 20 most recent comments on this repository carries a CR, because every
  one was posted through `gh`. It becomes reachable if a maintainer ever
  voids or posts a verdict through the GitHub web editor.

Note the interaction with `i2-F1`: fixing that finding applies this pattern
to all three locations, which widens this exposure rather than leaving it
flat. Worth hardening in the same edit — `[ \t]*\r?$` in place of `$` costs
nothing and preserves the void exclusion exactly, since the annotation
begins with a non-whitespace character.

---

### Correctness

Nothing else in the change is wrong. Specifically checked:

- **The fourth transport is consistent with the seat-key rule.** `:91-96`
  keeps the surface as `review`/`specs` for a verdict recorded on the
  logbook issue, which is right: the key names the artifact reviewed. It
  also means the logbook query can return `review/`- and `specs/`-keyed
  verdicts, which the widened surface alternation in the filter admits.
- **The `(?m)` flag is load-bearing, not decorative.** I expected jq's
  Oniguruma to treat `^`/`$` as line anchors by default and the flag as
  redundant; it does not. Dropping `(?m)` makes the canonical multi-line
  verdict body return `false`. The flag is correctly required.
- **The fallback path is coherent.** Where the orchestrator cannot edit the
  artifact (`:177-180`), the discriminator moves to the disposition record,
  whose location the instantiation brief already names — so a pass can
  actually reach it.
- **Generated outputs match sources.** `build-components.sh --target all
  --check` exits 0 at head; the 15 generated files in this change carry no
  independent content.
- **No second version bump is correct.** `docs/version-bump-convention.md`
  gates on the diff against the base ref; all four components are already
  bumped against `main` this cycle and nothing has shipped, so a second bump
  would claim two shipped revisions where there is one. Confirmed unchanged
  at `8ba6287`: 1.3.0 / 1.4.0 / 1.3.0 / 1.7.0.

### Convention compliance

Both commit subjects use Gitmoji (`🐛`). `iter:2` label present. The PR body
no longer quotes diff totals or a transport count, and gained a row for the
new `### What counts as a dossier entry` section. English throughout. The
*Countable invariant* drift remains ledgered to #961, unchanged and still the
proportionate call.

### Test coverage

Still nothing mechanical, still correctly disclosed. I verified the one
mechanism this iteration made executable rather than taking the record's
account of it:

- Re-derived the fixture result independently. The shipped anchored form
  matches the two valid keys and rejects the void annotation and the
  mid-sentence prose mention; the loose `^seat: ` form wrongly admits the
  void annotation. That reproduces the `i1-F2` defect and confirms the
  anchor is what closes it.
- Extended the fixture past what the record covered, which is where `i2-F2`
  came from — the record's five cases did not include whitespace or CRLF.
- `lint-markdown.sh` over all 16 changed files: **exit 0**. No `.json`,
  `.sh` or `.py` in scope, so the other three scripts do not apply.

### Linter findings

None. `lint-markdown.sh` exit 0 across the 16 files in scope; `build
--check` exit 0.

---

### Summary

Convergence is real: all three pass-1 findings are properly closed, two of
the three rejected halves were rejected for better reasons than I gave, and
the `8ba6287` sweep found two stale counts I had not raised. I could not
falsify any claim in the disposition record.

`i2-F1` is the reason this is not an approval, and it is narrow: the two
remedies landed in the same code block without meeting, leaving the new void
discriminator attached to one of the three locations the same edit added.
It is a two-line fix, and I would not hold the PR for `i2-F2`, which belongs
in the ledger.

Both findings are `class: tech` → DEV.

**Bounded scope was sufficient.** Everything above came from the 16 changed
files, the disposition record, the head's CI state, and the empirical
behaviour of the one command the change introduces. I raised no finding on
an unchanged surface and did not need to reach outside the bound. The one
thing I read outside the change — R6 at `:111` and *Retirement and
generation* at `:350-359` — was to judge the declined remedies, which is the
audit the bound explicitly keeps in scope, not a re-examination.

hcross:
## Verdict: APPROVE

seat: review/970

Pass 3 of this seat, generation 1. Bounded to the `8ba6287..6e5a01a` change
(one commit, one file, +29/−6 in `docs/reviewer-seat.md`), both disposition
comments (`5359055143` and its correction `5359293018`), the CI state of head
`6e5a01a`, and the surfaces that change reaches. Prior verdicts: `5358817507`
(pass 1), `5359161743` (pass 2).

No blocking findings. Three non-blocking findings below, all `class: tech`,
all for the deferred-findings ledger (#961) — spec 0166 runs
`interaction-mode: INTERMEDIATE`, so *Non-blocking conditional routing*
routes them there by default with no gate.

---

### Prior-finding audit

| Finding | Recorded disposition | This pass |
|---|---|---|
| `i2-F1` | addressed (`6e5a01a`) | **accepted** |
| `i2-F2` | addressed (`6e5a01a`), folded in rather than ledgered | **accepted** |

**`i2-F1` — accepted, and my remedy was wrong where the record says it was.**
The filter is now on all three queries (`:138-139`, `:144-145`, `:146-147`),
hoisted into one `SEAT` variable at `:135`, with the rationale stated at
`:150`.

I was asked to verify the accessor claim rather than take it, so I ran it:
`.comments[]` against the reviews endpoint fails with `expected an object but
got: array`, while the shipped `.[]` form returns `0` cleanly. My pass-2
remedy said "apply the same `--jq` test to the PR-comments query and pipe the
reviews query through it" — I never named the shape difference, and a literal
reading of my own sentence would have shipped a third query that errors. The
correction stands and it is mine to own: I proposed a uniform fix to two
endpoints with non-uniform shapes, having tested neither. The comment at
`:142-143` now warns the next reader about exactly the thing I missed.

**`i2-F2` — accepted, and folding it in rather than ledgering it was the
right call.** I had routed it to the ledger; the argument for pulling it
forward is the one I made myself — fixing `i2-F1` applies the pattern to all
three locations and widens the exposure rather than leaving it flat — so
deferring it would have shipped a wider hole than the one I measured.

I re-derived the pattern behaviour rather than reading the record's table,
across 14 cases through the real shell→jq path (where `\t` and `\r` are
interpreted as JSON escapes before reaching the regex engine, which is what
makes the bracket class and the optional CR mean what they appear to mean):

| Seat line | Old | New |
|---|---|---|
| clean | match | match |
| generation suffix `#2` | match | match |
| trailing space | **drop** | match |
| trailing tab | **drop** | match |
| real CRLF | **drop** | match |
| bare LF | match | match |
| generation + trailing space | **drop** | match |
| voided, em-dash | exclude | **exclude** |
| voided, tab-separated | exclude | **exclude** |
| voided + trailing space | exclude | **exclude** |
| voided, no space before dash | exclude | **exclude** |
| prose mid-sentence | exclude | **exclude** |
| leading whitespace | exclude | **exclude** |
| placeholder `<surface>/<ticket>` | exclude | **exclude** |

The property the fix had to preserve holds: all seven exclusions survive the
widened tail, because an annotation begins with a visible character. The four
old-pattern drops are the `i2-F2` defect, and they are as severe as the
record says — a single invisible trailing space dropped a seat's entire
dossier, since the enumeration is the only thing that finds it. Two of those
rows are also why the contract's own examples are safe by construction: the
seat-line spec uses placeholders and the void example is annotated, so
neither is self-matching.

**On the variable hoist, which I was asked to judge for a reader who copies
the block.** It holds. I ran the shipped block verbatim — substituting only
`<ticket>`, `<pr>` and `<owner>/<repo>`, leaving the quoting byte-identical —
in **both** `bash` and `zsh`, this project's interactive shell. Identical
results in both. The single-quoted assignment survives the paste, and the
`$SEAT` expansion inside the `\"…\"` jq string is not re-scanned, so the
trailing `$` of the pattern is safe. `i3-F3` is the one residual, and it is
about a *partial* copy, not this one.

**The live-forge claim is exact.** Run verbatim against the real forge:

| Query | Returned | Keys recovered |
|---|---|---|
| issue #970 (logbook) | **1** | `seat: plan/970` |
| PR #981 comments | **2** | `seat: review/970` ×2 |
| PR #981 reviews | **0** | — |

I checked the three are genuine verdicts rather than incidental matches:
`## PLAN review of v2 — issue #970` (the plan reviewer's voluntary
pre-adoption) and my own two `## Verdict: REQUEST CHANGES` bodies. The zero
on reviews is correct and not a miss — shared identity routes every verdict
on this repository to a plain comment, which is why the posting ladder's
rung 2 exists. This is the most checkable claim in the iteration and it
survives checking.

---

### CI status

**29/29 required checks pass** at head `6e5a01a`; zero failing, zero pending.
Preflight clear and never waived. Base unmoved: `crewrig/main` is still
`f8ba80c` and is the merge base of `6e5a01a`, so the branch is 0 behind.

`iter:3`. The guardrail halts at 5; that is not a reason to approve, and the
approval below does not rest on it.

---

### Findings

#### `i3-F1` — an illustrative seat line is counted as a dossier entry

**class: tech** · non-blocking → ledger #961

The enumeration matches raw comment text, and Markdown fencing is invisible
to it. A comment that merely *shows* a valid seat line on a line of its own —
a logbook entry explaining the mechanism, a verdict discussing the seat, a
training doc — becomes a phantom dossier entry.

Verified: a body containing

````text
```text
seat: review/970
```
````

returns `true` against the shipped pattern.

Consequence is the `i2-F1` family: a phantom entry inflates the `specs` pass
ordinal, which counts passes, and can make a vacant seat read as occupied.
Today's live counts are clean at 1/2/0, but only incidentally — my own pass-2
verdict displays seat lines inside a Markdown table, so those lines begin
with `|` and miss the anchor. Formatting luck, not design.

Non-blocking because it needs an authoring choice the contract's own rule
already discourages ("point at the definition; never restate its content"),
and because the document is immune to its own examples. Cheapest fix is one
sentence: an illustrative seat line must be shown in placeholder or
void-annotated form. Note that requiring a verdict header instead would
*not* work — the formal-review transport has no verdict line by design.

#### `i3-F2` — the fallback rule was absorbed into the whitespace paragraph

**class: tech** · non-blocking → ledger #961

The `i2-F2` prose insertion swallowed a pre-existing normative rule. At
`8ba6287` the fallback — where the discriminator lives when the orchestrator
cannot edit the artifact — was the tail of the paragraph about the voiding
mechanism, which is its topical home. At `6e5a01a` that paragraph ends at
`:191` ("visible to a human reading the verdict.") and the fallback now sits
mid-paragraph inside the new whitespace-rationale block (`:193-204`), joined
to it in the middle of `:200`.

Two consequences, both small and both real:

- **"Either way" lost its antecedent.** At `:203-204` it originally referred
  to the two voiding routes — annotate the seat line, or name the verdict in
  the disposition record. It now closes a paragraph whose first half is about
  trailing blanks, so it reads as if the two ways are the whitespace cases.
- **The line is the mechanical tell.** `:200` is 95 characters against 69–75
  for every other line in the paragraph — the insertion was made without
  re-wrapping. `.markdownlintrc` sets `MD013: false`, so no check sees it;
  I confirmed that setting rather than assuming it.

Nothing is factually wrong and no rule changed, so this is editorial. Fix is
a paragraph break before "Where the orchestrator cannot edit" and a re-wrap.

#### `i3-F3` — a partial copy of the block silently over-includes

**class: tech** · non-blocking → ledger #961

The hoist to a single `SEAT` variable is the right trade and I am not asking
for it to be reverted — a copy per query is a copy that can drift, as `:133-134`
says. The residual is that the failure mode inverted. Copying one query line
without the assignment leaves `$SEAT` empty, `test("")` matches every string,
and the query returns everything as a dossier entry.

Measured on issue #970, which has 10 comments:

| Invocation | Returned |
|---|---|
| whole block (SEAT set) | **1** — correct |
| one query line alone (SEAT unset) | **10** — every comment |

Before the hoist a mis-copy under-included, which surfaces as a false
vacancy; now it over-includes silently, which surfaces as phantom passes and
a dossier of non-verdicts. Neither is checked by anything.

Non-blocking: it requires ignoring the block's own comment that the pattern
is "used by every query below." A one-token fix makes it loud
instead of silent — `test(\"${SEAT:?run the SEAT= line first}\")` fails
locally with a message rather than returning a wrong answer. `set -u` would
be worse here, since a reader pastes this into an interactive shell.

---

### Correctness

Nothing in the change is wrong. Specifically checked beyond the audit above:
the `SEAT` assignment sits inside the same fenced block as all three queries,
so a whole-block copy always carries it; the widened surface alternation
still admits the `review`/`specs` keys the logbook query can legitimately
return under the fourth transport; and no `artifacts/` source was touched, so
no rebuild and no version bump were owed — the four component versions, the
`AGENTS.md` byte figures and the 16-output count all stand from earlier
iterations, and the record says so rather than re-deriving them.

The correction record's decision **not** to reprint the corrected pattern is
the right one, and it is the eighth instance of the class this logbook tracks
reaching the conclusion the class implies. I verified the pattern from
`docs/reviewer-seat.md`, which is where the record points, and had no second
copy to reconcile it against — which is the whole benefit.

### Convention compliance

Gitmoji commit subject (`🐛`), `iter:3` label, English throughout, single-file
change with no build or version obligation triggered. The *Countable
invariant* drift remains ledgered to #961.

### Test coverage

Still nothing mechanical, still correctly disclosed, and this iteration the
gap is narrower than the disclosure implies: the recipe is now executable
against the live forge and its output is checkable, which is as close to a
test as a doc-only contract gets. I ran it rather than reading it.

`markdownlint docs/reviewer-seat.md` in-tree at `6e5a01a`: **exit 0**, zero
output, with `.markdownlintrc` in scope.

One process note, since it is the same class the record has been counting and
this one is mine, this pass. Two of my own measurements this pass were false
before I corrected them. My first partial-copy test ran against PR #981, which
has exactly two comments and both are verdicts — so the broken and correct
forms both returned `2` and the test could not discriminate; re-running it on
issue #970 (10 comments) produced the 1-vs-10 contrast above. And an earlier
`EXIT=0` I read off a lint run was `head`'s exit status through a pipe, not
markdownlint's; the file is genuinely clean, but I only know that because I
re-ran it capturing the status directly. Both are the same shape as the eight
in the record: the instrument agreed with what I hoped rather than with what
it measured, and nothing announced it.

### Linter findings

None. `markdownlint` exit 0 on the one file in scope; no `.json`, `.sh` or
`.py` in scope, so the other four scripts do not apply.

---

### Summary

Approving. Both pass-2 findings are properly closed and I verified each
mechanism rather than accepting its description: the filter is on 3 of 3
queries with the correct accessor per endpoint, the widened tail admits every
valid form while still excluding all seven void and non-verdict shapes, the
block runs verbatim in both bash and zsh, and the shipped recipe reconstructs
this ticket's real dossier at 1/2/0 with the three keys it should recover and
no others. One of the two remedies also corrected my own pass-2 remedy, which
was wrong in a way I had not tested.

The three findings above are robustness and editorial residue, none of which
produces a wrong outcome on the contract's specified path, and all three go
to the ledger under INTERMEDIATE. Termination conditions are met: approving
verdict, zero blocking findings, green CI on the head examined, every
non-blocking finding disposed.

**The bound was sufficient, and I want to say that as a result rather than as
an absence of effort.** One commit and one file is the narrowest reading this
seat has had, and it was enough because everything the change touched was
reachable from it: the queries and the pattern are in that file, the two
prior findings were about that file, and the claims worth checking were
executable against the forge rather than located in some other document. I
reached outside the change exactly twice, both inside the bound's own terms —
`.markdownlintrc` to confirm `MD013: false` before calling `:200` a style
issue rather than a violation, and the pre-`6e5a01a` paragraph structure to
establish that `i3-F2` is a regression rather than a pre-existing shape. Both
are audit of the change, not re-examination of unchanged surfaces. I raised
no finding on a surface unchanged since pass 2.

hcross:
## Verdict: APPROVE — unchanged (addendum to pass 3)

seat: review/970

**This restates the standing verdict; it does not re-judge the artifact.** Head
is still `6e5a01a`, CI still 29/29, and the approval recorded in comment
`5359391666` stands on its own terms. This entry adds one non-blocking finding
and one measurement, neither of which disturbs it.

---

### `i3-F4` — the references-only rule bars authoring content, but not a brief's characterization of the artifacts it points at

**class: tech** · non-blocking

*Instantiating a seated pass* requires the brief to carry the **location** of
the seat's prior verdicts and of the disposition record, and bars "a diff, a
summary, an assessment, a rationale, or any other content that originates in
the authoring session." It says nothing about the brief **describing the
structure of the artifacts it points at** — and a description is a copy that
can drift.

**This is not hypothetical; it fired on this seat, on this pass.** The pass-3
brief characterized the R10 record as two comments that "read together and
neither supersedes the other." That was accurate when written and false within
minutes: the record's authoritative reading is that `5359055143` is superseded
in two passages by `5359293018`, with `5359315718` stating the order. The
brief carried no diff, no summary, no assessment and no rationale — it was
conforming under the rule as written, and still shipped a false claim about a
normative record.

The audit absorbed it without damage, because nothing in it was sourced from
the brief's characterization: the pattern was read from
`docs/reviewer-seat.md` and the recipe was executed against the live forge
rather than taken from either record's quotation of it. That is luck about
method, not a property the contract guarantees.

**Why this class of brief defect is unrepairable rather than merely
regrettable.** Every other artifact in the loop that restated something and
went stale — the PR body's transport count, a logbook comment's `jq` pattern,
two component sources' placement counts — was durable, so each was corrected
in place once found. The brief is the only artifact in the loop with **no
durable definition of its own**: it is generated fresh for each pass and
discarded when the pass ends. A restatement in it cannot be corrected later,
because by the time anyone knows it was wrong the artifact is gone and the
pass has already consumed it. It can only be avoided at authoring time. That
argument, not the incident, is the reason the clause is worth adding.

**Remedy.** One clause in *Instantiating a seated pass*: the brief points at
the disposition record and the prior verdicts, and does not characterize their
content, their structure, or their reading order. The generalized form of the
rule this ticket has now paid for nine times already covers it — *point at the
definition; never restate its content* — this extends its scope to the one
artifact that cannot be repaired after the fact.

**Unchanged-surface disclosure, stated plainly.** *Instantiating a seated
pass* is unchanged since this seat's pass 2 — `6e5a01a` touched only the
dossier and whitespace sections — so this finding lands on a surface outside
the change. Under *A finding on an unchanged surface* I must state which
condition of the bound returned that surface to scope, and **I cannot claim
one honestly.** It was not newly touched by the change, no prior finding's
remedy reaches it, and the change depends on no invariant of it. It came from
an out-of-band process incident, not from the diff. So by the contract's own
rule it is non-blocking, it is routed rather than looped, and the iteration is
not routed to an upstream stage on its account — which is why the approval
above is undisturbed.

I note for the record that this is the one finding on this ticket I had an
incentive to escalate, since it concerns the section governing how I am
briefed. The bound did not admit it, so it is non-blocking.

---

### `i3-F1` — measurement update: the hazard is held shut by a convention, not a mechanism

The two logbook comments appended after pass 3 were the first live test of
`i3-F1` after it was raised. It **did not fire**:

| Comment | Line-start seat lines | Total seat-key mentions |
|---|---|---|
| `5359293018` | 0 | 2 |
| `5359315718` | 0 | 2 |

The dossier recipe still returns 1 on issue #970, unchanged. Four illustrative
mentions of a seat key, and every one of them kept inline rather than on a line
of its own.

State the reason precisely, because it changes what the ledger entry means:
what holds the phantom-entry hazard shut today is **the authors' habit of
keeping illustrative mentions inline — a convention, not a mechanism.** No
check enforces it and no rule states it. It holds until the first person who
does not know the habit writes a line-start example: a training document, an
onboarding page, a verdict quoting the form it is discussing. That is a
negative result with a named expiry rather than a null one, and it is the
argument for the one-sentence fix `i3-F1` proposes.

---

### One observation, not a finding

This addendum is a **second dossier entry for a single pass**, which the
enumeration cannot distinguish from a second pass. On the `review` surface
that is harmless: `i<N>-F<M>` keys `<N>` to the `iter:N` label, so this entry
and the pass-3 verdict both sit at `i3-` with no ambiguity. On the `specs`
surface it would not be harmless — `s<N>` is defined as "the seat's pass
ordinal", but a reconstructing pass has only *entries* to count, so one pass
posting a verdict plus an addendum reads as two passes.

That is the same gap as `i3-F1` — the ordinal is defined over passes and
observed over entries — reached from the opposite direction, so I am recording
it against `i3-F1` rather than minting another identifier for it. Relatedly,
*The seat line, and where it goes* defines placement for four **verdict**
transports and says nothing about a dossier entry that is not a verdict; this
entry opens with a restated verdict line to stay inside the rule as written.

---

### Disposition

`i3-F4` is non-blocking and routes to the orchestrator's discretion: to the
deferred-findings ledger (#961) alongside `i3-F1`, `i3-F2` and `i3-F3` if this
PR merges as it stands, or to the author if a fourth iteration folds those
three in — in which case this one folds in with them at no extra cost, being a
single clause in a file that iteration would already be editing.

Verdict unchanged: **APPROVE**.

PR #1039, opened 2026-08-25
🐛 Deterministic fake server readiness wait in test-mcp-daemon.sh

### LINKED ISSUE #983: 🐛 test-mcp-daemon.sh section 13b flakes on a fixed sleep before probing a fake server
## What happens

`scripts/tests/test-mcp-daemon.sh` section 13b fails intermittently on CI:

```
FAIL no NOT ENFORCED report: MemPalace MCP HTTP daemon
passed: 92   failed: 1
```

It is a genuine flake, not a regression. The strongest evidence is that the
**same commit produced both outcomes inside a single workflow run**:

| Where | Execution | Result |
|---|---|---|
| run 32392966176, `chroma-mcp` job 96503166474 | cache miss, real run | **92 passed, 1 failed** |
| run 32392966176, `changeset-coverage` (same run, same SHA) | uncached full suite | **93 passed, 0 failed** |
| re-run of that same job, job 96507567335, no code change | cache miss, real run | **93 passed, 0 failed** |

Nothing changed between the failure and the two passes — not the SHA, not the
platform, not the test file.

## Probable cause: a fixed sleep instead of a readiness wait

Section 13b starts a fake server in the background and waits a flat second
before probing it:

```sh
"${MEMPALACE_PYTHON}" "${TEST_HOME}/fake-mcp.py" "${MEMPALACE_MCP_PORT}" 200 &
fake_pid=$!
sleep 1
out="$(bash "${REPO_DIR}/scripts/status-mcp-server.sh" 2>&1)"; rc=$?
```

If the interpreter has not bound the port within that second — ordinary on a
loaded shared runner — the probe finds nothing serving, so the report never
reaches the `NOT ENFORCED` branch and the `case` falls through to `nope`.

Consistent with the observed asymmetry: the paired assertion on the same lines
(`an unauthenticated probe returning non-401 exits non-zero`) passes even when
this one fails, because a probe that cannot reach the port also exits non-zero
— for the wrong reason. So the section reports one green and one red from the
same broken precondition, and the green is vacuous.

## Why it is worth fixing rather than re-running

- It is in the job that owns this repository's daemon coverage, so a flake
  there reads as a real failure and costs a triage cycle every time.
- The `ci-cache-guard` content key means a passing run gets **cached**, so a
  flake can be papered over by an unrelated cache hit on the next push — and
  conversely a red can persist in a reader's memory after it has stopped
  reproducing. Both directions mislead.
- The vacuous green on the sibling assertion means the section can be
  half-broken and still look half-right.

## Suggested direction

Not prescriptive:

- Replace the `sleep 1` with a bounded readiness wait — poll the port until it
  accepts a connection, with a timeout that fails loudly and names the
  precondition rather than the assertion.
- Consider making the sibling assertion distinguish "exited non-zero because
  authentication is not enforced" from "exited non-zero because nothing was
  listening", so a failed precondition cannot present as a pass.

The general shape worth removing is a fixed sleep guarding a network
precondition on a shared runner.

## How this was found

Surfaced while attributing a red `chroma-mcp` check on PR #969. That PR does not
modify this assertion (`git diff` over `scripts/tests/test-mcp-daemon.sh` shows
no change touching it), and the re-run cleared it, which is what established the
flake rather than a regression.

Note also, for anyone reading greens on that job: `chroma-mcp` had reported
green on the immediately preceding head with **every step a `ci-cache-guard`
cache hit and zero assertions printed**. That green certified that the keyed
inputs had not changed, not that the suite passed at that commit.

hcross:
This PR implements specification 0182 by replacing flat `sleep 1` intervals with an active readiness polling helper for background fake MCP servers in `scripts/tests/test-mcp-daemon.sh`. It resolves issue #983.

## How to read this PR?

Review `scripts/tests/test-mcp-daemon.sh`:
- The new `_start_fake_mcp` helper immediately after the `fake-mcp.py` fixture definition.
- The refactored calls across sections 13b, 13c, 13d, and 16.
- The precondition assertion added in section 13b to verify the server is live and responding to `/healthz` before probing with `status-mcp-server.sh`.

## How to test this PR?

Run the test suite locally:
```sh
```
Verify that all 98 assertions pass cleanly.

- Implemented `_start_fake_mcp` in `scripts/tests/test-mcp-daemon.sh`, which starts `fake-mcp.py` in the background with detached stdout/stderr (`>/dev/null 2>&1 &`) and polls `http://127.0.0.1:${port}/healthz` every 100ms up to 5s.
- Replaced 8 occurrences of `sleep 1` following `fake-mcp.py` launches across sections 13b, 13c, 13d, and 16.
- Added explicit `/healthz` live precondition verification in section 13b to eliminate vacuous passes on unreachable servers.
- Closes #983.

### DIFF
--- a/scripts/tests/test-mcp-daemon.sh
+++ b/scripts/tests/test-mcp-daemon.sh
@@ -724,6 +724,27 @@ with socketserver.TCPServer(("127.0.0.1", port), H) as s:
     s.serve_forever()
 PY
 
+# _start_fake_mcp — launch fake-mcp.py in the background and poll for readiness
+# on /healthz up to 5s. Replaces fixed sleep 1 intervals to prevent CI flakes (spec 0182).
+_start_fake_mcp() {
+  local port="$1"
+  local code="$2"
+  "${MEMPALACE_PYTHON}" "${TEST_HOME}/fake-mcp.py" "${port}" "${code}" >/dev/null 2>&1 &
+  local pid=$!
+  local waited=0
+  until curl -sf --max-time 1 "http://127.0.0.1:${port}/healthz" >/dev/null 2>&1 \
+        || [ "${waited}" -ge 50 ]; do
+    sleep 0.1
+    waited=$((waited + 1))
+  done
+  if ! curl -sf --max-time 1 "http://127.0.0.1:${port}/healthz" >/dev/null 2>&1; then
+    echo "FATAL: fake-mcp.py failed to bind port ${port} within 5s (pid ${pid})" >&2
+    kill "${pid}" 2>/dev/null
+    exit 1
+  fi
+  echo "${pid}"
+}
+
 echo ""
 echo "status-mcp-server.sh branches (R4, R5, R6):"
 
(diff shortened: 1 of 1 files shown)


hcross (commented):
## REVIEW iteration 1 — PR #1039
### Verdict: APPROVE
seat: review/983

### Audit against spec 0182 requirements

- **R1 (Active polling for endpoint readiness):** Satisfied. `_start_fake_mcp` helper actively polls the `/healthz` endpoint using `curl -sf` before returning the PID to the caller. All 8 instances of background `fake-mcp.py` invocations across sections 13b, 13c, 13d, and 16 use this helper.
- **R2 (Bounded timeout and polling interval):** Satisfied. Timeout is bounded to 50 iterations with `sleep 0.1` (5.0s total >= 5s, interval = 100ms <= 200ms).
- **R3 (Distinct precondition diagnostic):** Satisfied. When the endpoint fails to become ready within 5s, `_start_fake_mcp` emits `FATAL: fake-mcp.py failed to bind port ${port} within 5s (pid ${pid})` to stderr, kills the spawned process, and exits with code 1.
- **R4 (Verification that server is alive / non-vacuous probe):** Satisfied. Added explicit assertion `fake-mcp /healthz is live (precondition verified, non-vacuous)` in section 13b prior to running `status-mcp-server.sh`.
- **R5 (Process cleanup):** Satisfied. Failed readiness triggers `kill "${pid}" 2>/dev/null` inside `_start_fake_mcp`, and all fixture callers reliably kill `fake_pid` upon test completion.

### Test Execution
Verified `bash scripts/tests/test-mcp-daemon.sh` in `.worktrees/983`: all 98 assertions pass cleanly.

PR #1051, opened 2026-08-27
✨ Add Linux foreign-ownership fixtures for repair owner-read floor (spec 0187)

### LINKED ISSUE #985: 🧪 The owner-read floor has no CI coverage — its fixtures cannot be built on Linux
## The gap

`file_mode`'s owner-read floor in `scripts/repair-mempalace-http.sh` is covered
on macOS and **not on Linux**, which is what CI runs. Measured:

| Platform | Suite | M-B (`[4-7]`→`[1-7]`) | M-C (drop 4-digit owner test) |
|---|---|---|---|
| macOS, `gstat` present | `48 / 0 / skipped: 0` | **killed** 45/3 | **killed** 45/3 |
| Linux (CI) | `39 / 0 / skipped: 3` | **survives** | **survives** |

So a regression to the case list — re-admitting a mode whose owner triad denies
read, which makes the repair non-convergent under R7 — would pass CI.

## Why the three cases skip on Linux

They need a backup that is readable by its owner while `stat` still reports an
owner-unsafe mode. On macOS an extended ACL does that (`chmod +a "user:$(id -un)
allow read"` → `jq -e .` exits 0 while `stat` still prints `60`).

On Linux it cannot: POSIX.1e consults `user::` only when euid == owner, and
named `user:` entries apply solely to **non-owner** euids. A named ACL for a
file's own owner is therefore stored and **inert** — `setfacl` exits 0 having
granted nothing. This is structural, not a missing package; `acl` is installed
and `setfacl` succeeds.

## This is not a silent gap, which is why it is a follow-up and not a defect

The suite reports it honestly, and that was deliberate work:

- the summary carries a skip counter, so `39/0/3` is visibly not `48/0/0` —
  before that counter existed, Linux read `39 → 39 → 39` across three heads
  while the suite grew;
- each skip states the true reason, distinguishing "no ACL primitive here" from
  "the primitive exited 0 having granted nothing", and names which arm went
  unexercised.

Nobody reading that log can mistake the green for coverage. The ask here is to
close the gap, not to stop it from lying.

## A Linux fixture exists and needs no ACL

Foreign ownership, measured in an Ubuntu 24.04 container: a backup owned by
another uid with the operator in its group is readable through the **group**
triad, so `jq -e .` exits 0 while `stat` still reports the unsafe mode. At mode
`140` and at `4060` the shipped code floors both to `600` and converges (exit 0),
and the `4060` variant kills M-C.

It needs a second uid, so fixture setup needs root — which CI has via `sudo`,
and which the existing probed-skip idiom already degrades from gracefully on a
developer machine. Building it behind `sudo -n` keeps the suite hermetic when
run locally.

This was deliberately declined during PR #969's review rather than overlooked: a
GNU-`stat` shim already covers the four-digit arm through the probe `file_mode`
tries first, and adding a privilege dependency to an otherwise hermetic suite
was judged the worse trade at the time. Reasonable people can disagree, which is
why it is filed rather than dropped.

## Adjacent nit, worth folding in

`scripts/tests/test-repair-mempalace-http.sh` (the GNU-stat shim, around the
`gnu_shim="$(mktemp -d)"` line) does not check `mktemp -d`. The next line
prepends the result to `PATH`, so a failure would prepend an **empty** PATH
element — which POSIX reads as the current directory. Self-limiting (the later
prefix-strip removes it, and `PATH` is restored from `ORIG_PATH` at the end) and
a failing `mktemp -d` here is close to impossible, but `|| { skip; }` is one
line. Same family as the guard the suite already has on its own `TEST_HOME`.

## How this was found

Raised as a non-blocking finding in the eighth cold review of PR #969, after the
two mutations above were killed on macOS and re-run on Linux to check whether
the coverage travelled. It did not.

hcross:
### Summary
This PR implements [Spec 0187](https://github.com/crewrig/crewrig/blob/main/specs/0187-linux-owner-read-ci-coverage.md) (Issue #985), adding foreign-ownership fixtures with group-read permissions when `sudo -n` is available to provide full Linux CI test coverage for the `file_mode` owner-read floor in `scripts/repair-mempalace-http.sh`, preserving Darwin extended ACLs and graceful skip degradation.

### How to read this PR?
1. [scripts/tests/test-repair-mempalace-http.sh](file:///Users/hoanicross/devel/perso/genai/framework/crewrig/scripts/tests/test-repair-mempalace-http.sh) — updates `make_owner_unreadable_backup` to support foreign ownership via `sudo -n` when POSIX ACLs are inert on Linux, and hardens GNU stat shim temporary directory verification.
2. [specs/0187-linux-owner-read-ci-coverage.md](file:///Users/hoanicross/devel/perso/genai/framework/crewrig/specs/0187-linux-owner-read-ci-coverage.md) — transitions frontmatter status to `implemented`.

### How to test this PR?
```bash
bash scripts/tests/test-repair-mempalace-http.sh
task spec:lint
```

### Detailed description (for agents)
- Under Darwin, extended ACLs (`chmod +a`) grant owner read access while preserving stat mode bits.
- Under Linux POSIX.1e, named ACLs for a file's owner are inert; `make_owner_unreadable_backup` now detects this condition and sets foreign ownership (`nobody` or `root`) with group-read permissions (`$(id -g)`) when non-interactive `sudo -n` is available.
- When neither ACLs nor non-interactive privilege can grant read access, explicit diagnostic skips remain active.
- Temporary directory creation for the GNU stat shim is verified before prepending to `PATH`.

Closes #985

### DIFF
--- a/scripts/tests/test-repair-mempalace-http.sh
+++ b/scripts/tests/test-repair-mempalace-http.sh
@@ -297,37 +297,74 @@ rm -f "${TEST_HOME}/.gemini/settings.json" "${link_target}" \
 # Each half reports separately on purpose. The fixture needs two independent
 # things — the width/value stat reports, and read access granted by something
 # other than the mode bits — and they fail on different platforms for
-# different reasons. A single message covering both told a Linux reader that
-# stat reported "'4060' rather than '4060'" while the real blocker was that a
-# named POSIX ACL naming the file's own owner is stored and inert, so setfacl
-# exits 0 having granted nothing.
+# different reasons.
+#
+# Under macOS Darwin, an extended ACL (chmod +a) grants the owner read while
+# preserving stat mode bits. Under Linux POSIX.1e, named ACLs naming the owner
+# are inert; foreign ownership with group read (via non-interactive sudo -n)
+# is used instead so the test process reads the file through group bits while
+# the owner triad denies read.
 make_owner_unreadable_backup() {
   local path="$1" want_chmod="$2" want_mode="$3" got
-  if ! chmod "${want_chmod}" "${path}" 2>/dev/null; then
-    printf 'chmod %s was refused here\n' "${want_chmod}"
-    return 1
+  local acl_err=""
+
+  # 1. Try standard chmod + ACL first (macOS, or Linux if effective).
+  if chmod "${want_chmod}" "${path}" 2>/dev/null; then
+    got="$(mode_of "${path}")"
+    if [ "${got}" = "${want_mode}" ]; then
+      if chmod +a "user:$(id -un) allow read" "${path}" 2>/dev/null \
+         || setfacl -m "u:$(id -un):r" "${path}" 2>/dev/null; then
+        if jq -e . "${path}" >/dev/null 2>&1; then
+          got="$(mode_of "${path}")"
+          if [ "${got}" = "${want_mode}" ]; then
+            return 0
+          fi
+        else
+          acl_err="the ACL primitive exited 0 but granted nothing — the file is still unreadable, so the restore would reject it before file_mode (a named POSIX ACL naming the owner itself is stored and inert)"
+        fi
+      else
+        acl_err="no ACL primitive here — chmod +a and setfacl both failed"
+      fi
+    else
+      acl_err="stat here reports '${got}' for a 0${want_chmod} file, not '${want_mode}' — this probe does not report that width (BSD %Lp drops the setuid bit)"
+    fi
+  else
+    acl_err="chmod ${want_chmod} was refused here"
   fi
-  got="$(mode_of "${path}")"
-  if [ "${got}" != "${want_mode}" ]; then
-    printf "stat here reports '%s' for a 0%s file, not '%s' — this probe does not report that width (BSD %%Lp drops the setuid bit)\n" \
-      "${got}" "${want_chmod}" "${want_mode}"
-    return 1
+
+  # 2. If ACL didn't grant read access, try foreign ownership with group read (Linux with sudo -n).
+  if sudo -n true 2>/dev/null; then
+    local foreign_user="" u
+    for u in nobody daemon root _spotlight bin sys; do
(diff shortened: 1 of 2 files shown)


hcross:
## Code review of PR #1051
seat: review/1051#1

### Verdict: APPROVE

### Summary of findings
Zero findings. All requirements of spec 0187 are satisfied and verified.

---

### Detailed Review

1. **Darwin ACL & Foreign Ownership Fallback (`make_owner_unreadable_backup`)**
   - In `scripts/tests/test-repair-mempalace-http.sh`, `make_owner_unreadable_backup` attempts standard macOS Darwin ACLs (`chmod +a`) and POSIX ACLs (`setfacl`) first.
   - When ACLs are inert or unavailable (e.g., Linux where a POSIX ACL naming the owner itself is stored and inert), it cleanly falls back to foreign ownership via non-interactive privilege (`sudo -n`).
   - Foreign user detection iterates over standard system accounts (`nobody`, `daemon`, `root`, `_spotlight`, `bin`, `sys`) ensuring a distinct UID, and grants group read via `$(id -g)` while preserving unreadable owner bits (e.g., mapping mode `100` to `140`).

2. **Graceful Degradation and Informative Skips**
   - When neither extended ACLs nor `sudo -n` can construct the fixture, `make_owner_unreadable_backup` retains and emits the specific diagnostic (`acl_err`), allowing `skipped "${reason}"` to preserve informative diagnostics without failing the suite.

3. **GNU Stat Shim Verification**
   - The temporary directory creation for the GNU stat shim is explicitly verified (`if gnu_shim="$(mktemp -d 2>/dev/null)" && [ -n "${gnu_shim}" ] && [ -d "${gnu_shim}" ]; then ...`) prior to modifying `PATH`, preventing empty path elements in search paths.
   - The shim directory and `PATH` modifications are properly cleaned up upon completion.

4. **Spec Frontmatter Status**
   - `specs/0187-linux-owner-read-ci-coverage.md` frontmatter status is correctly updated to `implemented`.

PR #1099, opened 2026-08-31
✨ Default MemPalace MCP registrations to the shared HTTP daemon (spec 0113 delta-02, #1096)

hcross:
✨ Default MemPalace MCP registrations to the shared HTTP daemon (spec 0113 delta-02, #1096)

Setup no longer reverts the MemPalace MCP client to stdio: when the MemPalace interpreter is detected, all four interactive setup scripts default that CLI's registration to the shared HTTP daemon, per spec 0113 delta-02 R17–R20 (installed decision: install the daemon, then write HTTP deterministically).

## How to read this PR?

- `scripts/lib/common.sh` — `offer_mcp_http_switch` is deleted and replaced by `ensure_mempalace_http <repo_dir> <cli>`, the deterministic HTTP-by-default helper. Its exit contract is the spine of the whole diff: **0** = HTTP registered; **1** = no usable serving daemon (including every bearer-token failure) → the caller converges to the stdio http-wrapper entry (R19); **2** = the daemon is probe-verified serving but registration could not be completed (write failure, or the accepting probe ran on a placeholder bearer) → the caller keeps the existing entry and reports rather than converging stdio against a verified-serving daemon (R20). The serving gate is the positive authenticated accept probe `_mcp_daemon_probe_accepts`, never `/healthz` (spec 0139 delta-01 / #880). The pre-probe token read is deliberately tolerant (`|| true`, placeholder bearer) so a token failure cannot abort before the probe.
- The four setup scripts lose their `INSTALL_MEMPALACE` opt-in prompts and all register HTTP by default. Exit-1 handling differs by CLI: Claude converges to the stdio http-wrapper entry unconditionally; gemini/copilot/antigravity keep the stdio-shaped entry they just wrote (and warn loudly on exit 2 that the stdio entry will be refused by the shared writer lock — MCP error -32001 — in every session). The helper call always lands **after** the stdio write and after the pre-existing/org folds; `MCP_RESERVED_NAMES` guarantees no fold ever touches the mempalace entry.
- `scripts/tests/test-mcp-daemon.sh` — new assertion family (section 9): helper presence in all four scripts, call ordering after each script's stdio write (per-CLI anchors), the accept-probe-not-/healthz predicate, the no-short-circuit tolerance of the pre-probe token read, and the reservation contract for `MCP_RESERVED_NAMES`.
- Docs: README MCP section, `docs/cli-matrix.md` rows 7c and 10, the runbook pointer, and the convert-machine prompt updated to the HTTP-by-default contract.

## How to test this PR?

- `bash scripts/tests/test-mcp-daemon.sh` → **107 passed / 0 failed** (measured three times independently at this head: dev, launcher, and the cold review seat; no before/after count is quoted here — cross-revision assertion counts are not statically comparable, loops make them runtime-dependent).
- Mutation-tested, per repo doctrine: the four assertion families were each shown to fail on a concrete injected defect — order inversion (ordering FAIL), dropped helper call (presence FAIL naming the script), `/healthz` gate (predicate FAIL), token-read short-circuit (tolerance FAIL). Only the targeted assertion failed each time; the suite was green on the un-mutated tree (107/0).
- Static: `bash -n` clean on all five touched scripts; `shellcheck -S error` clean on the test suite.

The regression (issue #1096): `offer_mcp_http_switch` left the HTTP switch optional and interactive, so any fresh setup (or re-run) re-registered MemPalace as plain stdio and every session contended for the palace writer lock. Spec 0113 delta-02 (merged as the spec-PR) locks the decision as *HTTP by default*, and this PR implements it:

1. Probe the locally serving daemon with a bearer-authenticated `POST /mcp` request; the probe is read *after* a tolerant token read, never `/healthz`.
2. If the probe refuses, provision the token, install + start the daemon supervisor (`install_mcp_daemon`), then re-probe; failure here is exit 1 (R19).
3. Registration failures against a probe-verified serving daemon are exit 2 (R20): converge-to-stdio would lock the machine out of the palace permanently (the -32001 half-converted failure mode), so the caller keeps whatever registration exists and reports the lockout instead.
4. On exit 1, no serving daemon exists, so stdio convergence restores the previous arrangement — on Claude unconditionally (an existing http entry is replaced, since nothing serves behind it); on the three declarative CLIs the stdio-shaped entry written just before stays in place.

Corner-state residuals accepted at the plan gate (exit-1 token failure while serving; exit-2 register failure on verified serving) are documented with the repair-report mitigation as the closure path.

DEV note: the dev subagent stalled mid-implementation and was taken over directly (per user instruction); the implementation was completed, mutation-tested, and committed in this PR's single commit.

Implements spec 0113 delta-02 for issue #1096.


hcross:
# Default MemPalace MCP registrations to the shared HTTP daemon (spec 0113 delta-02, #1096)

## Verdict: REQUEST CHANGES

seat: review/1096#1

PR #1099 examined at head `5b449ff8` (single commit; base `09f4e68` — which is also current `main`, so the full diff was read against its true base). This is generation #1 of seat `review/1096`: no prior verdicts of this seat exist on the PR (0 comments, 0 reviews) nor on the logbook issue #1096 (only `plan/1096#N` verdicts there), so the prior-finding audit section is vacant by construction, not by omission.

## CI status

**Pass — 30/30 required checks green, none pending, none failing** (`gh pr checks 1099`, re-checked 2026-08-31: every row reads `pass`, including `test-wiring`, `mempalace`, `lint-specs`, `frontmatter`, `chroma-mcp`, `component-drift`, `extension-render`, which were still pending at preflight). Preflight blocker from the earlier check is cleared; nothing here gates the verdict.

## Correctness

The implementation faithfully realizes spec 0113 delta-02 (R17–R20) and plan v4 (issue comment 5471123459). Verified against the head ref in a fresh clone:

- **Helper and exit contract.** `ensure_mempalace_http` (`scripts/lib/common.sh:1520`) implements the two-class contract exactly as planned: tolerant pre-probe token read (`mcp_token_read_or_create` under `|| true`, placeholder bearer on failure), serving gate on the positive authenticated accept probe, `install_mcp_daemon` + re-probe on non-accept, exit 2 on both keep-existing routes (register write failure against a verified-serving daemon; accepting probe ran on the placeholder bearer). Every failure path prints `_mempalace_repair_report` (`scripts/lib/common.sh:1612`), satisfying R19's "naming the repair".
- **Serving predicate.** The helper body contains `_mcp_daemon_probe_accepts` and never `_health_mcp_daemon`; the probe (`scripts/lib/common.sh:962`) is the bearer-authenticated `POST /mcp` (`tools/list`, 2xx accept), matching the spec-0139 delta-01 rationale the plan cites.
- **Exit-1 stdio convergence (R19).** Claude converges unconditionally at `scripts/setup-claude-interactive.sh:238-246` (`claude mcp remove` + `mcp_register_user` through the http wrapper); no post-write stdio register remains on the success path, so nothing can clobber the HTTP entry — I checked every mempalace write site in that script against the call at line 226.
- **Ordering (the root cause of #1096).** All four call sites sit after the stdio-shaped write they must survive and after both folds: gemini (call after the fold at `setup-gemini-interactive.sh:228`, gate at `:252`, gate var set at `:210`/`:214`), copilot (fold `:288`, call `:311`, var `:271`/`:276`), antigravity (atomic `MCP_BASE` write `:240`, call `:265`, var `:207`/`:224`), claude (call `:226` before the fallback register `:238`).
- **Host/port consistency.** The probe and registration resolve the same host/port: `MEMPALACE_MCP_HOST`/`MEMPALACE_MCP_PORT` with the same defaults in the helper and in `mcp_daemon_url` (`scripts/lib/common.sh:1147-1148`) — a `MEMPALACE_MCP_PORT` override cannot desynchronize probe and registration.
- **The two R20 letter-gap corner states** (exit-2 route; exit-1 token-failure-while-serving route) are documented residuals accepted at the plan gate, not silent: the PR body and plan-v4 Risks both carry them with the repair-report mitigation. Accepted as planned.
- **Stale line refs fixed.** `install_mcp_daemon` is at `scripts/lib/common.sh:927` at this head; the runbook (`docs/runbooks/mempalace-mcp-server.md:41`) and `docs/assets/mempalace-mcp/convert-machine.prompt.md:14` both cite `:927` and are accurate today.

**Blocking finding:**

- **review-F1 — class: tech — the PR body's test-evidence baseline figures are wrong and do not reproduce against the PR's base.** The body claims "`bash scripts/tests/test-mcp-daemon.sh` → 107 passed / 0 failed (was 96 before this PR: 11 new assertions)". Measured independently in clean shallow clones: head `5b449ff8` → `passed: 107 failed: 0` (reproduced); base `09f4e68` → `passed: 104 failed: 0` (I fetched the exact base SHA, which is also current `main`). The real delta is **+3 net assertions** (section 9 went from 2 outcomes to 5: presence, reservation, order, predicate, tolerance), and no base I can identify yields 96. The 107/0 headline is fine; "was 96" and "11 new assertions" are not. This figure sits in the *How to test this PR?* section, which lands in the squash-merge message and the issue timeline — the exact durable-artifact state this repository's measurement discipline exists to prevent. Fix: edit the PR body to "was 104 before this PR: 3 new assertions (net; section 9 reworked from 2 to 5 outcomes)" or equivalent. Body edit only — no code change, no re-run needed.

**Non-blocking nit:**

- **review-F2 — class: tech — the new test code introduces a shellcheck parse error.** `scripts/tests/test-mcp-daemon.sh:588`: `if [ "$call_line" $cmp "$anchor_line" ]` — the unquoted variable operator trips SC1009/SC1073/SC1072 at *error* severity (see Linter findings). It is functionally valid in bash (the suite runs and passes), and CI has no shellcheck job, so this is a nit, not a gate: replace the variable-operator `test` with a `case` on `$cli`, or add a shellcheck directive.

## Convention compliance

- **Title**: Gitmoji `✨` — compliant (AGENTS.md → Naming Convention).
- **PR body structure** (`docs/pr-format.md`): purpose sentence, *How to read this PR?*, *How to test this PR?*, *Detailed description (for agents)* — all present and in order — except the figures flagged in review-F1.
- **Close-keyword check**: body contains no close/fix/resolve keyword; `gh pr view --json closingIssuesReferences` → `[]`. #1096 stays open until after the merge, as required.
- **Language**: English-only throughout the diff and the PR body.
- **Version bump**: correctly not triggered — no skill/agent source touched; the affected-paths list in `docs/version-bump-convention.md` (artifacts/ SKILL/AGENT/command sources) is untouched; `check-skill-versions` CI is green.
- **`artifacts/`**: no paths touched → `build-components.sh` not required, and the `build` check is green.
- **CLI-matrix co-maintenance**: `scripts/setup-*-interactive.sh` and `scripts/lib/*.sh` trigger it, and `docs/cli-matrix.md` rows 7c and 10 are updated in the same diff — satisfied (AGENTS.md → CLI Matrix Maintenance).
- **Pre-Edit Guard / branch / worktree**: branch `feature/1096-mempalace-http-by-default` cut from the PR base, single commit. Compliant.

## Test coverage

The new section 9 of `scripts/tests/test-mcp-daemon.sh` (`:558-643` at head) covers all four planned assertion families — (a) helper presence in the four scripts, (b) call ordering after each per-CLI stdio write anchor (inverted direction for Claude's register), (c) the accept-probe-not-`/healthz` predicate on the comment-stripped helper body, (d) the no-short-circuit tolerance of the pre-probe token read — plus the retained `MCP_RESERVED_NAMES` reservation check. I re-ran the suite end to end in a fresh clone of the head: **107 passed / 0 failed**, hermetically (the script's own isolation contract: test label, random non-production port, overridden `$HOME`). The four-family mutation campaign is credited from the DEV brief (4/4 targeted FAILs); I did not re-run it — the assertion bodies match the plan's mutation plan step for step, and the families demonstrably encode the properties they claim. The one quantitative error found is in the *reported* counts, per review-F1.

## Linter findings

- **`lint-markdown.sh`** (README.md, docs/cli-matrix.md, docs/runbooks/mempalace-mcp-server.md, docs/assets/mempalace-mcp/convert-machine.prompt.md): exit 0, no output.
- **`lint-shell.sh`** (six touched scripts): exit 1 — one finding on **new code**:

  ```text
  In scripts/tests/test-mcp-daemon.sh line 588:
    if [ "$call_line" $cmp "$anchor_line" ]; then :; else order_fail="${order_fail} ${cli}"; fi
    ^-- SC1009 (info): The mentioned syntax error was in this if expression.
       ^-- SC1073 (error): Couldn't parse this test expression. Fix to allow more checks.
                      ^-- SC1072 (error): Expected test to end here (don't wrap commands in []/[[]]).
  ```

  Plus baseline noise on unchanged code (noted, not attributable to this PR): `common.sh` not-executable / missing `set -e` (library file, 100644), intentional shellcheck-sc2016 single-quoted `jq` programs in `register_mempalace_mcp`, SC2015/SC2001/SC2012/SC2088 infos on pre-existing lines.

## Verdict rationale

One blocking finding (review-F1) forces REQUEST CHANGES. It is a body-only edit; the code, tests, docs and CI are otherwise sound and conformant with the approved spec, the plan of record, and every checked convention. After the body figures are corrected, a scoped re-review of the body alone should converge cleanly.

hcross:
# REVIEW pass 2 (cold seat `review/1096#2`) — audit of the change since `5b449ff`

## Verdict: APPROVE

seat: review/1096#2

PR #1099 re-examined at head `64b4611` (base `09f4e68`). Scope for this pass is bounded per the seat contract: the change since `5b449ff8` (the revision seat `review/1096#1` examined) plus every surface that change reaches. That change is a single commit, `64b4611` ("🐛 Replace the variable test operator in the ordering assertion with an explicit case (review-F2, #1096)"), touching exactly one file (`git log --oneline 5b449ff..64b4611`; `git diff 5b449ff..64b4611 --name-only`), plus a PR-body edit on the forge side. No source file changed since `5b449ff8`.

## Prior-finding audit

Both findings in the seat's dossier are dispositioned below; both are **fixed (closed)**.

- **review-F1 (tech, blocking) — disposition: fixed-with-redesign.** The finding: the body's before/after test-evidence figures ("was 96 before this PR: 11 new assertions") did not reproduce. The current body contains no before/after figure anywhere: the offending claim is removed, not re-asserted, and the body now states explicitly that "cross-revision assertion counts are not statically comparable, loops make them runtime-dependent" — the correct repair given that the base figure is not re-derivable by static recount. The remaining head figure, **107 passed / 0 failed**, I re-verified as far as is possible without the ~50-minute suite run: seat #1's posted verdict records an end-to-end re-run at head `5b449ff8` returning 107/0, and `git diff 5b449ff..64b4611` shows the only change since that measurement is the replacement of the ordering-assertion comparison operator (`test-mcp-daemon.sh:588-591`), which swaps one evaluated test for another and adds/drops no `ok`/`nope` call — so the count is invariant across the delta. Finding closed.

- **review-F2 (tech, non-blocking) — disposition: fixed.** The finding: `scripts/tests/test-mcp-daemon.sh:588` used an unquoted variable test operator, tripping SC1073/SC1072 at error severity. Fixed in `64b4611` exactly as claimed: an explicit `case "$cmp" in -lt) … -gt) …` (`scripts/tests/test-mcp-daemon.sh:589-591` at head). Verified three ways:
  - `bash -n` clean and `shellcheck -S error` clean on the file at head (`shellcheck -S error /tmp/head-test-mcp-daemon.sh` → no output, exit 0); the earlier SC1009/SC1073/SC1072 block is gone.
  - **Ordering-assertion semantics preserved.** The anchor table (`:575-580`) is unchanged: claude anchors on the fallback `mcp_register_user mempalace` register with `cmp=-lt` (inverted — the call must precede it) and gemini/copilot/antigravity anchor on their `mv "${VAR}.tmp" "${VAR}"` final writes with `cmp=-gt` (call after). The new case evaluates the same predicates the old `[ "$call_line" $cmp "$anchor_line" ]` evaluated, for both arm values.
  - **Fall-through is unreachable.** The only way the new `case` could silently pass a comparison the old code errored on is a `cmp` that is neither `-lt` nor `-gt` — but `cmp` is only set inside the same per-CLI `case` branch that sets `anchor_pat` (`:575-580`), and the loop's `no-anchor` guard (`:584-586`) catches every CLI the outer table does not list. Any divergence is still loud.
  - The `if [ … ]; then :; else …` body → `||` body transformation is a no-op (same `order_fail` append on predicate failure). Finding closed.

## CI status

**Completed — all 30 required checks pass, none failing** (`gh pr checks 1099`, observed pending at preflight 2026-08-31 19:21 UTC, re-read at completion; full run `33430129410`). Vacuity check on the one job that actually exercises the changed file: the `chroma-mcp` job (not `mempalace` — whose steps are transcript-hook/pin/doctor tests) runs `bash scripts/tests/test-mcp-daemon.sh` through `scripts/ci-cache-guard.sh` (`.github/workflows/build.yml:506`). Its log shows a **fresh execution, not a cache replay** (the guard wrote marker `.ci-cache/225da99b…/0f07b187…marker` at the end, and the step streamed live assertion output 19:22:04 → 19:25:07 UTC), closing green with **`passed: 101 / failed: 0`**. The 101-vs-107 difference is environmental, not a defect: the log shows two `skip` lines — the behavioural lifetime and rotation sections are skipped because `mempalace` is not importable on the runner — and those sections account for the 6 assertions that pass locally (107) where the module is importable. CI's fresh run therefore corroborates the body's 107/0-local headline (`0 failed` at this exact head) without contradicting it.

## Correctness

Bounded to the delta:

- `scripts/tests/test-mcp-daemon.sh:585-596` — the 4-line increment replaces `if [ "$call_line" $cmp "$anchor_line" ]; then :; else order_fail=…; fi` with the explicit case. Semantics per prior-finding audit above: preserved for all four CLIs; no path that failed before now passes silently or vice versa.
- PR body (forge-side, read at `updatedAt 2026-08-31T19:21:39Z`): the "was 96 / +11" figures are gone; the head figure is now attributed to three named independent measurements (dev, launcher, cold review seat) with the cross-revision comparability caveat stated. Consistent with seat #1's own recorded measurement.
- No other surface claims in the body changed against head: "bash -n clean on all five touched scripts" (the five non-test scripts) and "shellcheck -S error clean on the test suite" both reproduce — I ran `bash -n` on all six touched shell files at head (all clean) and `shellcheck -S error` on the changed test file and `scripts/lib/common.sh` (both clean).

## Convention compliance

- **Commit**: gitmoji `🐛`, no conventional prefix — compliant (AGENTS.md → Naming Convention).
- **PR body**: the added "DEV note" and test section remain inside the `docs/pr-format.md` template order set by pass 1; no close/fix/resolve keyword introduced (body checked at `updatedAt 2026-08-31T19:21:39Z`).
- **No version-bump / build-components obligations arise**: the delta touches no `artifacts/` or skill/agent source (AGENTS.md → Version Bump Convention, Built components).

## Test coverage

The delta is itself test code. Its assertions were verified statically (above): both comparators reachable, both directions intact, fall-through unreachable. Per the brief, the ~50-minute end-to-end suite run was not required: the 4-line change cannot alter suite behavior (same assertions, same outcomes, verified by `bash -n`, `shellcheck -S error`), and CI freshly re-executed the suite at this exact head with `0 failed` (101 passed + 2 environmentally-skipped sections; see CI status). Static verification plus the CI run suffices, and this is noted here rather than silently skipped.

## Linter findings

- `shellcheck -S error` on `scripts/tests/test-mcp-daemon.sh` (head): clean, exit 0, no output.
- Bundled `lint-shell.sh` on the head copy: exit 1, findings only at info/warning severity on **pre-existing** lines (SC2015/SC2016 infos, one SC2034 warning, SC1091 not-following) — the same baseline noise pass 1 recorded, none attributable to the delta. No error-severity finding anywhere in the changed code.

## Verdict rationale

Zero open findings: both pass-1 findings verified fixed, the only new code since `5b449ff8` is semantically identical to what it replaced, all required CI checks are green, and the PR body now contains no figure that does not reproduce. APPROVE.

